PR #1, opened 2019-04-02
added task description and tests

ivadimko:


ivadimko:
@mgrinko ci checks have failed because it is clean task without solution

PR #2, opened 2019-04-03
updated eslint config, updated jsdoc description, fixed lines length …

ivadimko:
…issue


src/averageAges.js
```diff
@@ -2,10 +2,12 @@
 
 /**
  * Implement calculateMenAverageAge function
- * It should return average age of men in array
- * If `century` is specified then calculate only for men living in this century
+ *
+ * Function returns average age of men in array
```
roman-mate: Add line break only
when 
you reach 80
characters. Line breaks in random places 
are hard to 
read. Like in this example :)

src/averageAges.test.js
```diff
@@ -8,26 +8,30 @@ const {
 } = require('./averageAges');
 
 test('Average men age', () => {
-    expect(+calculateMenAverageAge(people).toFixed(2)).toBe(61.67);
+  expect(+calculateMenAverageAge(people).toFixed(2))
```
roman-mate: 1. why is + needed here?
2. how about we use?
toBeCloseTo
https://jestjs.io/docs/en/expect#tobeclosetonumber-numdigits

PR #5, opened 2019-04-11
Yevhen Kryvdiuk

Yevhen-Kryvdiuk:


src/averageAges.js
```diff
@@ -15,11 +15,20 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with logical operators (&&, ||) or ternary operator (?:)
-  // without nesting
+  const MALE = 'm';
```
YuriiHoliuk: Will be better to declare constants like this outside the function. it will allow to overuse it

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
+  function fromCentury(personDied) {
```
YuriiHoliuk: Great, but ideally you can declare a pure function in upper scope
```
  function fromCentury(century, personDied) {
    return Math.ceil(personDied / 100) === century;
  }
```

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
+    return Math.ceil(personDied / 100) === century;
+  }
+  const mans = people.filter((person) => {
+    return century ? (person.sex === MALE && fromCentury(person.died)) : person.sex === MALE;
+  });
+  const mansAges = mans.map((person) => {
+    return person.died - person.born;
```
YuriiHoliuk: You can do it inside reduce, but ok

src/averageAges.js
```diff
@@ -34,7 +43,22 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const FEMALE = 'f';
```
YuriiHoliuk: ^

src/averageAges.js
```diff
@@ -34,7 +43,22 @@ function calculateMenAverageAge(people, century) {
...
+  function isMother(people, mother) {
```
YuriiHoliuk: ^

src/averageAges.js
```diff
@@ -34,7 +43,22 @@ function calculateMenAverageAge(people, century) {
...
+    return people.some((person) => {
+      return person.mother === mother;
+    });
+  }
+  const women = people.filter((person) => {
+    return withChildren ? (person.sex === FEMALE && isMother(people, person.name)) : person.sex === FEMALE;
+  });
+  const womenAges = women.map((person) => {
```
YuriiHoliuk: ^

src/averageAges.js
```diff
@@ -34,7 +43,22 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const FEMALE = 'f';
+  function isMother(people, mother) {
+    return people.some((person) => {
+      return person.mother === mother;
+    });
+  }
+  const women = people.filter((person) => {
+    return withChildren ? (person.sex === FEMALE && isMother(people, person.name)) : person.sex === FEMALE;
```
YuriiHoliuk: when using `isMother` complexity becomes O(n^2) (a loop inside a loop). But it may be solved with O(n) complexity. If you want, you could try to create a hash map and `isMother` operation will be cost O(1) instead of O(n)
https://www.youtube.com/watch?v=v4cd1O4zkGw
http://bigocheatsheet.com/

src/averageAges.js
```diff
@@ -52,7 +76,27 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const SON = 'm';
```
YuriiHoliuk: You can use the same constant `MAN`

src/averageAges.js
```diff
@@ -52,7 +76,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const mothersAndChildren = people.map((person) => {
+    const mother = people.find((mother) => {
```
YuriiHoliuk: ^ find make it O(n^2)

src/averageAges.js
```diff
@@ -52,7 +76,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      return mother.name === person.mother;
+    });
+    return mother ? [mother, person] : false;
+  });
+  const filteredMothersAndChildren = mothersAndChildren.filter((i) => {
```
YuriiHoliuk: do not use one letter names(group, motherWithChild, item...)

PR #19, opened 2019-08-20
Solution

PastushenkoVik:


src/averageAges.js
```diff
@@ -15,11 +15,14 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with logical operators (&&, ||) or ternary operator (?:)
-  // without nesting
+  const men = people
+    .filter(item =>
```
YuriiHoliuk: do not break line just after `=>`
it is not convenient, hard to read and I think may cause bugs during interpretation, like 
```
if (true) console.log('a')
console.log('b');
```
but I am not shure

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
+      item.sex === 'm'
+      && (century === undefined
+        || (item.died < century * 100 && item.died >= (century - 1) * 100)))
+    .map(item => item.died - item.born);
```
YuriiHoliuk: Your code looks intuitive, but you iterate array 3 times, can you move all code to reduce?

src/averageAges.js
```diff
@@ -34,7 +37,14 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people
+    .filter(item =>
+      item.sex === 'f'
+      && (withChildren === undefined
```
YuriiHoliuk: explicite is good, you do not must change it, but more convenient just `!withChildren` if you do not need differences between `false`, `null`, `undefined`, etc.

src/averageAges.js
```diff
@@ -34,7 +37,14 @@ function calculateMenAverageAge(people, century) {
...
+        || people.find(child => child.mother === item.name) !== undefined))
+    .map(item => item.died - item.born);
+
+  return women.reduce((sum, item) => sum + item, 0) / women.length;
```
YuriiHoliuk: the same comments as in first function

src/averageAges.js
```diff
@@ -34,7 +37,14 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people
+    .filter(item =>
+      item.sex === 'f'
+      && (withChildren === undefined
+        || people.find(child => child.mother === item.name) !== undefined))
```
YuriiHoliuk: can you rewrite without nested loops

src/averageAges.js
```diff
@@ -52,11 +62,17 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people
+    .filter(child => (onlyWithSon === undefined || child.sex === 'm')
+      && people.find(parent => parent.name === child.mother) !== undefined)
```
YuriiHoliuk: nested loop

YuriiHoliuk: Also, there are useful `some` method

src/averageAges.js
```diff
@@ -52,11 +62,17 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    .map(child =>
+      child.born - people.find(mother => mother.name === child.mother).born);
```
YuriiHoliuk: nested loop again, to find the same person, try to optimize performance

src/averageAges.js
```diff
@@ -52,11 +62,17 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  return children.reduce((sum, item) => sum + item, 0) / children.length;
```
YuriiHoliuk: Maybe, can you reuse this part or bigger

YuriiHoliuk (changes requested):

src/averageAges.js
PastushenkoVik: In this case we have only conditions, then I want indent them at the same column
I tried another indent. you will see

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with logical operators (&&, ||) or ternary operator (?:)
-  // without nesting
+  let counter = 0;
+
+  return people
+    .reduce((ageSum, man) => {
+      if (man.sex === 'm' && (century === undefined
+        || (man.died < century * 100 && man.died >= (century - 1) * 100))) {
+        counter++;
```
YuriiHoliuk: mutate variables outside callback(map, reduce) is bad practice

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with logical operators (&&, ||) or ternary operator (?:)
-  // without nesting
+  let counter = 0;
+
+  return people
+    .reduce((ageSum, man) => {
+      if (man.sex === 'm' && (century === undefined
+        || (man.died < century * 100 && man.died >= (century - 1) * 100))) {
```
YuriiHoliuk: hard to read, but ok

src/averageAges.js
```diff
@@ -34,7 +42,21 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothers = people.map(item => item.mother);
+
+  let counter = 0;
+
+  return people
+    .reduce((ageSum, woman) => {
+      if (woman.sex === 'f'
+        && (withChildren === undefined || mothers.includes(woman.name))) {
```
YuriiHoliuk: includes also have loop inside

src/averageAges.js
```diff
@@ -52,11 +74,25 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const mothers = new Map(people.map(item => [item.name, item.born]));
```
YuriiHoliuk: This is intuitive to use Map, but plain object will be enough, will work faster and has shorter syntax

YuriiHoliuk (changes requested):
Стосовно цього може і ок
Колбек мав би бути чистою функцією, але без цього буде більше коду, як хочеш

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // without nesting
+  return people.reduce((ageSum, man) => {
+    if (man.sex === 'm' && (century === undefined
+      || (man.died < century * 100 && man.died >= (century - 1) * 100))) {
+      return ageSum + man.died - man.born;
+    }
+
+    return ageSum;
+  }, 0) / people.filter(man => (man.sex === 'm' && (century === undefined
+    || (man.died < century * 100 && man.died >= (century - 1) * 100)))).length;
```
YuriiHoliuk: you duplicate a lot of code, also use unneeded filter
Proposed solution for reduce only:
```javascript
people.reduce(({ sum, count }, person) => {
  if (...) {
    return { sum: sum + ..., count: count + 1 }
  }
  return { sum, count };
}, { sum: 0, count: 0 })
```

Your previous solution with external variable was better then current
Or you can just use for loop

src/averageAges.js
```diff
@@ -34,7 +38,22 @@ function calculateMenAverageAge(people, century) {
...
+
+  return people.reduce((ageSum, woman) => {
+    if (woman.sex === 'f'
+      && (withChildren === undefined
+        || mothers[woman.name] !== undefined)) {
+      return ageSum + woman.died - woman.born;
+    }
+
+    return ageSum;
+  }, 0) / people.filter(woman => woman.sex === 'f'
```
YuriiHoliuk: the same

src/averageAges.js
```diff
@@ -52,11 +71,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  return people.reduce((age, child) => {
+    if ((onlyWithSon === undefined || child.sex === 'm')
+      && mothers[child.mother] !== undefined) {
+      return age + child.born - mothers[child.mother];
+    }
+
+    return age;
+  }, 0) / people
+    .filter(child => (onlyWithSon === undefined || child.sex === 'm')
```
YuriiHoliuk: the same

YuriiHoliuk (changes requested):

src/averageAges.js
```diff
@@ -34,7 +36,17 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothers = people.reduce((obj, item) => {
+    obj[item.mother] = '';
```
YuriiHoliuk: will be simpler with `obj[item.mother] = true;`

YuriiHoliuk (approved):

PR #21, opened 2019-08-24
solution

youvovk:
[LINK]()

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -15,13 +15,27 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with logical operators (&&, ||) or ternary operator (?:)
-  // without nesting
-}
+  let countMen = 0;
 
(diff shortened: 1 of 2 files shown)


src/averageAges.js
```diff
@@ -52,11 +89,36 @@ function calculateWomenAverageAge(people, withChildren) {
...
+        && countFamily++
+      : foundChild
+        && child.sex === 'm'
+        && countFamily++;
+
+    const multiplateMother = foundMother !== false
+      ? acum + child.born - bornMother
+      : acum;
+
+    const withSonMother = child.sex === 'm'
```
YuriiHoliuk: ^

src/averageAges.js
```diff
@@ -15,13 +15,18 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with logical operators (&&, ||) or ternary operator (?:)
-  // without nesting
-}
+  const filteredMan = people.filter(person => {
+    if (century === undefined && person.sex === 'm') {
+      return true;
+    } else if (century === Math.ceil(person.died / 100) && person.sex === 'm') {
+      return true;
```
YuriiHoliuk: you can just return correct expression, but ok

PR #23, opened 2019-08-26
solution Ihor Sobko (js_average-ages)

igaryok:


src/averageAges.js
```diff
@@ -15,11 +15,11 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with logical operators (&&, ||) or ternary operator (?:)
-  // without nesting
+  return people.filter(person =>
+    (person.sex === 'm')
+      && (!century || (century && (Math.ceil(person.died / 100)) === century)))
```
YuriiHoliuk: too hard to read, understand and maintain

src/averageAges.js
```diff
@@ -34,7 +34,12 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  return people.filter(person =>
+    (person.sex === 'f')
+      && (!withChildren || (withChildren && (people.some(child =>
```
YuriiHoliuk: the same

src/averageAges.js
```diff
@@ -52,11 +57,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+        !onlyWithSon || (onlyWithSon && child.sex) === 'm')))
+    .reduce((diffAges, mother) => {
+      const diffForMather = people.filter(
+        child => child.mother === mother.name && (
+          !onlyWithSon || (onlyWithSon && child.sex) === 'm'))
+        .map(child => child.born - mother.born);
+      diffAges.push(...diffForMather);
+
+      return diffAges;
+    }, []);
```
YuriiHoliuk: Why do you need reduce with array?

src/averageAges.js
```diff
@@ -52,11 +57,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // write code here
+  const arrayAllDifferences = people.filter(
+    person => people.some(
+      child => child.mother === person.name && (
+        !onlyWithSon || (onlyWithSon && child.sex) === 'm')))
+    .reduce((diffAges, mother) => {
+      const diffForMather = people.filter(
+        child => child.mother === mother.name && (
+          !onlyWithSon || (onlyWithSon && child.sex) === 'm'))
+        .map(child => child.born - mother.born);
```
YuriiHoliuk: Cannot read it, also code is duplicated

src/averageAges.js
igaryok: > too hard to read, understand and maintain

refactored

src/averageAges.js
igaryok: > the same

refactored

src/averageAges.js
igaryok: > Why do you need reduce with array?

refactored

src/averageAges.js
igaryok: > Cannot read it, also code is duplicated

refactored

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
-  // replace `if ()` statement with logical operators (&&, ||) or ternary operator (?:)
-  // without nesting
+  let onlyMen = people.filter(person => person.sex === 'm');
+
+  if (century) {
+    onlyMen = onlyMen.filter(man => Math.ceil(man.died / 100) === century);
+  }
+
+  return onlyMen.reduce(
+    (acc, man, i, arr) => acc + (man.died - man.born) / arr.length, 0);
```
YuriiHoliuk: if you leave `(` on previous line, `)` - should be on the next line

YuriiHoliuk: instead of apply filter twice will be better to filter once

igaryok: > instead of apply filter twice will be better to filter once

ok
refactored

PR #94, opened 2020-06-03
Solution

ogsevko:


solaryasha (approved):
Not bad ;)

ogsevko:
Well, I suppose there is a solution only with array methods, but it's too f-ing hard to think of one :)

mriynyk (approved):

PR #114, opened 2020-08-11
idk whats happening and my head explodes

ArtemiiKab:


src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let men = people.filter(person => person.sex === 'm');
+
+  if (arguments.length > 1) {
```
irenhh, line 20: just check for century explicitly

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
+    men = men.filter(man => century === Math.ceil(man.died / 100));
+  }
+
+  men = men.map(man => man.died - man.born);
```
irenhh, line 24: creete another variable for those actions, men is for other data

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
+
+  const result = parseFloat((men.reduce((acc, rec) => {
```
irenhh, line 26: Number()

src/averageAges.js
```diff
@@ -34,7 +42,20 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let women = people.filter(person => person.sex === 'f');
+
+  if (arguments.length > 1) {
+    women = women.filter(woman => people
+      .find(child => child.mother === woman.name));
+  }
+
+  women = women.map(woman => woman.died - woman.born);
```
irenhh, line 52: why did you decide to store different values in the array with women objects?

src/averageAges.js
```diff
@@ -52,7 +73,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  women.map(woman => people
+    .map(child => {
+      if (child.mother === woman.name) {
+        if (arguments.length === 1 || child.sex === 'm') {
+          sum += child.born - woman.born;
+          counter++;
+        }
+      }
+    }));
```
irenhh, line 88: map should return something without side effects

ArtemiiKab: Here I'm using map as a substitute for a 'for loop', I don't even store the result of mapping in the var, so inputing any return would go to nowhere anyway. In normal project, I think I would use a for loop in this case instead of mapping as I actually do not need an array. Maybe I'm wrong)

src/averageAges.js
ArtemiiKab: You mean (century !== undefined) ? I was discussing this check with Dennis Pryadka, and he kinda converted me to arguments.length instead, because it saves from the case when undefined is an actual value assigned to century. Could you please correct me if I'm missing something?

src/averageAges.js
ArtemiiKab: agree 100%

PR #115, opened 2020-08-11
Solution

uni-8:


src/averageAges.js
```diff
@@ -15,11 +15,15 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menAge = people
```
irenhh, line 18: from the name of the variable I can assume you wanna store a number here, but it's not true

fix the name so that it describes the array properly

uni-8: fixed

PR #119, opened 2020-08-12
Solution

karinaKudatska:


src/averageAges.js
```diff
@@ -15,11 +15,15 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const male = people.filter(person => century
```
krajnova, line 18: maybe it's better to use 'men' here?

src/averageAges.js
```diff
@@ -34,7 +38,15 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const female = people.filter(person => withChildren
```
krajnova, line 41: what`s about women?

src/averageAges.js
irenhh: absolutely, or males
array of values hence the plural name

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const male = people.filter(person => century
+    ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+    : person.sex === 'm');
+
+  const menAge = male.map(men => men.died - men.born);
+  const averageMenAge = menAge.reduce((age1, age2) =>
+    age1 + age2) / menAge.length;
```
irenhh, line 24:
```suggestion
  const averageMenAge = menAge.reduce((age1, age2) => (
    age1 + age2
  )) / menAge.length;
```

src/averageAges.js
irenhh: the same

src/averageAges.js
```diff
@@ -34,7 +38,15 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const female = people.filter(person => withChildren
+    ? person.sex === 'f' && people.some(child => child.mother === person.name)
+    : person.sex === 'f');
+
+  const womenAge = female.map(women => women.died - women.born);
+  const averageWomenAge = womenAge.reduce((age1, age2) =>
+    age1 + age2) / womenAge.length;
```
irenhh, line 47: format in a right way

PR #120, opened 2020-08-12
solution

Andrii256:


src/averageAges.js
```diff
@@ -52,7 +70,33 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const birthAges = [];
+
+  if (onlyWithSon) { // only son borning ages
```
irenhh, line 75: borning?)

src/averageAges.js
```diff
@@ -52,7 +70,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  } else { // every child borning ages
+    for (const child of people) {
+      const currentChildMotherName = child.mother;
+
+      for (const mother of people) {
+        if (mother.name === currentChildMotherName) {
+          birthAges.push(child.born - mother.born);
+        }
+      }
+    }
```
irenhh, line 94: use array methods

src/averageAges.js
Andrii256: they were cruel mothers

src/averageAges.js
Andrii256: (me)🔫

PR #121, opened 2020-08-12
first solution

krajnova:


src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  let men = people.filter(person => person.sex === 'm');
+
+  if (century !== undefined) {
```
irenhh, line 25:
```suggestion
  if (century) {
```

src/averageAges.js
```diff
@@ -52,7 +72,19 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const diff = people
+    .filter(person => {
+      return onlyWithSon
+        ? people.some(mother => person.mother === mother.name)
+          && person.sex === 'm'
+        : people.some(mother => person.mother === mother.name);
```
irenhh, line 80: you write the same iteration twice, could be assigned to a variable and used later

PR #128, opened 2020-08-12
Solution

rmnkk:


src/averageAges.js
```diff
@@ -52,7 +71,18 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(child => onlyWithSon
+    ? people.find(mother => mother.name === child.mother) && child.sex === 'm'
+    : people.find(mother => mother.name === child.mother)
+  );
+
+  const ageDiff = children.map(child => {
+    return child.born - people.find(mother => {
```
irenhh, line 80: what if the result of find is undefined?

rmnkk: There will be an error: TypeError: Cannot read property 'born' of undefined.
Do you mean that I should change code logic here?

src/averageAges.js
```diff
@@ -76,11 +76,14 @@ function calculateAverageAgeDiff(people, onlyWithSon) {
...
-    }).born;
-  });
+  const mothers = children.map(child => people
+    .find(mother => mother.name === child.mother));
+
+  const motherCheck = mothers.every(mother => mother !== undefined);
+
+  const ageDiff = motherCheck
+    ? children.map((child, i) => child.born - mothers[i].born)
+    : [0];
```
irenhh, line 86: zero of what?

irenhh, line 86: your prev code could be written like this
```
const ageDiff = children.map(child => {
  const mother = people.find(mother => mother.name === child.mother);

  if (mother) {
  }
});
```

PR #129, opened 2020-08-12
solution

varseniuk:


src/averageAges.js
```diff
@@ -1,58 +1,45 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person =>
+    century
+      ? Math.ceil(person.died / 100) === century && person.sex === 'm'
+      : (person.sex === 'm')
```
irenhh, line 7: there is no need to wrap it in parentheses

PR #137, opened 2020-08-12
first solution

leonbohdan:


src/averageAges.js
```diff
@@ -20,12 +20,28 @@ function calculateMenAverageAge(people, century) {
...
+  let menAges;
+
+  century
+    ? menAges = people
+      .filter(({ sex }) => sex === 'm')
+      .filter(({ died }) => Math.ceil(died / 100) === century)
+      .map(({ born, died }) => died - born)
+    : menAges = people
+      .filter(({ sex }) => sex === 'm')
+      .map(({ born, died }) => died - born);
```
irenhh, line 33: nice job with destructuring 👍 
you do the same filtering twice

src/averageAges.js
```diff
@@ -20,12 +20,28 @@ function calculateMenAverageAge(people, century) {
...
+
+  const menAverage = menAges
```
irenhh, line 35: this could be returned directly

src/averageAges.js
```diff
@@ -34,7 +50,20 @@ function calculateMenAverageAge(people, century) {
...
+  const women = people
+    .filter(({ sex }) => sex === 'f');
+
+  const womenAges = withChildren ? mothers : women;
+
+  const womenAverage = womenAges
+    .map(({ born, died }) => died - born)
+    .reduce((prev, next) => prev + next, 0) / womenAges.length;
+
+  return womenAverage;
```
irenhh, line 66: the same

src/averageAges.js
leonbohdan: rewritten

src/averageAges.js
leonbohdan: rewritten to return directly

src/averageAges.js
leonbohdan: rewritten

PR #229, opened 2020-12-17
solution

dtsehelnyk:


troospieler (approved):
Superb!

src/averageAges.js
```diff
@@ -15,6 +15,18 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  const menArr = people.filter(person => person.sex === 'm');
+  let resultMen;
+
+  century
+    ? resultMen = menArr.filter(man => Math.ceil(man.died / 100) === century)
+    : resultMen = menArr;
```
varseniuk, line 23:
```suggestion
  const resultMen = people.filter(person =>
    century
      ? Math.ceil(person.died / 100) === century && person.sex === 'm'
      : (person.sex === 'm')
  );
```

in such a way you can use 'filter' only once instead of two times

varseniuk (approved):
Good job! 
Take my comment into consideration for future

PR #231, opened 2020-12-18
Solution

vorfolomeevav:


WhiteRhin0 (approved):
d=(~_^)Z 

Nice job!

src/averageAges.js
```diff
@@ -52,7 +68,24 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      if (mother) {
+        acc.totalAge += person.born - mother.born;
+        acc.mothersCount += 1;
+      }
+    }
+
+    return acc;
+  }, {
+    mothersCount: 0, totalAge: 0,
+  });
```
varseniuk, line 86: You have chosen a little bit long and complicated way to solve this part of the task.
A more simple variant (and without any 'if-statements' and counters) would be to create an array of children according to 'onlyWithSon' argument using method `.some` inside '.filter'.
Then `.map` this array to find the age difference between mother and child (finding the child's mother using `.find` method inside `.map`.
And then just calculate the average difference of their ages using our new array and its length.

src/averageAges.js
```diff
@@ -34,7 +38,19 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothers = people
+    .map((person) => person.mother)
+    .filter((name) => !!name);
+
+  const women = people
+    .filter((person) => withChildren
+      ? person.sex === 'f' && mothers.includes(person.name)
+      : person.sex === 'f');
```
varseniuk, line 48:
```suggestion
  const women = people.filter(person =>
    withChildren
      ? people.some(child => child.mother === person.name) && person.sex === 'f'
      : person.sex === 'f'
  );
```
You could use the method `.some` inside your method `.filter` and avoid creating another variable 'mother'. It would make your code more simple and shorter

varseniuk (approved):
The solution is fine but you could do it much simpler.
Read my comments and try to make changes according to them.
It will be useful for you

PR #233, opened 2020-12-18
Completed

Serhii-Alieksieievych:


vovador (approved):
good job!)

src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let persons;
+
+  century === undefined
+    ? persons = people.filter((n) => n.sex === 'm')
+    : persons = people.filter((person) => person.sex === 'm'
+      && Math.ceil(person.died / 100) === century);
```
vsm2k17, line 23: use the ternary operator to decide which condition use to filter the array, don't repeat this part of the code
```persons = people.filter((person => )```

File: src/averageAges.js [hunk 2]
```diff
@@ -34,7 +39,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothers = people.map((n) => n.mother);
+  let persons;
```
vsm2k17, line 43: assign filtered array here no need to create variables like this and use const

src/averageAges.js
```diff
@@ -52,7 +66,43 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    ? diferences
+    = fullInfoValidPeopleWithMothersInList.map(child => child.born
+      - child.mothersBirthYear)
+    : diferences
+    = fullInfoValidMansWithMothersInList.map(child => child.born
+      - child.mothersBirthYear);
+
+  const sum = diferences.reduce((s, difference) => s + difference, 0);
+
+  return sum / diferences.length;
```
vsm2k17, line 105: that's too many strings of code, it should be shorter, I'll give you a hint:
split this task into blocks and implement them one by one:
1) create an array with kids who have a mother (use ternary operator for deciding which condition choose don't repeat your code)
2) create an array with a difference between a child "born" field and mother "born" field, which equal to the age of the mother when she born her child 
3)and then just find the average age from the previous array

vsm2k17 (changes requested):

vsm2k17 (approved):

PR #234, opened 2020-12-19
solution

YuriiFitsyk:


Roman-Matsuk (approved):
Great!

src/averageAges.js
```diff
@@ -52,7 +77,23 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const mom = people.filter(person => person.sex === 'f'
+    && people.some(x => x.mother === person.name));
+  let child;
```
vsm2k17, line 82: just assign filtered array here and use different condition for filter method depend on onlyWithSon parameter

src/averageAges.js
```diff
@@ -52,7 +77,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+        && mom.some(x => x.name === person.mother));
+    }
+  } else {
+    child = people.filter(person => mom.some(x => x.name === person.mother));
+  }
+
+  return +(child.reduce((sum, person) => sum
+    + person.born
+    - mom.find(mother => mother.name === person.mother).born, 0
+  ) / child.length).toFixed(2);
```
vsm2k17, line 96:
```suggestion
  return +(child.reduce((sum, person) => (
    sum + person.born - mom.find(mother => (
      mother.name === person.mother
    )).born,
    0,
  )) / child.length).toFixed(2);
```

vsm2k17: and toFixed could be removed

vsm2k17 (changes requested):

vsm2k17 (approved):
good job

PR #235, opened 2020-12-20
solution

xipholena:


HaidaiAndrii (approved):

src/averageAges.js
```diff
@@ -15,11 +15,27 @@
...
+      && Math.ceil(person.died / 100) === century) {
+      counter++;
+
+      return sum + (person.died - person.born);
+    }
+
+    return sum;
+  }, 0);
+
+  return totalAge / counter;
```
varseniuk, line 38: Interesting solution and use of 'reduce'.
But it's very complicated.
Can you do it in a more simple way?
You can just 'filter' the incoming 'people' array getting the required list of men.
After that, you need to 'map' this array to get the array of 'ages'
And last of all - calculate average ages and use the length of the array instead of the 'counter'.

src/averageAges.js
```diff
@@ -34,7 +50,26 @@ function calculateMenAverageAge(people, century) {
...
+    }
+
+    if (person.sex === 'f' && withChildren === undefined) {
+      counter++;
+
+      return sum + (person.died - person.born);
+    };
+
+    return sum;
+  }, 0);
```
varseniuk, line 70: Same here. 'Filter' incoming array, then use 'some' and use the length of an array instead of 'counter'

src/averageAges.js
```diff
@@ -52,7 +87,41 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      count += children.length;
+
+      const diffAllChildren = children.reduce((sumAge, child) =>
+        sumAge + child.born - mother.born, 0) + sum;
+
+      return diffAllChildren;
+    }
+
+    return sum;
+  };
```
varseniuk, line 122: Here as well. Filter (with 'some' inside) => map (with 'find' inside) => reduce

varseniuk (changes requested):
Your solution is working. It's good. But it is a little bit complicated.
Our goal is to write clean, simple, and effective code.
Your solution contains 125 lines of code and can easily be replaced with 50.
Let's try to use 'filter' instead of unnecessary 'ifs' and 'counters' and make your code simple

src/averageAges.js
```diff
@@ -14,12 +14,18 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const man = people.filter((person) => person.sex === 'm'
+    && ((!century)
+      ? true
+      : Math.ceil(person.died / 100) === century)
+  );
```
varseniuk, line 23:
```suggestion
const men = people.filter(person =>
    !century
      ? (person.sex === 'm')
      : Math.ceil(person.died / 100) === century && person.sex === 'm'
  );
```

If you want to use a ternary operator and only one 'filter; then you should use it like this. Why would we need `? true` here? It's really hard to read and understand.
Use 'men' instead of 'man' (plural form man-men, woman-women)

src/averageAges.js
```diff
@@ -14,12 +14,18 @@
...
+  const averageAge = (man
+    .reduce((acc, person) => acc + (person.died - person.born), 0)
+  ) / man.length;
```
varseniuk, line 26:
```suggestion
  const averageAge = (man
    .reduce((acc, person) => 
      acc + (person.died - person.born), 0)) / man.length;
```

Don't translate arithmetical operators on new lines

src/averageAges.js
```diff
@@ -33,8 +39,19 @@ function calculateMenAverageAge(people, century) {
...
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const woman = people.filter(person => person.sex === 'f'
+  && (!withChildren)
+    ? true
+    : people.some(potentialChild => potentialChild.mother === person.name)
+  );
```
varseniuk, line 48:
```suggestion
  const women = people.filter(person =>
    !withChildren
      ? person.sex === 'f'
      : people.some(child => child.mother === person.name) && person.sex === 'f'
  );
```

same story as with 'men'

src/averageAges.js
```diff
@@ -33,8 +39,19 @@ function calculateMenAverageAge(people, century) {
...
+
+  const averageAge = (woman
+    .reduce((acc, person) => acc + (person.died - person.born), 0)
+  ) / woman.length;
```
varseniuk, line 52:
```suggestion
  const averageAge = (woman
    .reduce((acc, person) => 
      acc + (person.died - person.born), 0)) / woman.length;
```

src/averageAges.js
```diff
@@ -51,8 +68,24 @@ function calculateWomenAverageAge(people, withChildren) {
  *
  * @return {number}
  */
+
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const woman = people.filter(child =>
+    (!onlyWithSon)
```
varseniuk, line 74:
```suggestion
    !onlyWithSon 
```

no need of the parentheses

varseniuk (changes requested):

varseniuk (approved):
Awesome! Good job 👍

xipholena:
> Awesome! Good job 👍

Thanx! It wouldn't be that good withiout your advices

PR #236, opened 2020-12-20
Solution

vovador:


Serhii-Alieksieievych (approved):

src/averageAges.js
```diff
@@ -15,11 +15,23 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  function menFilter(el) {
+    return el.sex === 'm';
+  }
+
+  function deathFilter(el) {
```
vsm2k17, line 22: el, not the best name, that not even top 3

src/averageAges.js
```diff
@@ -15,11 +15,23 @@
...
+    const deathCentury = Math.ceil(el.died / 100);
+
+    return century ? deathCentury === century : el.died !== century;
+  }
+
+  const men = people.filter(menFilter).filter(deathFilter);
```
vsm2k17, line 28: no need to filter two times just use different callbacks when century specified and if it's doesn't

src/averageAges.js
```diff
@@ -15,11 +15,23 @@
...
+
+  function callback(acc, curV) {
```
vsm2k17, line 30: callback, why not function function then?

src/averageAges.js
```diff
@@ -34,7 +46,25 @@ function calculateMenAverageAge(people, century) {
...
+
+  function motherFilter(el) {
+    function isMother(element) {
+      return el.name === element.mother;
+    }
+
+    return withChildren ? people.find(isMother) : true;
+  }
+
+  const women = people.filter(womenFilter).filter(motherFilter);
```
vsm2k17, line 61: same mistake

src/averageAges.js
```diff
@@ -34,7 +46,25 @@ function calculateMenAverageAge(people, century) {
...
+
+  function callback(acc, curV) {
```
vsm2k17, line 63: try to use meaningful names

src/averageAges.js
```diff
@@ -52,7 +82,36 @@ function calculateWomenAverageAge(people, withChildren) {
...
+        : el.name === element.mother;
+    }
+
+    return people.find(isMother);
+  }
+
+  const mothers = people.filter(motherFilter);
+  let totalChildNumber = 0;
+
+  function callback(acc, curV) {
```
vsm2k17, line 98: bad name

vsm2k17: good job, but if you split logic from this "callback" you can simplify your code.
hint no need to look mothers just fin all children who have mothers

vsm2k17 (changes requested):

src/averageAges.js
```diff
@@ -34,7 +44,25 @@ function calculateMenAverageAge(people, century) {
...
+  function womenFilter(person) {
+    return person.sex === 'f';
+  }
+
+  function motherFilter(person) {
+    function isMother(child) {
+      return person.name === child.mother;
+    }
+
+    return withChildren ? people.find(isMother) : true;
```
vsm2k17, line 56:
```suggestion
    return people.find(isMother);
```

src/averageAges.js
```diff
@@ -34,7 +44,25 @@ function calculateMenAverageAge(people, century) {
...
+  }
+
+  const women = people.filter(withChildren ? motherFilter : womenFilter);
+
+  function agesAccumulator(acc, person) {
+    return acc + (person.died - person.born);
+  }
+
+  return (women.reduce(agesAccumulator, 0) / women.length);
```
vsm2k17, line 65:
```suggestion
  return women.reduce(agesAccumulator, 0) / women.length;
```

vsm2k17 (approved):
good job

PR #242, opened 2020-12-21
Solution

AlieksieienkoAndriy:


deelray (approved):

src/averageAges.js
```diff
@@ -34,7 +49,25 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let womenWithChildren = [];
```
vsm2k17, line 52: the same mistake as in the previous function, no need to use an empty array, just assign filtered array.

src/averageAges.js
```diff
@@ -34,7 +49,25 @@ function calculateMenAverageAge(people, century) {
...
+    const age = woman.died - woman.born;
+
+    return sum + age;
+  }
+
+  const summ = womenWithChildren.reduce(countSum, 0);
+
+  const averageAge = summ / womenWithChildren.length;
+
+  return Math.round(averageAge * 100) / 100;
```
vsm2k17, line 70: for what you use Math.round here?

src/averageAges.js
```diff
@@ -52,7 +85,24 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    const mother = people.find((mom) => mom.name === child.mother);
+
+    const diff = child.born - mother.born;
+
+    return prev + diff;
+  }
+
+  const averageDiff = sum / children.length;
+
+  return Math.round(averageDiff * 100) / 100;
```
vsm2k17, line 105: isn't enough just return averageDiff?

vsm2k17 (changes requested):

vsm2k17 (approved):
good job

PR #243, opened 2020-12-21
Solution

olesiatk:


maryna-zinchenko (approved):

src/averageAges.js
```diff
@@ -34,7 +44,17 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const peopleWoman = people.filter(person => person.sex === 'f');
+
+  const peopleWomanAges = (withChildren)
```
vsm2k17, line 49:
```suggestion
  const womenAges = (withChildren)
```
we know that women are people so we can skip this

vsm2k17 (approved):

PR #252, opened 2020-12-21
solution

dima-bond:


khrulenko (approved):

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const allMaleList = people.filter(person => person.sex === 'm');
+  let maleList;
```
vsm2k17, line 19: just assign filtered list in this line no need to use let use const instead

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const allMaleList = people.filter(person => person.sex === 'm');
```
vsm2k17, line 18: this array is redundant use one array which is a filtered array of people, but use different conditions for the filter method.

src/averageAges.js
```diff
@@ -34,7 +42,19 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const allFemaleList = people.filter(person => person.sex === 'f');
+  let femaleList;
```
vsm2k17, line 46: same mistakes

src/averageAges.js
```diff
@@ -52,7 +72,23 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people
+    .filter(child => people
+      .some(mother => child.mother === mother.name));
+
+  let childrenList;
```
vsm2k17, line 79: use const

vsm2k17 (changes requested):
good job couple of fixes left

vsm2k17 (approved):
good job

PR #254, opened 2020-12-21
Solution

khrulenko:


dima-bond (approved):

vsm2k17 (approved):
good job, but don't use letters as names. (x)

PR #347, opened 2021-03-26
Solution

PeterShp:


src/averageAges.js
```diff
@@ -36,8 +44,20 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
-function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+function calculateWomenAverageAge(people, withChildren = false) {
```
vkryvytskyy, line 47:
```suggestion
function calculateWomenAverageAge(people, withChildren) {
```
it would be undefined, if not passed

src/averageAges.js
```diff
@@ -36,8 +44,20 @@ function calculateMenAverageAge(people, century) {
...
+  let amount = 0;
+  const result = people.reduce((summ, person) => {
+    if (person.sex === 'f'
+    && (!withChildren || people.find(el => el.mother === person.name))) {
+      amount++;
+
+      return summ + person.died - person.born;
```
vkryvytskyy, line 54: use parentheses in calculations

src/averageAges.js
```diff
@@ -55,7 +75,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  let amount = 0;
+  const result = people.reduce((summ, person) => {
+    const temp = people.find(el => el.name === person.mother);
+
+    if (person.mother
+    && ((onlyWithSon && person.sex === 'm')
+    || (!onlyWithSon)) && temp) {
+      amount++;
```
vkryvytskyy, line 85: make it simple, it's hard to understand what is going on here
use something like that 
```
const children = people.filter(person =>
    onlyWithSon
      ? 
      : 
 
const ageDiff = children.map(child =>
    child.born - mother.born);
    
 return (ageDiffSum / ageDiff.length)
```

vkryvytskyy (changes requested):

PR #363, opened 2021-04-27
Solution

Burunduk003:


src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
...
+  const men = century
+    ? people.filter(person => person.sex === 'm'
+    && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === 'm');
+
+  const averageAge = men
+    .reduce((a, man) => a + (man.died - man.born), 0)
+    / men.length;
+
+  return averageAge;
```
Ihor-Karpyn, line 32: I think this const is redundant
```suggestion
 return men
    .reduce((a, man) => a + (man.died - man.born), 0)
    / men.length;
```

Ihor-Karpyn (approved):
Nice solution

src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const men = century
+    ? people.filter(person => person.sex === 'm'
+    && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === 'm');
```
mykola-yarchuk, line 26:
```suggestion
  const men = people.filter(person => century
    ?  person.sex === 'm' && Math.ceil(person.died / 100) === century
    : person.sex === 'm';)
```

mykola-yarchuk: another way to write this

src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
...
+
+  const averageAge = men
+    .reduce((a, man) => a + (man.died - man.born), 0)
+    / men.length;
```
mykola-yarchuk, line 30:
```suggestion
  const averageAge = men.reduce((a, man) => (
    a + (man.died - man.born), 0
  )) / men.length;
```

src/averageAges.js
```diff
@@ -56,6 +76,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const mothers = people.filter(person => person.sex === 'f'
+    && people.some(women => women.mother === person.name));
+
+  const children = onlyWithSon ? people
+    .filter(person => mothers.some(mother => mother.name === person.mother)
+    && person.sex === 'm') : people
+    .filter(person => mothers.some(mother => mother.name === person.mother));
+
+  const ageDifference = children.reduce((a, child) => a
```
mykola-yarchuk, line 88: `a` that's a bad name for the variable

mykola-yarchuk: use meaningful names

mykola-yarchuk (changes requested):
almost done

src/averageAges.js
Burunduk003: "a" is a short way to write accumulator as we use "i" for index, but maybe you got a point.

mykola-yarchuk: `i` inside `for` loop that's an exception, when you use methods like `reduce` you want to know what it is, so it's better to write accumulator if you mean accumulator

src/averageAges.js
```diff
@@ -56,6 +74,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const children = people.filter(person => onlyWithSon
+    ? mothers.some(mother => mother.name === person.mother)
+    && person.sex === 'm'
+    : mothers.some(mother => mother.name === person.mother));
+
+  const ageDifference = children.reduce((totalAgeDiff, child) =>
+    totalAgeDiff + (child.born - (mothers.find(mother => mother.name
+    === child.mother)).born), 0
+  ) / children.length;
```
mykola-yarchuk, line 89:
```suggestion
  const ageDifference = children.reduce(
    (totalAgeDiff, child) => (
      totalAgeDiff + (child.born - (mothers.find(mother => (
        mother.name=== child.mother
      ))).born)
    ), 
    0,
  ) / children.length;
```

mykola-yarchuk: if I would writing this, I'd follow some rules
- if the parameter of the function very long, I would use a new line for every parameter
- if I want to return immediately something from the arrow function I open brace `(` and start from the new line (don't forget to close brace `)`)

mykola-yarchuk: also would be nice to save somewhere the result of the find method because, if you got `undefined` you can't use it in math operations, if you save it somewhere you can check if the result is a number before using it

mykola-yarchuk: and try to avoid so many nested methods, it's hard to read and very hard to debug

mykola-yarchuk (approved):

PR #367, opened 2021-04-27
Solution

Taras-Konovalov:


src/averageAges.js
```diff
@@ -55,7 +67,30 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const women = [];
+  const withSon = [];
+
+  people.filter(child =>
+    people.filter(woman => {
+      if (child.mother === woman.name) {
+        women.push(child.born - woman.born);
+      }
```
iryna-vys, line 77: I think we shouldn't use 'if' in this task.

src/averageAges.js
```diff
@@ -55,7 +67,30 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    })
+  );
+
+  people.filter(child =>
+    people.filter(woman => {
+      if (child.mother === woman.name && child.sex === 'm') {
+        withSon.push(child.born - woman.born);
+      }
```
iryna-vys, line 85: The same.

src/averageAges.js
```diff
@@ -55,7 +67,30 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const women = [];
+  const withSon = [];
+
+  people.filter(child =>
```
iryna-vys, line 73: You don't use the results of the filter method, so I think you can use forEach in these cases.

iryna-vys (approved):
Good job!

src/averageAges.js
mykola-yarchuk: згоден, нема сенсу використовувати filter якщо не зберігаєш результат, але forEach тут теж не потрбіний

VinnieJ-2k20 (approved):
Good job 👍

PR #372, opened 2021-04-30
Solution

iryna-vys:


progmari (approved):
👍

src/averageAges.js
```diff
@@ -55,7 +76,29 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      (child) => child.sex === 'm' && child.mother === person.name)
+  );
+  const sons = children.filter(
+    (person) => person.sex === 'm');
+  const avarageDifference = (onlyWithSon)
+    ? sons.reduce(
+      (acc, son) => acc + son.born
+      - mothersWithSon.find((mother) => son.mother === mother.name).born, 0)
+      / sons.length
+    : children.reduce(
```
varseniuk, line 96: Don't repeat yourself. Try to find a solution where you need only one `reduce` method.
Hint - you can create an array of `children` and use it with reduce. This array should depend on `onlyWithSon` argument. You can use a ternary operator for this purpose (like I've shown in examples above)

varseniuk (changes requested):
Nice start and good logic, but let's try to make the code cleaner. You are repeating yourself with `reduce` method (duplicating the method).
Pay attention to my comments and try to find better solution

src/averageAges.js
iryna-vys: Thank you for such detailed explanations.

aionova (approved):
LGTM!

PR #383, opened 2021-05-01
Develop

vova-myronenko:
first when I started coding, destructuring looked like a very good idea for me but after I finished, it looks short but not so obvious and clear. So what do you think?

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -75,9 +75,9 @@ function calculateAverageAgeDiff(people, onlyWithSon) {
     .filter(({ mother, sex }) => onlyWithSon
       ? sex === 'm'
         && people
-          .find(({ name }) => name === mother)
+          .some(({ name }) => name === mother)
       : people
-        .find(({ name }) => name === mother));
+        .some(({ name }) => name === mother));
 
   const averageAgeDiff = children
     .reduce((sum, { born: childBirth, mother }) => sum + (childBirth - people
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -1,5 +1,7 @@
 'use strict';
 
+// const { map } = require('./people');
+
 /**
  * Implement calculateMenAverageAge function
  *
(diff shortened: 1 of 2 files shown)


Andriykooo (approved):
I will agree with you that it is shorter but if we have an array with cats or dogs, it would be easier for me to understand to what object we address to get what we need, for example <person.name>, <cat.name> and others.

vova-myronenko:
> I will agree with you that it is shorter but if we have an array with cats or dogs, it would be easier for me to understand to what object we address to get what we need, for example <person.name>, <cat.name> and others.

Ok, thanks :)

src/averageAges.js
```diff
@@ -37,7 +44,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people
+    .filter(({ name, sex }) => withChildren
+      ? people
+        .some(({ mother }) => mother === name)
```
VinnieJ-2k20, line 50: It is a smart solution.
But it will fail if we are going to have some man with the same name - theoretically :)
You still have to have `sex === 'f'` filter in place.

src/averageAges.js
```diff
@@ -55,7 +71,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    .filter(({ mother, sex }) => onlyWithSon
+      ? sex === 'm'
+        && people
+          .some(({ name }) => name === mother)
+      : people
+        .some(({ name }) => name === mother));
+
+  const averageAgeDiff = children
+    .reduce((sum, { born: childBirth, mother }) => sum + (childBirth - people
+      .find(({ name }) => name === mother).born), 0) / children.length;
```
VinnieJ-2k20, line 84: Use separate variable for `mother` reference.

VinnieJ-2k20 (changes requested):
Nice solution, but couple of fixes needed.

VinnieJ-2k20 (approved):

PR #387, opened 2021-05-02
solution

den-bulaev:


src/averageAges.js
```diff
@@ -20,6 +20,25 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const isCenturySpecified = arguments.length > 1;
```
aliceblunt8, line 23: no need for this variable, you can use just century

src/averageAges.js
```diff
@@ -56,6 +99,41 @@ function calculateWomenAverageAge(people, withChildren) {
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  const isOnlySonsSpecified = arguments.length > 1;
```
aliceblunt8, line 102: just use onlyWithSon

src/averageAges.js
```diff
@@ -38,6 +57,30 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const mothersData = new Set(people.map(person => {
+    return person.mother;
+  }));
+  const hasChildren = arguments.length > 1;
```
aliceblunt8, line 63: just use withChildren

src/averageAges.js
```diff
@@ -20,6 +20,25 @@ function calculateMenAverageAge(people, century) {
...
+          ? [person.name, person.died - person.born]
+          : null;
+
+      case false:
+        return [person.name, person.died - person.born];
+    }
+  };
+
+  const maleData = people.filter((person) => person.sex === 'm').map(
+    callback).filter(person => person !== null);
```
aliceblunt8, line 38:
```suggestion
  const maleData = people
    .filter((person) => person.sex === 'm')
    .map(callback)
    .filter(Boolean);
```

src/averageAges.js
```diff
@@ -20,6 +20,25 @@ function calculateMenAverageAge(people, century) {
...
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const isCenturySpecified = arguments.length > 1;
+
+  const callback = (person) => {
+    switch (isCenturySpecified) {
+      case true:
+        return Math.ceil(person.died / 100) === century
+          ? [person.name, person.died - person.born]
+          : null;
```
aliceblunt8, line 30: use logical operator instead of returning null

```suggestion
        return Math.ceil(person.died / 100) === century
          && [person.name, person.died - person.born]
```

src/averageAges.js
```diff
@@ -20,6 +20,25 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const isCenturySpecified = arguments.length > 1;
+
+  const callback = (person) => {
+    switch (isCenturySpecified) {
```
aliceblunt8, line 26: in this case it's better to use if else, not switch

src/averageAges.js
```diff
@@ -20,6 +20,25 @@ function calculateMenAverageAge(people, century) {
...
+      case true:
+        return Math.ceil(person.died / 100) === century
+          ? [person.name, person.died - person.born]
+          : null;
+
+      case false:
+        return [person.name, person.died - person.born];
```
aliceblunt8, line 33: do you need person's name in this array? just return age

```suggestion
        return person.died - person.born;
```

src/averageAges.js
```diff
@@ -20,6 +20,25 @@ function calculateMenAverageAge(people, century) {
...
+    }
+  };
+
+  const maleData = people.filter((person) => person.sex === 'm').map(
+    callback).filter(person => person !== null);
+
+  return maleData.reduce(
+    (sum, person) => person[1] + sum, 0) / maleData.length;
```
aliceblunt8, line 41:
```suggestion
  return maleData.reduce(
    (sum, age) => age + sum, 0) / maleData.length;
```

src/averageAges.js
```diff
@@ -56,6 +99,41 @@ function calculateWomenAverageAge(people, withChildren) {
...
+          : prev;
+
+      case true:
+        return person.mother in mothersBirthdayData && person.sex === 'm'
+          ? [
+            ...prev,
+            [person.mother, mothersBirthdayData[person.mother], person.born],
+          ]
+
+          : prev;
```
aliceblunt8, line 130: this code is way too complicated
use filter to first find all mothers, then another filter to find all children (with onlyWithSon check)
then find average age diff

aliceblunt8 (changes requested):

mykola-yarchuk (approved):
great

PR #388, opened 2021-05-03
solution

vitalya2626:


aionova (changes requested):
Good, but let's think of the way to refactor it to make your code DRY!

aionova (approved):

PR #389, opened 2021-05-03
Solution

starzhitskiy:


aionova (approved):
Looks good, but can be refactored not to repeat code

PR #391, opened 2021-05-03
solution

Nazar-Kudatsky:


avgusta92 (approved):
Хорошая работа!)

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => {
+    if (century) {
+      return person.sex === 'm' && Math.ceil(person.died / 100) === century;
+    } else {
+      return person.sex === 'm';
+    }
```
VinnieJ-2k20, line 23: > replace `if ()` statement with &&, || or ?:

Let's follow the task guidelines.

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
+  });
+  const age = men.map(person => person.died - person.born);
+  const averageAge = age.reduce((a, b) => (a + b), 0) / age.length;
+
+  return averageAge;
```
VinnieJ-2k20, line 28: You don't need to create a variable if you return its value right away.
```suggestion
  return age.reduce((a, b) => (a + b), 0) / age.length;
```

VinnieJ-2k20 (changes requested):
Solution looks good, just need to fix couple of details.

src/averageAges.js
Nazar-Kudatsky: > replace `if ()` statement with &&, || or ?:

з тернарним заплутаюся, залишу так як знаю

VinnieJ-2k20 (approved):

PR #400, opened 2021-05-04
Average ages

leonid-vegera:


aionova (approved):
It's good, but could be refactored to make code DRY

PR #403, opened 2021-05-05
solution

vadym-liubych:


src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
+      .map(person => person.died - person.born)
+  );
+}
+
+function getCentury(person) {
+  return Math.ceil(person.died / 100);
+}
+
+function getAvarage(numbers) {
+  return numbers.reduce((a, b) => a + b, 0) / numbers.length;
```
dmytro-abram, line 31: Використовуй зрозумілі імена для змінних

dmytro-abram (approved):

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  return getAvarage(
+    people
+      .filter(person => person.sex === 'm')
+      .filter(man => !century || getCentury(man) === century)
+      .map(person => person.died - person.born)
+  );
+}
```
VinnieJ-2k20, line 24: Very nice job with reusing same function for three cases.
However, it is a bad practice to operform unnecessary complex operations inside fuction call.
Store `people` array processing results in a variable and then pass it to your fuction. 
Do it for all three cases.

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  return getAvarage(
```
VinnieJ-2k20, line 18: Pay attention to your English

```suggestion
  return getAverage(
```

VinnieJ-2k20 (changes requested):
Nice solution, but you need to fix some things before approve.

VinnieJ-2k20 (approved):
Good job 👍

PR #414, opened 2021-05-05
Solution

MinchukO:


src/averageAges.js
```diff
@@ -14,18 +19,34 @@
...
+  const arrAges = [];
+
+  if (century !== undefined) {
+    arrMan = people
+      .filter(person => person.sex === 'm'
+      && Math.ceil(person.died / 100) === century);
+  } else {
+    arrMan = people
+      .filter(person => person.sex === 'm');
+  }
```
VinnieJ-2k20, line 39: Let's follow task guidelines:

> replace `if ()` statement with &&, || or ?:

src/averageAges.js
```diff
@@ -56,6 +93,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    .map(child => [child, getMother(people, child)])
+    .filter(([, mother]) => mother)
+    .map(([child, mother]) => child.born - mother.born);
+
+  return getAvg(arrAges);
+}
+
+function getMother(people, child) {
+  return people.find(mother => mother.name === child.mother);
 }
```
VinnieJ-2k20, line 109: Good job with reusable functions 👍

src/averageAges.js
```diff
@@ -37,7 +58,23 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let arrWomen = [];
+
+  if (withChildren === true) {
+    arrWomen = people.filter(person => person.sex === 'f')
+      .filter(woman => getChildren(people, woman).length > 0);
```
VinnieJ-2k20, line 65: You just need to know if a woman has any children or not.
You really don't need to know number of children.
I think, `find` method would be better for `getChildren`, so it returns just `true` or `false`.

VinnieJ-2k20 (changes requested):
Almost done, just need to fix some details.

src/averageAges.js
```diff
@@ -36,8 +46,25 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let arrWomen = [];
+
+  if (withChildren === true) {
+    arrWomen = people.filter(person => person.sex === 'f')
+      .filter(woman => getChildren(people, woman));
+  } else {
+    arrWomen = people.filter(person => person.sex === 'f');
+  }
```
VinnieJ-2k20, line 58: This should also be written with ternary operator.

VinnieJ-2k20 (approved):

PR #507, opened 2021-08-06
Solution

alex-glotov67:


src/averageAges.js
```diff
@@ -20,6 +20,17 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const mens = people.filter(({ sex }) => sex === 'm');
+
+  const sortedMens = century
+
+    ? mens.filter(({ died }) => Math.ceil(died / 100) === century)
+
+    : mens;
```
illiarerun, line 29: I'm not really sure about these blank lines (same in all ternary operators throughout code)
```suggestion
  const sortedMens = century
    ? mens.filter(({ died }) => Math.ceil(died / 100) === century)
    : mens;
```

src/averageAges.js
```diff
@@ -38,6 +49,18 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const womens = people.filter(({ sex }) => sex === 'f');
```
illiarerun, line 52: That's actually interesting one, in English they don't have such word as womens. Plural for women will actually be women (pronounced like wimin). And 'mens' variable is also a bit wrong, it should be men. I'd suggest either remove [s] ending in both of these variables or rename them to females/males.

illiarerun (approved):
Good job 🔥  But check my comments)

src/averageAges.js
alex-glotov67: You're right! Thanks! I need to visit English classes more 😄

illiarerun: It's actually kinda hard to find free class to visit. All of them are filled up when i check 😄

src/averageAges.js
```diff
@@ -38,6 +47,16 @@ function calculateMenAverageAge(people, century) {
...
+  const women = people.filter(({ sex }) => sex === 'f');
+
+  const sortedWomen = withChildren
+    ? women.filter(({ name }) =>
+      people.some(({ mother }) => mother === name))
+    : women;
+
+  return sortedWomen.reduce((sumOfAges, { died, born }) => {
+    return sumOfAges + (died - born);
+  }, 0) / sortedWomen.length;
```
Denchepornyuk, line 59: такий reduce в тебе повторюється двічі, давай винесемо цей код в окрему функцію і будемо викликати її

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #508, opened 2021-08-06
The task has been done

Moroz-Dmytro:


alex-glotov67 (approved):
All great, but in description we shell not use if:
"replace `if ()` statement with &&, || or ?:"

Moroz-Dmytro:
> All great, but in description we shell not use if:
> "replace `if ()` statement with &&, || or ?:"

herkoss (approved):
Brilliantly!💎✨

PR #509, opened 2021-08-07
Solution

ivanchenko-kirill:


Moroz-Dmytro (approved):
Cool 👍

src/averageAges.js
```diff
@@ -37,7 +40,14 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const females = withChildren
+    ? people.filter(female =>
+      people.some(child => child.mother === female.name) && female.sex === 'f'
```
herkoss, line 45:
```suggestion
      people.some(child => child.mother === female.name)
```

herkoss: I think the second check is unnecessary, because if `child.mother === female.name`, variable female definitely is a woman and is the mother

src/averageAges.js
```diff
@@ -55,7 +65,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // write code here
+  const children = onlyWithSon
+    ? people.filter(child =>
+      child.sex === 'm' && people.some(mother => mother.name === child.mother)
+    )
+    : people.filter(child =>
+      people.some(mother => mother.name === child.mother)
+    );
+
+  children.map(child => {
```
herkoss, line 76:
```suggestion
  children.forEach(child => {
```

herkoss: Method `map` returns a new array, but you don't need the new array, you mutate the existing array

herkoss (changes requested):
Not bad, but there are some little mistakes)
Let's fix it)

herkoss (approved):
Excellent👍👌

PR #510, opened 2021-08-07
solution

SashaRudenko:


SashaLifashkin:
Хорошая работа!

src/averageAges.js
```diff
@@ -14,12 +14,21 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const getData = people.filter(person => person.sex === 'm');
+
+  const getDataWithCentury = people
+    .filter(person => Math.ceil(person.died / 100) === 18
+      && person.sex === 'm');
+
+  const getListOfMens = century !== 0 ? getDataWithCentury : getData;
```
VinnieJ-2k20, line 25:
```suggestion
  const listOfMen = century !== 0 ? getDataWithCentury : getData;
```

src/averageAges.js
```diff
@@ -14,12 +14,21 @@
  *
  * @return {number}
  */
-function calculateMenAverageAge(people, century) {
+function calculateMenAverageAge(people, century = 0) {
```
VinnieJ-2k20, line 17: No need to apply default 0 here, theoretically, century can be 0.
Later you can check if `century === undefined`

src/averageAges.js
```diff
@@ -36,8 +45,22 @@ function calculateMenAverageAge(people, century) {
...
-function calculateWomenAverageAge(people, withChildren) {
+function calculateWomenAverageAge(people, withChildren = 0) {
   // write code here
+  const getData = people.filter(person => person.sex === 'f');
+
+  const getDataWithChildren = people
+    .filter(person => people
+      .find(children => person.name === children.mother));
+
+  const getListOfWomans = withChildren !== 0 ? getDataWithChildren : getData;
```
VinnieJ-2k20, line 56: Use correct English
```suggestion
  const getListOfWomen = withChildren !== 0 ? getDataWithChildren : getData;
```

VinnieJ-2k20 (changes requested):
Solution looks good, but you need to rewrite it solution using correct variable names.
Also, think about how it can be simplified (try using array methods in chain instead of creating variable for each step).

src/averageAges.js
```diff
@@ -16,10 +16,20 @@
...
+  const personsFilteredWithCentury = personsFiltered
+    .filter(person => Math.ceil(person.died / 100) === century);
+
+  const listOfMens = century !== undefined
+    ? personsFilteredWithCentury
+    : personsFiltered;
+
+  const listOfAges = listOfMens.map(person => person.died - person.born);
+
+  const sumOfAges = listOfAges.reduce((prev, item) => prev + item, 0);
```
VinnieJ-2k20, line 30: Could use chained methods here, instead of creating extra variables.

VinnieJ-2k20 (approved):
I'll approve, but think about using chained array methods.

PR #515, opened 2021-08-09
Ages averaged

ihor-jpeg:


nastyyvolk (approved):
як на мене, все круто!
Сподобалась ідея з функціями-заготовками. На великому проекті таке явно би пригодилось. Самі обчислення з ними виглядають neat

src/averageAges.js
```diff
@@ -1,61 +1,61 @@
...
- * @return {number}
- */
-function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+const filterBySex = (people, sex) => {
+  return people.filter(x => x.sex === sex);
```
herkoss, line 4: Bad name for the variable. What is `x`? You need to give meaningful names to variables

src/averageAges.js
```diff
@@ -1,61 +1,61 @@
...
+};
+
+const getAverageAge = (people) => {
+  return people.reduce((x, y) => x + (y.died - y.born), 0) / people.length;
```
herkoss, line 8: What is `x` and `y`? I don’t understand

herkoss (changes requested):
Smart solution with custom functions!🧠
But it needs some refactoring. Details in comments
Good luck🍀🤞

src/averageAges.js
```diff
@@ -1,61 +1,67 @@
...
+  return people
+    .reduce((sumOfAges, person) =>
+      sumOfAges + (person.died - person.born), 0) / people.length;
+};
+
+const filterMothers = (people, women) => {
+  return women
+    .filter(mother => people
+      .map(daughter => daughter.mother).includes(mother.name));
+};
```
herkoss, line 17:
```suggestion
const filterMothers = (people, women) => {
  return women.filter(mother => (
    people.map(daughter => daughter.mother).includes(mother.name)
  ));
};
```

herkoss: You can read about code style here

mate-academy.github.io/style-guides

src/averageAges.js
```diff
@@ -1,61 +1,67 @@
...
-  // without nesting
+const filterBySex = (people, sex) => {
+  return people.filter(person => person.sex === sex);
+};
+
+const getAverageAge = (people) => {
+  return people
+    .reduce((sumOfAges, person) =>
+      sumOfAges + (person.died - person.born), 0) / people.length;
+};
```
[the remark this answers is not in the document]
herkoss: You can read about code style here

mate-academy.github.io/style-guides

src/averageAges.js
```diff
@@ -1,61 +1,67 @@
...
+
+const filterMothers = (people, women) => {
+  return women
+    .filter(mother => people
+      .map(daughter => daughter.mother).includes(mother.name));
```
herkoss, line 16: Why daughter? It can be son also, not only daughter.

src/averageAges.js
```diff
@@ -1,61 +1,67 @@
...
+
+  let children = filterChildren(people, mothers);
+
+  children = onlyWithSon === true
+    ? filterBySex(children, 'm')
+    : children;
+
+  const motherChildAgeDiff = parentChildAgeDiff(mothers, children);
+
+  return motherChildAgeDiff / children.length;
```
herkoss, line 64: Why children.length? mothers.length seems more logical

herkoss (changes requested):
Much better, but there are some little mistakes
Let's fix it💪

src/averageAges.js
ihor-jpeg: it does seem more logical. but the thing is that in this function we add the difference of ages of mother and their child / ren. and because one woman can have many children, but one child only one mother, the number of mothers and children is different. and in this case we need to use the number of children.

herkoss (approved):
Great!👍😊

PR #516, opened 2021-08-10
Average ages

nastyyvolk:


src/averageAges.js
```diff
@@ -37,8 +52,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
-}
+  const arrayToUse = withChildren
+    ? people.filter(x => people.some(({ mother }) => mother === x.name))
+    : people;
+
```
ihor-jpeg, line 58: круто ти тут з some зробила. також дуже сподобався фільтр по ключах, я щось до цього не додумався

ihor-jpeg (approved):
Хороша робота, як на мене. Молодець!

src/averageAges.js
```diff
@@ -20,6 +20,21 @@ function calculateMenAverageAge(people, century) {
...
+    = this.filter(x => Math.ceil(x.died/100) * 100 === century);
+  }
+  */
+  const arrayToUse = century
+    ? people.filter(({ died }) => Math.ceil(died / 100) === century)
+    : people;
+
+  const men = arrayToUse.filter(({ sex }) => sex === 'm');
+  const age = men.map(({ born, died }) => died - born);
+  const sumOfAge = age.reduce((prev, x) => prev + x);
```
VinnieJ-2k20, line 35: Don't create variable for each step. You can apply array methods in chain:
```
arrayToUse
  .filter(...)
  .map(...)
  .reduce(...)
```

VinnieJ-2k20 (changes requested):
Nice solution!
But rewrite it using chained methods.
Also, `arrayToUse` is not the best variable name.
Select name based on contents of this variable, not how you use it.

src/averageAges.js
```diff
@@ -37,8 +41,16 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
-}
+  const filteredPeople = withChildren
+    ? people.filter(x => people.some(({ mother }) => mother === x.name))
+    : people;
+
+  const womenAges = filteredPeople.filter(x => x.sex === 'f')
+    .map(({ born, died }) => died - born);
+  const sumOfAge = womenAges.reduce((prev, x) => prev + x);
```
VinnieJ-2k20, line 50: First - use correct indentations. If you chain several methods, each should start from new line;
Second - you can chain it with reduce too.

Same - in the other parts of task.
```suggestion
  const sumOfAges = filteredPeople
    .filter(x => x.sex === 'f')
    .map(({ born, died }) => died - born);
    .reduce((prev, x) => prev + x);
```

VinnieJ-2k20 (approved):
Good job 👍 
Need little bit more practice with chaining array methods.

src/averageAges.js
nastyyvolk: Actually, I need womenAges.length to calculate the average in the end, that's why I didn't chain reduce. I'm not sure if I can get this info(number of people) out of .reduced sum.

PR #534, opened 2021-08-13
Solution

poliandiya:


evgen1sider (approved):
awasome
looks better then my

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+
+  const menWithCentury = century
```
VinnieJ-2k20, line 20: Let's use more reliable check here because century may theoretically have valid falsy value (0)
```suggestion
  const menWithCentury = century !== undefined
```

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment.

PR #537, opened 2021-08-13
solution

derelicttt:


src/averageAges.js
```diff
@@ -14,12 +14,18 @@
  *
  * @return {number}
  */
+const middleAge = (people) => {
```
sasha-morozov, line 17: Nice usage of function!

sasha-morozov (approved):
Good job!

src/averageAges.js
VinnieJ-2k20, line 17: Clever decision.
However, function names should start with verb (action)
```suggestion
const getAverageAge = (people) => {
```

src/averageAges.js
```diff
@@ -14,12 +14,18 @@
...
+    count + (person.died - person.born), 0) / people.length;
+};
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => century
```
VinnieJ-2k20, line 23: Let's use more reliable check here, because century may theoretically have valid falsy value (0)
```suggestion
  const men = people.filter(person => century !== undefined
```

VinnieJ-2k20 (changes requested):
Almost done, let's make your code more sustainable.

VinnieJ-2k20 (approved):
Good job 👍

PR #541, opened 2021-08-13
solution

roman-hado:


Clonvhs (changes requested):
task was make without IF ELSE

src/averageAges.js
```diff
@@ -15,11 +15,21 @@
...
+    if (century !== undefined) {
+      if (Math.ceil(x.died / 100) === +century && x.sex === 'm') {
+        return true;
+      }
+
+      return false;
+    } else if (x.sex === 'm') {
+      return true;
+    }
+  });
```
VinnieJ-2k20, line 28: You need to learn how to use ternary operators.
> replace `if ()` statement with &&, || or ?:

src/averageAges.js
```diff
@@ -37,7 +47,17 @@ function calculateMenAverageAge(people, century) {
...
+  const womenWithChild = women.filter(woman => {
+    return people.some(child => child.mother === woman.name);
+  });
+  const totalAges = (prev, x) => prev + (x.died - x.born);
+
+  if (withChildren) {
+    return womenWithChild.reduce(totalAges, 0) / womenWithChild.length;
+  } else if (!withChildren) {
+    return women.reduce(totalAges, 0) / women.length;
+  }
```
VinnieJ-2k20, line 60: Same here.

src/averageAges.js
```diff
@@ -55,7 +75,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const arrayAges = [];
+  let children = people.filter(kid =>
+    people.some(mother => mother.name === kid.mother));
+
+  if (onlyWithSon) {
+    children = children.filter(son => son.sex === 'm');
+  }
```
VinnieJ-2k20, line 84: And here

VinnieJ-2k20 (changes requested):
You are on the right track, but need to restructure your code.
Next time, read the task guidelines more carefully.

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(x => x.sex === 'm');
+  const menFilter = century
```
VinnieJ-2k20, line 19: You need more reliable check here because century may theoretically have valid falsy value (0).
```suggestion
  const menFilter = century !== undefined
```

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
+    ? men.filter(x => (Math.ceil(x.died / 100) === +century))
+    : men;
+  const totalAges = (prev, x) => prev + (x.died - x.born);
```
VinnieJ-2k20, line 22: 1. If you create a function its name should start with verb (action). This one should be smth like `getTotalAges`
2. It is better to write this function inside reduce method because it is not used anywhere else.

src/averageAges.js
```diff
@@ -55,7 +68,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const findDiference = childrenFilter.map(kid =>
+    people.some(mother => {
+      if (mother.name === kid.mother) {
+        arrayAges.push(kid.born - mother.born);
+      }
+    })
+  );
+
+  const totalAges = (prev, x) => prev + x;
```
VinnieJ-2k20, line 87: Same as in previous comment.

VinnieJ-2k20 (changes requested):
Almost done. Let's fix some more details.

VinnieJ-2k20 (approved):

PR #547, opened 2021-08-15
done task

Uran-web:


RomSmile:
interesting solution in third task!

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const mans = people.filter(({ sex }) => (sex === 'm'));
```
varseniuk, line 18:
```suggestion
  const men = people.filter(({ sex }) => (sex === 'm'));
```
Use correct names

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
+
+  const sameCentury = mans.filter(({ died }) => (
+    Math.ceil(died / 100) === century
+  ));
+
+  const isCentury = (typeof century) === 'number' ? sameCentury : mans;
```
varseniuk, line 24:
```suggestion
```
Instead of creating so many variables you can filter `people` array in one place like
```
    century
      : (person.sex === 'm')
```

Here you just check if `century` is undefined or not.

src/averageAges.js
```diff
@@ -37,9 +45,23 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenAll = people.filter(({ sex }) => sex === 'f');
```
varseniuk, line 48: here same as with men

src/averageAges.js
```diff
@@ -55,7 +77,23 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const onlyChildren = people.filter(({ mother }) => (
```
varseniuk, line 80: And same here.
Use `onlyWithSon` argument for conditional filtering and make your code shorter

varseniuk (changes requested):
Logic is good, but it would be great not to create so many variable and make your code more simole

varseniuk (approved):
Good job

PR #549, opened 2021-08-15
task done

karolina-nad:


AndreySerhovetc (approved):

src/averageAges.js
```diff
@@ -20,6 +20,14 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const male = people.filter(person => {
+    return century
```
VinnieJ-2k20, line 24: You need to use more reliable check here because century theoretically can have valid falsy value (0)
```suggestion
    return century !== undefined
```

VinnieJ-2k20 (changes requested):
Almost done.
Fix one small detail.

VinnieJ-2k20 (approved):

PR #558, opened 2021-08-17
solution

00-22-11:


src/averageAges.js
```diff
@@ -15,17 +15,33 @@
...
+      currentCentury = diedInThisCentury;
+    }
+
+    if (item.sex === 'm' && diedInThisCentury === currentCentury) {
+      counter++;
+
+      return prev + (item.died - item.born);
+    } else {
+      return prev;
+    }
```
DmytroMykoliv, line 35: You can return prev without else

if (item.sex === 'm' && diedInThisCentury === currentCentury) {
      counter++;

  return prev + (item.died - item.born);
} 

  return prev;

DmytroMykoliv (approved):
Looks good )

src/averageAges.js
```diff
@@ -15,17 +15,33 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let counter = 0;
+  let diedInThisCentury;
+  let currentCentury = century;
```
varseniuk, line 20:
```suggestion
```

Let's get rid of all these variables and simplify your code. You can filter (as you did with children) all men according to the `century` argument and then count age difference and reduce that array.

Like:

```
    century
      : (person.sex === 'm')
    )
```

src/averageAges.js
```diff
@@ -37,7 +53,27 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let counter = 0;
```
varseniuk, line 56: Same as with men

src/averageAges.js
```diff
@@ -55,7 +91,17 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = onlyWithSon
```
varseniuk, line 94: let's make with `men` and `women` like you did it here

varseniuk (changes requested):
Let's simplify your code a bit

Denchepornyuk (approved):

PR #586, opened 2021-09-22
solution 👩‍👧‍👧

caelum29:


AVronskyi (approved):
Interesting approach! Looks like good

src/averageAges.js
```diff
@@ -9,17 +9,33 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const allMen = people.filter(getGender('m'));
+  let menToCountAverageAge = [];
+
+  !century
+    ? menToCountAverageAge = allMen
+    : menToCountAverageAge = allMen.filter(({ died }) => {
+      return getCentury(died) === century;
+    });
```
aliceblunt8, line 36: let's use if, it looks more readable
```suggestion
  if (century){
    allMen = allMen.filter(({ died }) => getCentury(died) === century);
 }
```

src/averageAges.js
```diff
@@ -37,7 +53,19 @@ function calculateMenAverageAge(people, century) {
...
+  let womenToCountAverageAge = [];
+
+  withChildren
+    ? womenToCountAverageAge = allWomen
+      .filter(({ name }) => {
+        return people.some(({ mother }) => {
+          return name === mother;
+        });
+      })
+    : womenToCountAverageAge = allWomen;
```
aliceblunt8, line 66: same here, just rewrite the variable if there is withChildren

src/averageAges.js
```diff
@@ -9,17 +9,33 @@
...
+const getCentury = year => Math.floor((year + 99) / 100);
+const getAge = (born, died) => died - born;
+const getAverageAge = (arr) => {
+  return arr.reduce((years, { born, died }) => years
+      + getAge(born, died), 0)
+    / arr.length;
+};
+const getGender = (gender) => {
+  return ({ sex }) => sex === gender;
+};
```
aliceblunt8, line 26: so many helper functions ;0

src/averageAges.js
```diff
@@ -55,7 +83,36 @@ function calculateWomenAverageAge(people, withChildren) {
...
+        return arr.some(({ name }) => {
+          return name === mother;
+        });
+      })
+    : childrenToProcess = people
+      .filter(({ mother, sex }, _, arr) => {
+        return (arr.some(({ name }) => {
+          return name === mother;
+        })) && sex === 'm';
+      });
```
aliceblunt8, line 108: let's first find all children and if there is onlyWithSon, we will filter only boys from all the children

aliceblunt8 (changes requested):

src/averageAges.js
caelum29: в условии задачи было что иф низзя использовать

Ihor-Karpyn (approved):

PR #612, opened 2021-09-26
solution

krasnobokydima:


src/averageAges.js
```diff
@@ -15,11 +15,23 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const pAge = person => person.died - person.born;
+  const pSum = (x, y) => x + y;
```
Anna-Tkachenko, line 19: Let's create more understandable variable name
```suggestion
  const getPersonAge = person => person.died - person.born;
  const getSum = (x, y) => x + y;
```

src/averageAges.js
```diff
@@ -37,11 +49,23 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const pAge = person => person.died - person.born;
+  const pSum = (x, y) => x + y;
```
Anna-Tkachenko, line 53: Same here

src/averageAges.js
```diff
@@ -37,11 +49,23 @@ function calculateMenAverageAge(people, century) {
...
+  if (withChildren === undefined) {
+    const allAge = people.filter(x => x.sex === 'f').map(pAge);
+
+    return Math.round(allAge.reduce(pSum) / allAge.length * 100) / 100;
+  } else {
+    const allAge = people.filter(x => x.name
+      && people.some(y => y.mother === x.name)).map(pAge);
+
+    return Math.round(allAge.reduce(pSum) / allAge.length * 100) / 100;
+  }
```
Anna-Tkachenko, line 64: Don't repeat yourself: return Math.round(allAge.reduce(pSum) / allAge.length * 100) / 100;
Use ternary operator for defining allAge array (allAge = withChildren ? : ) and then call reduce and return result only once

src/averageAges.js
```diff
@@ -55,7 +79,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+    return Math.round(resultAll.reduce((a, b) => a + b)
+    / resultAll.length * 100) / 100;
+  } else {
+    const result = childrenBoys.map(x =>
+      x.born - mom.find(y => y.name === x.mother).born);
+
+    return Math.ceil(result.reduce((a, b) => a + b)
+    / result.length * 100) / 100;
+  }
```
Anna-Tkachenko, line 101: Same here, don't repeat yourself

Anna-Tkachenko (changes requested):

PR #629, opened 2021-10-15
solved

arielg0rl:


src/averageAges.js
```diff
@@ -14,12 +14,35 @@
...
+    maleAges = maleAges.map(x => x.died - x.born);
+
+    let averageAge = 0;
+
+    averageAge = maleAges.reduce((sum, x) => sum + x) / males.length;
+
+    return averageAge;
+  }
+
+  if (century === 18) {
```
cyb1ove, line 35: Hardcoding is not well. Tests should pass with any century value.

src/averageAges.js
```diff
@@ -14,12 +14,35 @@
...
+
+    let maleAges = [...males];
+
+    maleAges = maleAges.map(x => x.died - x.born);
+
+    let averageAge = 0;
+
+    averageAge = maleAges.reduce((sum, x) => sum + x) / males.length;
+
+    return averageAge;
```
cyb1ove, line 32: You can return value without extra variable.

src/averageAges.js
```diff
@@ -14,12 +14,35 @@
...
+  if (century === 18) {
+    males = people.filter(x => x.sex === 'm' && Math.ceil(x.died / 100) === 18);
+
+    let maleAges18 = [...males];
+
+    maleAges18 = maleAges18.map(x => x.died - x.born);
+
+    const average = maleAges18.reduce((sum, x) => sum + x) / males.length;
+
+    return average;
```
cyb1ove, line 44: Same here.

PR #637, opened 2021-10-18
Solution

Linetstg:


src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let averageAge = 0;
+  let died18 = [];
+
+  century === 18
+    ? died18 = people.filter(d => d.sex === 'm'
+&& Math.ceil(d.died / 100) === 18)
+    : died18 = people.filter(d => d.sex === 'm');
```
denys-anosov, line 24: What if century isn't 18, but 19? Then your filter will not behave in the right way, as it's programmed to detect men only born in 18 century. It should be able to filter out men from any century, if century is provided as a function argument.

```suggestion
  
  const men = people.filter(person => person.sex === 'm'); // here we have all men from people array
  
  const menFromCentury = men.filter(person => Math.ceil(person.died / 100) === century); // here we filter out all men, whose death date divided by 100 and rounded by Math.ceil() is equal to century, provided as a function argument.
  
  const menForCalculations = century ? menFromCentury : men; // here we choose what array will be used for further calculations, depending on if century was provided or not.

// Now you can use menForCalculations for your further operations.
```

src/averageAges.js
```diff
@@ -37,7 +43,17 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let averageAge = 0;
+  let womanWith = [];
+
+  withChildren === true
+    ? womanWith = people.filter(d => d.sex === 'f'
+&& people.find(person => person.mother === d.name))
+    : womanWith = people.filter(d => d.sex === 'f');
```
denys-anosov, line 52: It might be better to not use ternary operator on different lines, ideally it should be written in one line. To do so, you need to create separate variables: for women (filtered out of people array) and for women with children (filtered out of women array).
Then you can use ternary operator to decide which array to use for further operations ([women] or [women with children]), depending on if withChildren argument was provided.

For example, const arrayForCalculations = withChildren ? [array of women with children] : [array of women];

denys-anosov (changes requested):
Good job, but some parts need to be reviewed. Men average age function should work for any century provided, not only 18. 
Also, it's better to use ternary operator in one line, not in multiple lines;

PR #655, opened 2021-11-15
Solution

Vitaliy-0:


alex-kostomarov (approved):
wow, really clear and structured code

src/averageAges.js
```diff
@@ -35,9 +39,17 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const womans = people.filter(person =>
+    withChildren
+      ? people.find(child => child.mother === person.name)
+      : person.sex === 'f'
+  );
+
+  return womans.reduce((sumOfAge, { died, born }) => {
+    return sumOfAge + died - born;
+  }, 0) / womans.length;
```
ihor-jpeg, line 52: you are using almost an identical function for calculating the average age of men. you can group this logic

src/averageAges.js
```diff
@@ -55,7 +67,26 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const mothersWithChild = people.filter(({ mother, sex }) =>
+    onlyWithSon
+      ? hasMother(mother, people) && sex === 'm'
+      : hasMother(mother, people)
```
ihor-jpeg, line 73: you can move this logic inside of the hasMother function. make it accept third parameter and set it to false by default

ihor-jpeg (approved):
good work! Have a look at the comments.

YuriiHoliuk (changes requested):
Test

YuriiHoliuk (approved):

PR #673, opened 2021-11-17
init

junglq:


oleksiy-pochkayev (approved):

src/averageAges.js
```diff
@@ -54,8 +72,29 @@ function calculateWomenAverageAge(people, withChildren) {
  *
  * @return {number}
  */
+
+function asMother(people, child) {
+  return people.some(person => person.name === child.mother);
+}
+
+function motherBorn(people, mother) {
+  return people.find(person => person.name === mother).born;
```
ihor-jpeg, line 81: if the `person` does not exist, the `find` method returns `undefined`. `undefined.born` will trow an error.

ihor-jpeg (approved):
Good work. Let's fix a few things.

PR #685, opened 2021-11-18
average ages

lera-permiakova:


src/averageAges.js
```diff
@@ -54,8 +72,29 @@ function calculateWomenAverageAge(people, withChildren) {
  *
  * @return {number}
  */
+function findMother(people, child) {
+  return people.some(mother => mother.name === child.mother);
+}
+
+function motherBorn(people, mother) {
```
WorkAccCom, line 79: It will be better to name functions using verbs. For instance, if you will call this function getMotherBorn will be easier to understand a function's purpose.

src/averageAges.js
```diff
@@ -54,8 +72,29 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const mom = people.find(person => person.name === mother);
+
+  return mom.born;
+}
+
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const child = people.filter(person => (
```
WorkAccCom, line 86: name children for this constant would be more appropriate because the .filter() method will return an array of objects  — children instead of single one child.

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^0.9.7",
```
WorkAccCom, line 19: Please do not add useless files to your PR.
You can use **git add src/** command to add needed files
**git status** command — for checking file status

WorkAccCom (approved):
Good work, but, please pay attention to functions and variables naming.

src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredMen = people.filter(person => (
+    century
+      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+      : person.sex === 'm'
+  ));
+
+  const averegeAge = filteredMen.reduce((prev, { died, born }) =>
+    (prev + (died - born)), 0) / filteredMen.length;
```
aionova, line 25:
```suggestion
  const averegeAge = filteredMen.reduce((prev, { died, born }) => (
    prev + (died - born)
  ), 0) / filteredMen.length;
```

it's easier to get where reduce ends, if you write it this way

src/averageAges.js
```diff
@@ -36,8 +41,21 @@ function calculateMenAverageAge(people, century) {
...
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filteredWomen = people.filter(person => (
+    withChildren
+      ? person.sex === 'f' && checkMother(people, person)
+      : person.sex === 'f'
+  ));
+
+  const averegeAge = filteredWomen.reduce((prev, { died, born }) =>
```
aionova, line 55: same here

src/averageAges.js
aionova: agree! we name functions as verbs, e.g. 'getMotherBorn'

src/averageAges.js
aionova: agree. If you receive an array, you can't name it 'child', as it's not one child. It's children

src/averageAges.js
```diff
@@ -54,8 +72,29 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const child = people.filter(person => (
+    onlyWithSon
+      ? person.sex === 'm' && findMother(people, person)
+      : findMother(people, person)
+  ));
+
+  const diff = child.map(kid => kid.born - motherBorn(people, kid.mother));
```
aionova, line 92:
```suggestion
  const ageDifferences = child.map(kid => kid.born - motherBorn(people, kid.mother));
```
We name array in plural form

src/averageAges.js
```diff
@@ -54,8 +72,29 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const averageAgeDiff = diff.reduce((acc, year) =>
+    acc + year) / child.length;
+
```
aionova, line 96:
```suggestion
  const averageAgeDiff = diff.reduce((acc, year) => (
    acc + year
  )) / child.length;

```

aionova (changes requested):

aionova (approved):
GJ!

PR #709, opened 2021-12-15
Solution

DariaVeretyak:


src/averageAges.js
```diff
@@ -15,11 +15,27 @@
...
+    const startCenture = century - 1 + '01';
+    const endCenture = century + '00';
+    const centureResultArr = men.filter(pers =>
+      pers.died > startCenture && pers.died < endCenture
+    );
+    const ageOfDeath = centureResultArr.map(pers => pers.died - pers.born);
+    const sumAge = ageOfDeath.reduce((sum, curent) => sum + curent, 0);
+
+    result = sumAge / centureResultArr.length;
+  }
```
aliceblunt8, line 36: let's remove if else with century
add this condition inside filter - find all men or find all men who died in certain venture when century is passed (you can use ternary operator)

src/averageAges.js
```diff
@@ -15,11 +15,27 @@
...
+  if (!century) {
+    const ageOfDeath = men.map(person => person.died - person.born);
+    const sumAge = ageOfDeath.reduce((sum, curent) => sum + curent, 0);
+
+    result = sumAge / men.length;
+  } else {
+    const startCenture = century - 1 + '01';
+    const endCenture = century + '00';
+    const centureResultArr = men.filter(pers =>
+      pers.died > startCenture && pers.died < endCenture
```
aliceblunt8, line 30: no need to find start and end century, you can use `Math.ceil(person.died / 100)` to find the person's century of death and then `Math.ceil(person.died / 100) === century` to compare them

src/averageAges.js
```diff
@@ -37,7 +53,26 @@ function calculateMenAverageAge(people, century) {
...
+  } else {
+    const mothers = people.map(x => x.mother);
+    const mothersWithChild = people.filter(x => mothers.includes(x.name));
+    const ageOfDeath = mothersWithChild.map(person =>
+      person.died - person.born
+    );
+    const sumAge = ageOfDeath.reduce((sum, curent) => sum + curent, 0);
+
+    result = sumAge / mothersWithChild.length;
+  }
```
aliceblunt8, line 73: same, remove if else, check withChildren inside filter where you filter women

src/averageAges.js
```diff
@@ -55,7 +90,31 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  let result = 0;
+  const children = people.filter(child => people.some(mother =>
+    mother.name === child.mother)
+  );
+
+  if (!onlyWithSon) {
```
aliceblunt8, line 98: same

aliceblunt8 (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,22 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+  const ageOfDeath = men.map(person => person.died - person.born);
+  const centureResultArr = men.filter(person =>
+    Math.ceil(person.died / 100) === century
+  );
```
aliceblunt8, line 22: combine men and centureResultArr filters into one
then you don't need to use map twice for both ageOfDeath and diedThisCentury, you will already have ONE array of needed men

src/averageAges.js
```diff
@@ -15,11 +15,22 @@
...
+  const diedThisCentury = centureResultArr.map(pers =>
+    pers.died - pers.born
+  );
+
+  const result = !century ? (
+    ageOfDeath.reduce((sum, age) => sum + age, 0) / men.length
+  ) : (
+    diedThisCentury.reduce((sum, age) => sum + age, 0) / centureResultArr.length
+  );
```
aliceblunt8, line 31: when you do like I said previously you won't need condition here, you'll need only one reduce

src/averageAges.js
```diff
@@ -37,7 +48,21 @@ function calculateMenAverageAge(people, century) {
...
+    person.died - person.born
+  );
+
+  const result = !withChildren ? (
+    ageOfDeath.reduce((sum, age) => sum + age, 0) / women.length
+  ) : (
+    ageOfDeathMom.reduce((sum, age) => sum + age, 0) / mothersWithChild.length
+  );
+
+  return result;
```
aliceblunt8, line 65: I think your solution looks better now, but it still overcomplicated, so here's an alternative solution, maybe it will be useful if you go step by step and analyze it for the future ;)

```
  let women = people.filter(person => person.sex === 'f');

  if (withChildren) {
    women = women.filter(woman =>
      people.some(person => person.mother === woman.name));
  }

  return women.reduce((sum, woman) =>
    sum + (woman.died - woman.born), 0
  ) / women.length;

```

aliceblunt8 (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let men = people.filter(person => person.sex === 'm');
+
+  if (century) {
+    men = men.filter(person => Math.ceil(person.died / 100) === century);
+  };
```
vkryvytskyy, line 22: Это можно объединить, один filter с условием внутри, которое определяет по какому параметру ты фильтруешь

vkryvytskyy (changes requested):

aliceblunt8 (approved):

PR #723, opened 2021-12-22
v1

makssisua:


src/averageAges.js
```diff
@@ -7,20 +7,24 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
-}
+  const isCenturyFilter = people.filter(kid => (century
+    ? kid.sex === 'm' && Math.ceil(kid.died / 100) === century
+    : kid.sex === 'm'));
```
Elena-brv, line 20: you check 'century' on every iteration. You can check the 'century' only 1 time and choose what filter to apply (the same in next functions)

src/averageAges.js
```diff
@@ -7,20 +7,24 @@
...
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
-}
+  const isCenturyFilter = people.filter(kid => (century
```
Elena-brv, line 18: and it's not a filter, it's already filteredPeople

Elena-brv (changes requested):

src/averageAges.js
```diff
@@ -37,7 +40,15 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const isHaveChildFiltered = people.filter(kid => (withChildren
+    ? kid.sex === 'f' && people.some(child => child.mother === kid.name)
+    : kid.sex === 'f'));
+
+  const age = isHaveChildFiltered.map(woman => woman.died - woman.born);
+
+  const womenAverageAge = age.reduce((a, b) => a + b) / age.length;
+
+  return womenAverageAge;
```
Olexandr6, line 51:
```suggestion
  return age.reduce((a, b) => a + b) / age.length;;
```

PR #732, opened 2021-12-22
Solution

marina-tilniak:


AlonaSnigur (approved):

src/averageAges.js
```diff
@@ -7,19 +7,33 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+  const menDiedInCentury = men
+    .filter(person => Math.ceil(person.died / 100) === century);
```
Elena-brv, line 34: now function does both of these filters. Better to check which filter to apply and only then to apply it (at all functions)

Elena-brv (changes requested):

Olexandr6 (approved):

PR #734, opened 2021-12-23
My solution

SavelovSerhii:


src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // without nesting
+  const men = people.filter((el) => century
+    ? el.sex === 'm' && Math.ceil(el.died / 100) === century : el.sex === 'm');
+
+  const len = men.length;
+  const ages = men.map(a => a.died - a.born);
+
+  const result = ages.reduce((a, b) => a + b) / len;
+
+  return result;
```
Olexandr6, line 26: You can return value immediately 
```suggestion
  return ages.reduce((a, b) => a + b) / len;;
```

src/averageAges.js
```diff
@@ -55,7 +67,22 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const women = people.filter(el =>
+    people.some(parent => el.name === parent.mother)
+    && el.sex === 'f');
+
+  const child = people.filter(el => onlyWithSon
```
Olexandr6, line 74:
```suggestion
  const children = people.filter(el => onlyWithSon
```

src/averageAges.js
```diff
@@ -55,7 +67,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    ? people.some(children => el.mother === children.name && el.sex === 'm')
+    : people.some(children => el.mother === children.name));
+
+  const diff = child.map(children =>
```
Olexandr6, line 78:
```suggestion
  const ageDifference = child.map(children =>
```

Olexandr6 (approved):

PR #738, opened 2021-12-23
Solution

Valeria-Valuyskaya:


Volodymyr1806 (approved):

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let menInArray = people.filter(person => person.sex === 'm');
+  const menDied = menInArray.filter(person => Math.ceil(person.died / 100)
+  === century);
+
+  menInArray = century
+    ? menDied
+    : menInArray;
```
Elena-brv, line 24: do you need to make 2 filters? you can check 'century' firstly and then decide which filter to apply (the same in next functions)

Elena-brv (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menInArray = century
```
Olexandr6, line 18: Avoid words `array, obj, str` etc. in var naming
```suggestion
  const men = century
```

Olexandr6 (approved):

PR #748, opened 2021-12-26
solution_v1.0

Rozdorozhnii:


Mikedacheka (approved):

src/averageAges.js
```diff
@@ -37,7 +43,12 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenHasCHildren = people.filter(person =>
```
Olexandr6, line 46: Not correct var name, cause women don't always have children in this case

Olexandr6, line 46:
```suggestion
  const women = people.filter(person =>
```

Olexandr6 (approved):
Good job 👍

PR #781, opened 2022-01-30
Dev

KostyaKovbel:
seems like true

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -15,11 +15,21 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const allMens = century
+    ? people.filter(person => century === Math.ceil(person.died / 100)
+      && person.sex === 'm')
+    : people.filter(person => person.sex === 'm');
+
+  // creating array with ages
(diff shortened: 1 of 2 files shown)


src/averageAges.js
```diff
@@ -15,11 +15,21 @@
...
+  const allMens = century
+    ? people.filter(person => century === Math.ceil(person.died / 100)
+      && person.sex === 'm')
+    : people.filter(person => person.sex === 'm');
+
+  // creating array with ages
+  const ages = [];
+
+  // pushing age of all mens
+  allMens.forEach(men => ages.push(men.died - men.born));
```
Anna-Tkachenko, line 27: You can use map here or calculate value of age in reduce method

src/averageAges.js
```diff
@@ -55,7 +75,30 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const childAgesSum = childrens
+    .map((child) => child.born)
+    .reduce((age, currentAge) => age + currentAge, 0);
+
+  const mothersAge = childrens.map((child) =>
+    mothers
+      .filter((mom) => mom.name === child.mother)
+      .map((age) => age.born)
+      .reduce((age, curent) => age + curent)
```
Anna-Tkachenko, line 97: Check all methods here, some of them are redundant
Don't create additional iterators
For example, you can calculate needed age value in reduce without map method

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #782, opened 2022-01-30
avg ages

Vadym-Mishchenko:


src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const male = (century === undefined)
+    ? people.filter(human => human.sex === 'm')
+    : people.filter(human => human.sex === 'm'
+      && century === Math.ceil(human.died / 100));
```
Anna-Tkachenko, line 21:
```suggestion
  const male = century
    ? if it's not undefined
    : if it's undefined;
```

src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
+
+  const sumOfFilteredMale = male.map(human => human.died - human.born)
+    .reduce((accumulator, age) => accumulator + age);
```
Anna-Tkachenko, line 24: You can calculate age value in reduce method

src/averageAges.js
```diff
@@ -37,7 +42,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const female = (withChildren === undefined)
+    ? people.filter(human => human.sex === 'f')
+    : people.filter(human => human.sex === 'f'
+      && people.find(child => child.mother === human.name));
```
Anna-Tkachenko, line 48: Same here with condition and code style

Anna-Tkachenko (approved):

PR #783, opened 2022-01-30
Solution

maxim2310:


src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
+  && (century
+    ? Math.ceil(person.died / 100) === century
+    : true
+  ));
+
+  const ages = man.map(x => x.died - x.born);
+
+  const sumOfAges = ages.reduce(
+    (prevValue, currentValue) => prevValue + currentValue
+  );
```
Anna-Tkachenko, line 28: Map is redundant here
You can calculate age value in reduce method

src/averageAges.js
```diff
@@ -37,7 +45,19 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filtered = people.filter(person => person.sex === 'f'
```
Anna-Tkachenko, line 48: Create more meaningful variable name instead of 'filtered'

Anna-Tkachenko (approved):

PR #789, opened 2022-01-31
age

pogrebnyakp:


src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
+  const male = people
+    .filter(person => person.sex === 'm'
+    && (century
+      ? Math.ceil(person.died / 100) === century
+      : true)
+    );
+
+  const menAverageAge = male
+    .map(person => person.died - person.born)
+    .reduce((currentValue, age) => currentValue + age);
```
Anna-Tkachenko, line 27: You can also calculate age value in reduce method instead of using map

src/averageAges.js
```diff
@@ -37,7 +44,18 @@ function calculateMenAverageAge(people, century) {
...
+  const female = people
+    .filter(person => person.sex === 'f'
+    && (withChildren
+      ? people.some(children => children.mother === person.name)
+      : true)
+    );
+
+  const womenAverageAge = female
+    .map(person => person.died - person.born)
+    .reduce((currentValue, age) => currentValue + age);
```
Anna-Tkachenko, line 56: Same here

Anna-Tkachenko (approved):

PR #795, opened 2022-01-31
Solution

cxrmx:


src/averageAges.js
```diff
@@ -37,7 +39,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const rightWomen = people.filter(person =>
```
Anna-Tkachenko, line 42: maybe womenWithChild will be better variable name? same with 'rightMen' and 'rightMothers'

Anna-Tkachenko (approved):

PR #798, opened 2022-01-31
Solution

DanyloLipar:


src/averageAges.js
```diff
@@ -15,11 +15,22 @@
...
+      .filter(man => !century || getCentury(man) === century)
+      .map(e => e.died - e.born)
+  );
+}
+
+function getCentury(person) {
+  return Math.ceil(person.died / 100);
+}
+
+function getAvarage(number) {
```
Anna-Tkachenko, line 30:
```suggestion
function getAverage(number) {
```

src/averageAges.js
```diff
@@ -15,11 +15,22 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  return getAvarage(
+    people
+      .filter(per => per.sex === 'm')
```
Anna-Tkachenko, line 20: it's better to set full variable name: person, element

Anna-Tkachenko (approved):

PR #813, opened 2022-02-15
solution for #averageAges

jstmpelowycz:


SanyaBratashchuk (approved):
good readable code. good job

src/averageAges.js
```diff
@@ -14,12 +14,15 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let men = getPeopleBySex(people, 'm');
+
+  if (century) {
```
solaryasha, line 21: Let's practice ternary and logical operators
```
replace `if ()` statement with &&, || or ?:

File: src/averageAges.js [hunk 2]
```diff
@@ -36,15 +39,24 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let women = getPeopleBySex(people, 'f');
+
+  if (withChildren) { // O(N^2) :(
```
solaryasha, line 46: Use hash map for avoiding n**2 complexity

You can implement hash map with `reduce` and `{}` as initial value

solaryasha (changes requested):

src/averageAges.js
solaryasha: It's not required, but it will be good practice for you.

solaryasha (approved):

src/averageAges.js
jstmpelowycz:
```s

jstmpelowycz: ''фф''

jstmpelowycz: *и*

PR #847, opened 2022-03-25
js-avarage-ages

xapg6acc:


marina-tilniak (approved):
you can move calculation of average value to separate function

PR #863, opened 2022-04-06
My solution

OlexandrIstomin:


src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
-  // without nesting
+  const array = century ? people.filter(element => element.sex === 'm'
+    && century === Math.ceil(element.died / 100))
+    : people.filter(element => element.sex === 'm');
+
+  const finalAge = century ? (array.map((element) =>
+    element.died - element.born).reduce((prev, next) => prev + next)
+    / array.length)
+    : (array.map((element) => element.died
+    - element.born).reduce((prev, next) => prev + next) / array.length);
```
Bohdan-Maksymiuk, line 26: do we need `?:` here
we filter our array previously and here you do the same logic

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #871, opened 2022-04-10
done

Yurii837:


src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let sumOfAges = 0;
+  let countOfMens = 0;
+
+  people.forEach(e => {
```
YegorKochetkov, line 21: uninformative name of variable - 'e'

src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
+    const ifCentury = (Math.ceil(e.died / 100) === century || !century);
+    const menAge = e.sex === 'm' && ifCentury ? e.died - e.born : 0;
+
+    sumOfAges += menAge;
+    countOfMens = menAge !== 0 ? countOfMens + 1 : countOfMens;
+  });
```
YegorKochetkov, line 27: wierd logic
wasn't it easier to get the right men use .filter()

YegorKochetkov (approved):
ok)

src/averageAges.js
AlinaMatuschak, line 21: avoid using loop and forEach

AlinaMatuschak (changes requested):

src/averageAges.js
Yurii837: I used one one loop. I think it's impossible to solve this task without loop.

YegorKochetkov: for example, you can use filter to get men. then map men to get ages

MaksimusCRS (approved):

PR #875, opened 2022-04-11
Solution

lem47:


HelenPetrovska (approved):
I like your option of solution!)

PR #897, opened 2022-04-13
getAvarageAges

ShevchyshynRoman:


src/averageAges.js
```diff
@@ -55,7 +62,18 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const women = people.filter(person =>
+    person.sex === 'f' && people.some(child => child.mother === person.name));
+
+  const children = people.filter(person => onlyWithSon
+    ? person.sex === 'm' && people.some(mother => person.mother === mother.name)
```
MariaBondar, line 69: sometimes it's better to use "if (), else" in solution, because it's very hard to read )

src/averageAges.js
```diff
@@ -55,7 +62,18 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const women = people.filter(person =>
+    person.sex === 'f' && people.some(child => child.mother === person.name));
```
AlinaMatuschak, line 66: If it's a mother, it's a woman too )

```suggestion
  const women = people.filter(person =>
    people.some(child => child.mother === person.name));
```

AlinaMatuschak (approved):

PR #912, opened 2022-04-29
Solution

59MrRobot:
I think my brain is dead... 🥲

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -15,16 +15,16 @@
  * @return {number}
  */
 
-const isMale = person => person.sex === 'm';
+const isMale = (person) => person.sex === 'm';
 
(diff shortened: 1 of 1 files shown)


LukaszGryziewicz (approved):
Awesome 🔥 🥇

src/averageAges.js
```diff
@@ -14,12 +14,29 @@
  *
  * @return {number}
  */
+
+const isMale = person => person.sex === 'm';
```
ihor-jpeg, line 18: let's use thes for beter visualization 
```suggestion
const isMale = (person) => person.sex === 'm';
```

ihor-jpeg (approved):

VinnieJ-2k20 (approved):

PR #913, opened 2022-04-29
solution

LukaszGryziewicz:


59MrRobot (approved):
Very clean code, well done! 👍

src/averageAges.js
```diff
@@ -1,5 +1,15 @@
 'use strict';
 
+const isMale = person => person.sex === 'm';
```
ihor-jpeg, line 3:
```suggestion
const isMale = (person) => person.sex === 'm';
```

src/averageAges.js
```diff
@@ -15,11 +25,17 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const calculateCentury = num => Math.ceil(num / 100);
```
ihor-jpeg, line 28: this function is alright, but it can do more. below you take the result of the function and compare it to the century and then there's a logical and. it's quite a lot.

src/averageAges.js
```diff
@@ -15,11 +25,17 @@
...
+
+  const men = people.filter(person =>
+    century
+      ? calculateCentury(person.died) === century && isMale(person)
```
ihor-jpeg, line 32: you can implement this `calculateCentury(person.died) === century` inside of the calculateCentury and return a boolean, just like isMale

src/averageAges.js
```diff
@@ -37,7 +53,17 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const isFemale = person => person.sex === 'f';
```
ihor-jpeg, line 56:
```suggestion
  const isFemale = (person) => person.sex === 'f';
```

ihor-jpeg (approved):

VinnieJ-2k20 (approved):

PR #914, opened 2022-05-01
SolutionI

EnjSadman:


src/averageAges.js
```diff
@@ -129,7 +129,54 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    }
+  }); */
+
+  switch (arguments.length) {
+    case 0: {
+      return 0;
+    }
+
+    case 1: {
+      people.forEach(e => {
```
vlasiuk-anatolii, line 153: avoid using loop and forEach

src/averageAges.js
```diff
@@ -129,7 +129,54 @@ function calculateWomenAverageAge(people, withChildren) {
...
+        const isMother = people.find(el => e.mother === el.name);
+
+        if (isMother !== undefined) {
```
vlasiuk-anatolii, line 156: replace `if ()` statement with &&, || or ?:

src/averageAges.js
```diff
@@ -129,7 +129,54 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      result = averageAgeDiff / count;
+
+      return result;
+    }
+
+    case 2: {
+      man.forEach(e => {
+        const isMother = people.find(el => e.mother === el.name);
+
+        if (isMother !== undefined) {
```
vlasiuk-anatolii, line 170: replace `if ()` statement with &&, || or ?:

vlasiuk-anatolii (commented):
Avoid using loop and forEach and replace `if ()` statement with &&, || or ?:.

Anna-Tkachenko (changes requested):
Switch / case in this task is redundant
You write the same logic two times for every function
Try to get the needed array of people at the start of each function according to the passed arguments
And then call reduce and get average age one time at the end

src/averageAges.js
```diff
@@ -20,6 +20,28 @@ function calculateMenAverageAge(people, century) {
...
+  const centuryMan = man.filter(
+    el => (Math.ceil(el.died / 100) === century)
+  );
+  let centuryManAverage = centuryMan.reduce(
+    (accum, curr) => accum + (curr.died - curr.born),
+    avAge);
+
+  centuryManAverage /= centuryMan.length;
+
+  (century === undefined) ? result = menAvAge : result = centuryManAverage;
```
Anna-Tkachenko, line 42: Don't repeat yourself: method reduce and calculating average age is the same for both of variants: will be better to check at the start of the function, if century is undefined - you should work with array of all men
if century is not undefined - you should work with array of men with correct century

src/averageAges.js
```diff
@@ -36,8 +58,35 @@ function calculateMenAverageAge(people, century) {
...
+
+  const mothers = woman.filter(isMother, woman);
+
+  let mothersAvAge = mothers.reduce(
+    (accum, curr) => accum + (curr.died - curr.born),
+    avAge);
+
+  mothersAvAge /= mothers.length;
+
+  (withChildren === undefined) ? result = womanAvAge : result = mothersAvAge;
```
Anna-Tkachenko, line 87: Same here, you should check the second argument at the start of the function
And create correct array of people depending on it

Anna-Tkachenko (changes requested):

src/averageAges.js
```diff
@@ -36,8 +52,29 @@ function calculateMenAverageAge(people, century) {
...
+
+    return motherCheck !== undefined;
+  }
+
+  let currentList;
+  const avAge = 0;
+  const woman = people.filter(el => (el.sex === 'f'));
+  const mothers = woman.filter(isMother, woman);
+
+  (withChildren === undefined) ? currentList = woman : currentList = mothers;
```
Denchepornyuk, line 69: the same here

src/averageAges.js
```diff
@@ -36,8 +52,29 @@ function calculateMenAverageAge(people, century) {
...
+
+  let womanAvAge = currentList.reduce(
+    (accum, curr) => accum + (curr.died - curr.born),
+    avAge);
```
Denchepornyuk, line 73: you have two identical reduce, create a function for them

src/averageAges.js
```diff
@@ -56,6 +93,40 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+    if (a !== undefined) {
+      return e;
+    }
+  });
+
+  const allSons = people.filter((e) => {
+    const a = people.find(el => (e.mother === el.name) && (e.sex === 'm'));
+
+    if (a !== undefined) {
```
Denchepornyuk, line 110:
```suggestion
    if (a) {
```

Denchepornyuk (changes requested):

Denchepornyuk (changes requested):
you didn't fix all comments

Denchepornyuk (approved):

PR #918, opened 2022-05-03
solution

ZUBOV-ILLIA:


src/averageAges.js
```diff
@@ -55,7 +74,32 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const chields = people.filter(x => {
+    return people.some(y => y.name === x.mother);
```
SavMary, line 78: You should use more semantic variables than x and y for better understanding

SavMary (approved):
check the names of variables: mans, womens, chields - are wrong

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #919, opened 2022-05-03
Develop

alexandra-protyanova:


vlad-demchuk (approved):
in my opinion, you should create more variables to understand your code better

src/averageAges.js
```diff
@@ -54,8 +68,20 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+    ? people.filter(person =>
+      people.some(mother => mother.name === person.mother)
+    && person.sex === 'm')
+
+    : people.filter(person =>
+      people.some(mother => mother.name === person.mother));
+
+  const differenceInAge = family.map(person =>
+    person.born - people.find(mother => mother.name === person.mother).born);
```
Anna-Tkachenko, line 82: In some cases, more variables it's good thing)
For example, in this function you can create separate variable for mother, and then use this variable in age calculation
It'll help to better code understanding)

Anna-Tkachenko (approved):
Well Done ⭐️

PR #922, opened 2022-05-04
solution

efdalyalcin:
I got brain damage from the task, it will leave a scar 😃

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -39,21 +39,6 @@ function calculateWomenAverageAge(people, withChildren) {
   return average;
 }
 
-/**
- * Implement calculateAverageAgeDiff function.
- *
- * The function returns an average age difference between a child and his or her
- * mother in the array. (A mother's age at child birth)
- *
- * If `onlyWithSon` is specified then function calculates age difference only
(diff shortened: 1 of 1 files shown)


src/averageAges.js
```diff
@@ -1,61 +1,64 @@
...
- * Hint: To check if a woman has children you should find someone who mention
- * her as mother.
- *
- * @param {object[]} people
- * @param {boolean} withChildren - optional
- *
- * @return {number}
- */
+const isFemale = person => person.sex === 'f';
+const isMother = (mother, child) => mother.name === child.mother;
```
rHashTag, line 25: Could I suggest grouping these 2 global variables with the other global variables at the top.

rHashTag (changes requested):
Well done, code looks great.!

I've added one suggestions.

rHashTag (approved):
Looks good to me!

VinnieJ-2k20 (approved):
Great job!
Really liked the thing with variables.
One small comment - you may need to consider what will happen in case of empty arrays. You gonna have NaN as a result.

PR #923, opened 2022-05-04
solution

yelnikov-andrii:


alexandra-protyanova (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,25 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person['sex'] === 'm');
```
Anna-Tkachenko, line 18: Don't repeat yourself: just check at the start of the function if century argument exists
if yes - you should create the array with men for correct century
if no - you should create the array with all men

and then logic for both of variants is the same: reducing and getting the average value
```suggestion
  const men = century
    ? people.filter(...)
    : people.filter(person => person['sex'] === 'm');
```

src/averageAges.js
```diff
@@ -37,7 +51,28 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(x => x['sex'] === 'f');
```
Anna-Tkachenko, line 54: Same here, use ternary operator, find correct array of people and call reduce for it
```suggestion
  const women = withChildren
    ? people.filter(...)
    : people.filter(x => x['sex'] === 'f');
```

Anna-Tkachenko (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,21 @@
...
-  // without nesting
+  let amountOfexistence = 0;
+
+  const men = century
+    ? people.filter(person => Math.ceil(person['died'] / 100)
+    === century && person['sex'] === 'm')
+
+    : people.filter(person => person['sex'] === 'm');
+
+  amountOfexistence = men.reduce((sum, manData) => (manData['died']
```
Anna-Tkachenko, line 26: It's not amountOfExistence, it's sum of ages)

src/averageAges.js
```diff
@@ -37,7 +47,23 @@ function calculateMenAverageAge(people, century) {
...
+  let amountOfexistence = 0;
+
+  const women = withChildren
+    ? people.filter((person) => {
+      const peopleHaveMother = people.map(persone => persone['mother']);
+
+      return peopleHaveMother.includes(person['name']);
+    }) : people.filter(x => x['sex'] === 'f');
+
+  amountOfexistence = women.reduce((sum, woman) =>
```
Anna-Tkachenko, line 59: Same here with variable name

src/averageAges.js
```diff
@@ -55,7 +81,61 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      if (mother !== undefined) {
+        sum += child['born'] - mother['born'];
+        countOfpeople++;
+      }
+
+      return sum;
+    }, 0);
+
+    result = amountOfDifferAges / countOfpeople;
+  }
```
Anna-Tkachenko, line 111: Your this logic are the same for both of variants: if you have onlyWithSon argument and if not
Move this check at the start of the function as you did for previous functions
Just check at the start, if this argument is passed, you should find all kids that are boys
If this argument isn't passes - you should find all kids
And then logic is the same (reduce and calculating average age)

Anna-Tkachenko (changes requested):

src/averageAges.js
```diff
@@ -37,7 +47,23 @@ function calculateMenAverageAge(people, century) {
...
+
+  const women = withChildren
+    ? people.filter((person) => {
+      const peopleHaveMother = people.map(persone => persone['mother']);
+
+      return peopleHaveMother.includes(person['name']);
+    }) : people.filter(x => x['sex'] === 'f');
+
+  sumOfAges = women.reduce((sum, woman) =>
+    (woman['died'] - woman['born'] + sum), 0);
```
Denchepornyuk, line 60: you have two identical reduce, create a function for them

src/averageAges.js
```diff
@@ -55,7 +81,38 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const amountOfDifferAges = children.reduce((previousValue, child) => {
+    const mother = mothers.find((mom) => {
+      if (mom['name'] === child['mother']) {
+        return mom;
+      }
+    });
+
+    let sum = previousValue;
+
+    if (mother !== undefined) {
```
Denchepornyuk, line 105:
```suggestion
    if (mother) {
```

src/averageAges.js
```diff
@@ -55,7 +81,38 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const mothersArr = people.map(person => person['mother']);
+  const mothers = people.filter(person => mothersArr.includes(person['name']));
+  const men = people.filter(person => person['sex'] === 'm');
+  let countOfpeople = 0;
+  let result = 0;
+
+  const children = onlyWithSon
+    ? men.filter(y => y['mother']).filter(person => person !== null)
```
Denchepornyuk, line 91: you can get key from object by name, fix in all places
```suggestion
    ? men.filter(y => y.mother).filter(person => person !== null)
```

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #928, opened 2022-05-05
task completed

andreisucman:


src/averageAges.js
```diff
@@ -1,61 +1,51 @@
...
-  // without nesting
+  const men = people.filter(person => isMan(person));
+  const menAges = century
+    ? men.filter(person => Math.ceil(person.died / 100) === century)
+      .map(person => calculateAge(person)) : men.map(person => (
+      calculateAge(person)));
+
+  const averageAges = calculateAvgAge(menAges);
+
+  return averageAges;
```
meforoma, line 18: you can skip variable declaration when you're returning it next-line
```suggestion
  return calculateAvgAge(menAges);
```

meforoma (changes requested):
one tiny comment about simultaneous variable declaration and return
and wondering whether your tests went all right without rounding `averageAges`

andreisucman:
> one tiny comment about simultaneous variable declaration and return and wondering whether your tests went all right without rounding `averageAges`

Yes, all the tests pass. I also noticed that I didn't need rounding. 
image

PR #929, opened 2022-05-06
Solution

iLebo-stack:


elkhayate (approved):
Great job !

meforoma (approved):
great application of de-structuring 👍 
thought about introducing few helper functions that are re-used along the code,
but it's quite descriptive so can skip it.

Although it's always advised to abstract re-usable functions.

PR #942, opened 2022-05-09
Implemented solution

dellachaise:


src/averageAges.js
```diff
@@ -14,12 +14,21 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+
+  const menCentury
+    = men.filter(person => (Math.ceil(person.died / 100)) === century);
+
+  return (arguments.length < 2) ? getAverage(men) : getAverage(menCentury);
```
valerii-derkach, line 31: you can use here 
return century ? ...
instead arguments.length < 2;

valerii-derkach (approved):
Interesting solution with callback function!

one remark - for it is better to use name of argument instead of arguments.length in ternar operator ? :
it would be more understandable

PR #947, opened 2022-05-10
Solution

RomekIvantsiv:


src/averageAges.js
```diff
@@ -37,7 +44,20 @@ function calculateMenAverageAge(people, century) {
...
+      people.some(child => person.name === child.mother))
+    : people.filter((person) => person.sex === 'f');
+
+  const ages = womens.map(women => women.died - women.born);
+
+  const agesWomen = ages.reduce((total, age) => {
+    return total + age;
+  }, 0);
+
+  const avarageAgesWomen = agesWomen / ages.length;
```
iwanttosmile4u, line 58: You can return only divide operation without creating new variable.

src/averageAges.js
```diff
@@ -55,7 +75,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      people.some(child => child.name === person.mother));
+
+  const difference = children.map(child => {
+    const motherOfChild = people.find(mother => mother.name === child.mother);
+
+    return child.born - motherOfChild.born;
+  });
+
+  const ages = difference.reduce((total, age) => total + age);
+  const differenceAges = ages / difference.length;
```
iwanttosmile4u, line 91: The same remark - just return divide operation and don't create redundant variable. Or you can return ages - just add dividing on  'difference.length' after reduce: difference.reduce((total, age) => total + age) / difference.length;

src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
+  const mens = century
+    ? people.filter(men => Math.ceil(men.died / 100)
+  === century && (men.sex === 'm'))
+    : people.filter(person => person.sex === 'm');
+
+  const ages = mens.map(men => men.died - men.born);
+
+  const sumOfAges = ages.reduce((total, age) => {
+    return total + age;
+  }, 0);
```
Denchepornyuk, line 27: you use identical reduce three times, it is better to create a separate function with this reduce and use this function

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #950, opened 2022-05-10
Average age

dasha-chekmariova:


src/averageAges.js
```diff
@@ -55,7 +74,17 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    ? people.filter(person => person.sex === 'm'
+    && people.some(mother => mother.name === person.mother))
+    : people.filter(person => (
+      people.some(mother => mother.name === person.mother)));
+
+  const ageBetween = children.map(child => (
+    child.born - people.find(mother => mother.name === child.mother).born));
+
+  return ageBetween.reduce((prev, current) => (
+    prev + current), 0) / ageBetween.length;
```
Denchepornyuk, line 87: you use identical reduce three times, it is better to create a separate function with this reduce and use this function

PR #952, opened 2022-05-10
Solution

PRO-GRAM-MER:


elkhayate (approved):
Great job !

src/averageAges.js
```diff
@@ -38,6 +54,14 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const filteredWomen = withChildren
+    ? people.filter(person => people.find(child => isMother(child, person)))
+    : people.filter(person => person.sex === 'f');
```
meforoma, line 59: since you have a helper function `isMan` - you could use the opposite of it:
```suggestion
    : people.filter(!isMan(person));
```

meforoma (approved):

PR #953, opened 2022-05-10
 js_average-ages

anastasiia-nikita:


rostyslav-pyliak (approved):
Good

src/averageAges.js
```diff
@@ -55,7 +67,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    : people.filter(person => people.some(child =>
+      child.name === person.mother));
+
+  const difference = children.map(child => {
+    const motherChild = people.find(mother => mother.name === child.mother);
+
+    return child.born - motherChild.born;
+  });
+
+  const age = difference.reduce((sum, x) => (sum + x)) / difference.length;
```
Denchepornyuk, line 82: you use identical reduce three times, it is better to create a separate function with this reduce and use this function

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #959, opened 2022-05-11
solution

Sardorbek1122:


MabuzaM (approved):
Good job

meforoma (approved):
legit, just a common comment - when using method `.reduce()` - do specify the initial value
`.reduce((a, b) => a + b, 0)`

PR #960, opened 2022-05-11
finaly made it

uliton:


valerii-derkach (approved):
Good job!

Only one remark 18, 48, 78 line : method `.filter` returns new array, so you do not need to make copy before. You can write `people.filter` instead of  `[...people].filter` and your code will work

Denchepornyuk:
> Good job!
> 
> Only one remark 18, 48, 78 line : method `.filter()` returns new array, so you do not need to make copy before. You can write `people.filter()` instead of `[...people].filter()` and your code will work

Agree with @valerii-derkach

Denchepornyuk (approved):

PR #962, opened 2022-05-11
task is solved

enrikadej:


Andriyk0 (approved):

src/averageAges.js
```diff
@@ -54,8 +84,25 @@ function calculateWomenAverageAge(people, withChildren) {
  *
  * @return {number}
  */
-function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+function calculateAverageAgeDiff(people, onlyWithSon = 0) {
+  let allChildren = [...people].filter(
```
Denchepornyuk, line 88: method filter returns a new array
```suggestion
  let allChildren = people.filter(
```

src/averageAges.js
```diff
@@ -54,8 +84,25 @@ function calculateWomenAverageAge(people, withChildren) {
  *
  * @return {number}
  */
-function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+function calculateAverageAgeDiff(people, onlyWithSon = 0) {
```
Denchepornyuk, line 87: do you need 0 here?
you can use `onlyWithSon` without default value

src/averageAges.js
```diff
@@ -54,8 +84,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  let allChildren = [...people].filter(
+    child => people.some(person => child.mother === person.name)
+  ); // people who have mother === our children
+
+  if (onlyWithSon !== 0) {
```
Denchepornyuk, line 92:
```suggestion
  if (onlyWithSon) {
```

Denchepornyuk (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #997, opened 2022-06-10
Average ages

vsupruniuk:


jstmpelowycz (changes requested):

vsupruniuk:
I will try to fix the problem of time complexity when I finish the course of algorithms and data structures

src/averageAges.js
```diff
@@ -14,12 +14,21 @@
...
  */
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+  const menInCentry = century
```
Olexandr6, line 20: It's not `menInCentry` in cause `century` === `undefined`

src/averageAges.js
```diff
@@ -14,12 +14,21 @@
...
+    ? men.filter(
+      person => Math.ceil(person.died / 100) === century
+    )
+    : men;
```
Olexandr6, line 24: You filter 2 times when there is `century`. Let's implement this using only one filter

src/averageAges.js
```diff
@@ -37,7 +46,22 @@ function calculateMenAverageAge(people, century) {
...
+  const women = people.filter(person => person.sex === 'f');
+
+  const womenWithChild = withChildren
+    ? people.filter(
+      person => people.some(
+        child => child.mother === person.name
+      )
+    )
+    : women;
+
```
Olexandr6, line 58: The same case here, let's use filter only once

src/averageAges.js
```diff
@@ -55,7 +68,17 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const child = people
+    .filter(children => people.some(person => children.mother === person.name));
```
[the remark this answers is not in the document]
Olexandr6: HashMap is crying🥲

src/averageAges.js
```diff
@@ -55,7 +79,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const childIsSon = onlyWithSon
+    ? child.filter(
+      person => person.sex === 'm'
+    )
+    : child;
```
Olexandr6, line 89: Let's use filtering once🥲

src/averageAges.js
```diff
@@ -55,7 +79,23 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const child = people
+    .filter(children => people.some(person => children.mother === person.name));
+
+  const childIsSon = onlyWithSon
```
Olexandr6, line 85: It's not always sons here. Pay attention to variables naming
```suggestion
  const childIsSon = onlyWithSon
```

G-YaSha (changes requested):
Almost done

G-YaSha (approved):

PR #1004, opened 2022-06-12
solution

vinnibea:


src/averageAges.js
```diff
@@ -20,6 +20,20 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  const findMen = arguments.length < 2
+    ? people.filter(listOfMen =>
+      listOfMen.sex === 'm')
+
+    : people.filter(listOfMen =>
+      Math.ceil(listOfMen.died / 100) === century && listOfMen.sex === 'm');
```
illiarerun, line 29: you do the same thing in these 2 filter, by sorting out men. better filter out men first, and then, if needed filter them out by century
```suggestion
    ? people.filter(listOfMen =>
      listOfMen.sex === 'm')
    : people.filter(listOfMen =>
      Math.ceil(listOfMen.died / 100) === century && listOfMen.sex === 'm');
```

src/averageAges.js
```diff
@@ -20,6 +20,20 @@ function calculateMenAverageAge(people, century) {
...
+
+  const sumOfYears = findMen.reduce((previousValue, years) =>
+    previousValue + (years.died - years.born), 0);
```
illiarerun, line 32:
```suggestion
  const sumOfYears = findMen.reduce((previousValue, years) => (
    previousValue + (years.died - years.born)
  ), 0);
```

src/averageAges.js
```diff
@@ -38,6 +52,30 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const children = people.filter(child =>
+    child.mother !== null);
```
illiarerun, line 56: do you need `children` array before you know that `withChildren` was passed?

src/averageAges.js
```diff
@@ -38,6 +52,30 @@ function calculateMenAverageAge(people, century) {
...
+  let findWomen = people.filter(listOfWomen =>
+    listOfWomen.sex === 'f');
+
+  if (arguments.length > 1) {
+    const listWithChildren = [];
+
+    for (let i = 0; i < children.length; i++) {
```
illiarerun, line 63: // avoid using loop and forEach

illiarerun: you can definetely make it easier, all you need to find is if there are any children whose `mother` value is equal to any `woman.name`

src/averageAges.js
```diff
@@ -38,6 +52,30 @@ function calculateMenAverageAge(people, century) {
...
+      && typeof trueMothers !== 'undefined') {
+        listWithChildren.push(trueMothers);
+      }
+    }
+    findWomen = listWithChildren;
+  }
+
+  const sumOfYears = findWomen.reduce((previousValue, years) =>
+    previousValue + (years.died - years.born), 0);
+  const getAvarage = Math.round((sumOfYears / findWomen.length) * 100) / 100;
```
illiarerun, line 76: again, you do the same thing as in previous function, so make an external helper for that

src/averageAges.js
```diff
@@ -55,7 +93,29 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = arguments.length < 2
+    ? people.filter(child =>
+      child.mother !== null)
+
+    : people.filter(child =>
+      child.mother !== null && child.sex === 'm');
```
illiarerun, line 101: same here, filter out children first, and then filter them by sex, if needed

src/averageAges.js
```diff
@@ -55,7 +93,29 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const differencesOfYears = [];
+
+  for (let i = 0; i < children.length; i++) {
```
illiarerun, line 105: use array methods

illiarerun (changes requested):
GJ, now let's work on fixes:

PR #1013, opened 2022-06-15
Solution

AndriiNychaiuk:


illavodonis (approved):

src/averageAges.js
```diff
@@ -55,7 +71,21 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const allChildren = people
```
Olexandr6, line 74:
```suggestion
  const children = people
```

src/averageAges.js
```diff
@@ -55,7 +71,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    .filter(person => people.some(mama => mama.name === person.mother));
+
+  const maleChildren = allChildren.filter(person => person.sex === 'm');
```
Olexandr6, line 77:
```suggestion
  const sons = allChildren.filter(person => person.sex === 'm');
```

Olexandr6, line 77: You use here `filter` > `some` and then again `filter`. Let's use only `filter` > `some` without additional filtering

Olexandr6 (changes requested):
Well done👍 Your solution is good, but let's improve it a bit

Olexandr6 (approved):
🔥

PR #1016, opened 2022-06-15
Develop

illavodonis:


Tsymbal-Viacheslav (approved):

src/averageAges.js
```diff
@@ -55,7 +72,33 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(child => {
+    const hasMather = people.some(mather => child.mother === mather.name);
+    const sexIsMale = child.sex === 'm';
```
jstmpelowycz, line 77: if you see, you duplicate `sexIsMale` and `sexIsFemale`, so you could create a function so that you filter people by certain sex

src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menDiedSelectCentury = people.filter(human => {
+    const sexIsMale = human.sex === 'm';
+    const diedInSelectCentury = century
+      ? Math.ceil(human.died / 100) === century
+      : human.born;
+
+    return (sexIsMale && diedInSelectCentury);
+  });
```
jstmpelowycz, line 25: let's do something like this:

```
const men = getBySex(...);
const menDiedInSelectedCentury = men.filter(...);
```
by doing so, you make sure you split tasks logically so that it doesn't look messy

src/averageAges.js
```diff
@@ -37,7 +44,17 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenHasChildren = people.filter(human => {
+    const sexIsFemale = human.sex === 'f';
+    const hasChildren = withChildren
+      ? people.some(children => children.mother === human.name)
+      : human.born;
+
+    return (sexIsFemale && hasChildren);
+  });
```
jstmpelowycz, line 54: the same way here

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #1017, opened 2022-06-15
Develop

ostapfriend:


src/averageAges.js
```diff
@@ -14,12 +30,13 @@
...
  * @return {number}
  */
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const AverageManAge = century
```
Olexandr6, line 35: It's not avarageManAge, rather menAgesSum
```suggestion
  const menAgesSum = century
```

src/averageAges.js
```diff
@@ -14,12 +30,13 @@
...
+    ? people.filter(person => Math.ceil(person.died / 100) === century)
+    : people;
+
+  return getAverageAge(getSex(AverageManAge, 'm'));
```
Olexandr6, line 39: Ohh, you use `filter` two times here, let's immediately filter only man ages? In this case, we need to use `filter` only once

src/averageAges.js
```diff
@@ -36,8 +53,15 @@ function calculateMenAverageAge(people, century) {
...
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = getSex(people, 'f');
+
+  const womenWithChild = withChildren
+    ? women.filter(person => getAllMother(people, person))
+    : women;
```
Olexandr6, line 62: Let's improve performance here and use only filter + some, instead of 2 filters + find

src/averageAges.js
```diff
@@ -54,8 +78,20 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
+
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  let kind = people
+    .filter(child => people.some(mother => mother.name === child.mother));
+
+  if (onlyWithSon) {
+    kind = kind.filter(person => person.sex === 'm');
+  }
```
Olexandr6, line 88: Don't use if statements in this task, let's practice ternary operators

src/averageAges.js
```diff
@@ -54,8 +78,20 @@ function calculateWomenAverageAge(people, withChildren) {
  *
  * @return {number}
  */
+
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  let kind = people
```
Olexandr6, line 83:
```suggestion
  let children = people
```

src/averageAges.js
```diff
@@ -54,8 +78,20 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    .filter(child => people.some(mother => mother.name === child.mother));
+
+  if (onlyWithSon) {
+    kind = kind.filter(person => person.sex === 'm');
+  }
+
+  const result = kind
```
Olexandr6, line 90: Let's rename this var

src/averageAges.js
Olexandr6, line 88: Again, you use 2 `filters` + `some`, let's better use `filter` > `find`

Olexandr6 (changes requested):
Pay attention to variables' naming.

src/averageAges.js
ostapfriend: I don't understand, I use 1 filter

src/averageAges.js
```diff
@@ -54,8 +78,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  let children = people
+    .filter(child => people.find(mother => mother.name === child.mother));
+
+  let result = onlyWithSon
+    ? children = children.filter(person => person.sex === 'm') : 0;
+
+  result = children
+    .reduce((sum, child) => sum + child.born - findParent(people, child), 0);
```
jstmpelowycz, line 93:
```suggestion
  result = children.reduce(
  	(sum, child) => sum + child.born - findParent(people, child),
  	0,
 );
```

jstmpelowycz (commented):
GJ 🔥

jstmpelowycz (approved):

ostapfriend:
Thank you for approved)

PR #1036, opened 2022-06-18
solution was added

Bielichenko:


src/averageAges.js
```diff
@@ -20,6 +20,30 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  let num = 0;
+  const sum = people.reduce(calculateSum, 0);
+
+  function calculateSum(prev, person) {
+    if (century) {
+      if (person.sex === 'm' && Math.ceil(person.died / 100) === century) {
```
jstmpelowycz, line 28: you duplicate `person.sex === '...'` all over your code, so you can create a separate function allowing you to filter people by their sex, and then you just like
```
const men = ...;
const menDiedInCentury = men.filter(...);
```

src/averageAges.js
```diff
@@ -20,6 +20,30 @@ function calculateMenAverageAge(people, century) {
...
+      }
+    }
+
+    if (!century) {
+      if (person.sex === 'm') {
+        num++;
+
+        return prev + (person.died - person.born);
+      }
+    }
```
jstmpelowycz, line 41: using nested conditions is a bad approach, consider using ternary or something like that

jstmpelowycz (changes requested):

Olexandr6 (approved):

PR #1041, opened 2022-06-19
Add average ages

juajum:


Bielichenko (approved):
GJ

src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter((user) =>
+    century
+      ? Math.ceil(user.died / 100) === century && user.sex === 'm'
```
jstmpelowycz, line 20: you'd rather call `person`, not `user`

src/averageAges.js
```diff
@@ -55,7 +73,24 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      person => children.mother === person.name
+    )
+  );
+
+  const averageAgeDiff = child.reduce(
+    (sum, children) => sum + (
+      children.born - people.find(
+        mom => children.mother === mom.name).born),
+    0)
+    / child.length;
```
jstmpelowycz, line 91: agree it looks way much readable
```suggestion
  const ageDifferencesSum = child.reduce((sum, children) => {
    const mother = people.find(person => children.mother === person.name);
    const ageDifference = children.born - mother.born;
      
    return sum + ageDifference;
  });
  
  const averageAgeDifference = ageDifferencesSum / children.length;
```

jstmpelowycz (changes requested):

PR #1060, opened 2022-06-28
average-ages done

MaxShylo:


ArtemTopolskyi (approved):
strange variables names, but code style is cool

PR #1066, opened 2022-07-04
Solution

BudnikOleksii:


src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = century
+    ? people.filter(({ sex, died }) => sex === 'm'
+      && Math.ceil(died / 100) === century)
+    : people.filter(({ sex }) => sex === 'm');
```
solaryasha, line 21: You could just pass different callbacks to the `filter` method

src/averageAges.js
```diff
@@ -37,7 +39,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = withChildren
+    ? people.filter(
+      ({ name }) => people.some(({ mother }) => mother === name))
+    : people.filter(({ sex }) => sex === 'f');
```
solaryasha, line 45: Here as well

solaryasha (approved):
Good job !

src/averageAges.js
BudnikOleksii: Thank you, I didn't think about it)

PR #1076, opened 2022-07-07
Solution

frontend112:


src/averageAges.js
```diff
@@ -1,5 +1,9 @@
 'use strict';
 
+// callback for average age from first two tasks
+const calcAvg = (acc, curr) => {
+  return acc + curr.died - curr.born;
+};
```
meforoma, line 6: let's make it an actual `calcAvg`, so we provide it with an array of people
and get the average on `return`:
```
const calcAvg = (people) => {
  return people.reduce((acc, {died, born}) => acc + died - born, 0) / people.length;
};
```

src/averageAges.js
```diff
@@ -14,12 +18,20 @@
...
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const men = people.filter(person => person.sex === 'm');
+
+  const menBasedCentury = century
+    ? men.filter(person => century === Math.ceil(person.died / 100))
+    : men;
+
+  return menBasedCentury.reduce(calcAvg, 0) / menBasedCentury.length;
```
meforoma, line 34: with proper `calcAvg` you just do 
```suggestion
  return calcAvg(menBasedCentury);
```
```

Comment (meforoma) [reply src/averageAges.js:34]: and same for women

File: src/averageAges.js [hunk 3]
```diff
@@ -55,7 +77,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    return onlyWithSon ? (person.sex === 'm' && isChild) : isChild;
+  });
+
+  const diffAges = children.map(child => {
+    const mother = people.find(human => human.name === child.mother);
+
+    return child.born - mother.born;
+  });
+
+  return diffAges.reduce((a, b) => a + b, 0) / diffAges.length;
```
amitpatiljc, line 92: Here also you can use `calcAvg` function like you have already used

src/averageAges.js
```diff
@@ -14,12 +19,20 @@
...
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const men = people.filter(person => person.sex === 'm');
+
+  const menBasedCentury = century
+    ? men.filter(person => century === Math.ceil(person.died / 100))
+    : men;
+
+  return calcAvg(menBasedCentury);
```
amitpatiljc, line 35: I know its passing tests but try to round the resulting average ages at all three returns! 
prepare `roundAge `function and use
here for example - `return roundAge(calcAvg(menBasedCentury));`  will give you answer in round number

amitpatiljc (changes requested):
Good Job ! just few changes and good to go!

src/averageAges.js
frontend112: HI @amitpatiljc 
function calculateAverageAgeDiff returns arrays with only numbers
functions calculateMenAverageAge, calculateWomenAverageAge, returns object with filtered elements
The only solution that I see is simplify objects from calculateMenAverageAge and calculateWomenAverageAge (with map) and then calcAvg might be use for all cases

amitpatiljc (approved):
Good Job!

meforoma (approved):

PR #1077, opened 2022-07-08
Solution

Andrii-Kuzmenko:


marina-tilniak (changes requested):
Good job, but for the future use meaningful names for all variables ❗

alexandra-protyanova (approved):
Good job!

PR #1084, opened 2022-07-11
Oleksandr-Pustovit

Oleksandr-Pustovit:


src/averageAges.js
```diff
@@ -20,6 +20,18 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  const countedMen = people
+    .filter(person => person.sex === 'm')
```
dashasm, line 25: Думаю краще об'єднати два метода фільтр в один

dashasm (changes requested):

src/averageAges.js
Oleksandr-Pustovit: To be honest I agree pretty much with this point of view. 
BUT when I have to read all that enormous amounts of text in every ternary operator inside ONE filter my eyes bleed.
I think having separate filters each doing it's work one after another is a good aproach in this exact task.

dashasm (approved):

src/averageAges.js
```diff
@@ -20,6 +20,18 @@ function calculateMenAverageAge(people, century) {
...
+    .filter(man => century !== undefined
+      ? Math.ceil(man.died / 100) === century
+      : true
+    );
```
Ihor-Karpyn, line 29: decrease to 1 filter, fix in all cases

PR #1105, opened 2022-07-14
Solution

blcts:


yuliia-dronova (approved):
Everything looks correct :)

src/averageAges.js
```diff
@@ -15,6 +15,15 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  const mens = people
+    .filter(person => person.sex === 'm')
+    .filter(person => (century
+      ? Math.ceil(person.died / 100) === century
+      : person
+    ));
```
Ihor-Karpyn, line 23: Move condition inside filter, fix in all cases

Ihor-Karpyn (approved):

PR #1108, opened 2022-07-14
solution

butenko108:


yuliia-dronova (approved):
Everything looks correct)
Maybe it would be more conveniant to replace some parts into separate variables (for example, 'onlyWomen' or 'hasChild') for clear understanding what kind of people and why we use here. But it's optional)

ivan-kadykalo (approved):
🔥

PR #1143, opened 2022-08-05
all done

Yevhen-Pavlovskyi:


src/averageAges.js
```diff
@@ -38,6 +46,18 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const allMoms = people.map(human => human.mother);
+  const actualMoms = people.filter(human => allMoms.includes(human.name));
+
+  const filterWoman = (!withChildren)
```
alexandra-protyanova, line 52: it is better to remove `!` and replace options

alexandra-protyanova (approved):

PR #1149, opened 2022-08-10
Solution

aleksandra-prognimak:


src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filterMen = !century
+    ? people.filter(person => person.sex === 'm')
+    : people.filter(person => person.sex === 'm'
+      && Math.ceil(person.died / 100) === century);
+
+  return filterMen.reduce((sum, per) =>
+    sum + (per.died - per.born), 0) / filterMen.length;
 }
```
tiserett, line 25:
```suggestion
  const filterMen = !century
    ? people.filter(person => person.sex === 'm')
    : people.filter(person => (
      person.sex === 'm' && Math.ceil(person.died / 100) === century));

  return filterMen.reduce((sum, per) =>
    sum + (per.died - per.born), 0) / filterMen.length;
}
```

I think this will be more readable if you put all checks on the same line
You should also use round brackets if you move your shortened return statement in arrow function to a newline (that's from checklist)

src/averageAges.js
```diff
@@ -37,7 +39,13 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filterWomen = !withChildren
+    ? people.filter(person => person.sex === 'f')
+    : people.filter(person =>
+      people.find(child => child.mother === person.name));
+
+  return filterWomen.reduce((sum, per) =>
+    sum + (per.died - per.born), 0) / filterWomen.length;
```
tiserett, line 48:
```suggestion
  const filterWomen = !withChildren
    ? people.filter(person => person.sex === 'f')
    : people.filter(person => (
      people.find(child => child.mother === person.name)));

  return filterWomen.reduce((sum, per) => (
    sum + (per.died - per.born)), 0) / filterWomen.length;
```

examples how to put your brackets,  you can read it again in checklist

src/averageAges.js
```diff
@@ -55,7 +63,16 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const filterChildren = !onlyWithSon
+    ? people.filter(person =>
+      people.find(mom => person.mother === mom.name))
+    : people.filter(person =>
+      people.find(mom => person.mother === mom.name)
+      && person.sex === 'm');
+
+  return filterChildren.reduce((sum, per) =>
+    sum + per.born - people.find(person =>
+      per.mother === person.name).born, 0) / filterChildren.length;
```
tiserett, line 75:
```suggestion
  const filterChildren = !onlyWithSon
    ? people.filter(person => (
      people.find(mom => person.mother === mom.name)))
    : people.filter(person => (
      people.find(mom => person.mother === mom.name) && person.sex === 'm'));

  return filterChildren.reduce((sum, per) => (
    sum + per.born - people.find(person => (
      per.mother === person.name)).born), 0) / filterChildren.length;
```

tiserett (changes requested):
few chages required, according to the checklist

tiserett (approved):
well done :)

src/averageAges.js
```diff
@@ -15,11 +15,13 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filterMen = !century
```
AndriiRut, line 18: Maybe it's just me but try to think how another programmer will read your code and understand your variables.
For example `filterMen`, at first sight, I thought it should be a callback. Perhaps `filteredMen` or `onlyMen` would be better. 🤗

src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
+    ? people.filter(person => person.sex === 'm')
+    : people.filter(person => (
+      person.sex === 'm' && Math.ceil(person.died / 100) === century));
+
+  return filterMen.reduce((sum, per) => (
```
AndriiRut, line 23: reduce take a callback with next arguments:
(previousValue, currentValue, index, array) where 
`previousValue` - is the result of your previous action, like if you are adding all values in the array step by step it'll be written inside this parameter -> reduce(callback, `0`). Also if you didn't write start value -> reduce(callback) and passed only callback the first value of the array will be written inside the `previousValue` the next value will be written inside the `currentValue`.

`currentValue` - so this parameter of callback takes a current item of the array.

I thought you need this description because your second parameter of callback inside the reduce is named as `pre` - previousValue I think)

AndriiRut (approved):
I like your code and solutions 🤗 but I think you didn't figure out exactly how `reduce` works on this step)

By the way 🚀

PR #1150, opened 2022-08-11
Develop

tiserett:


src/averageAges.js
```diff
@@ -55,7 +61,16 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const kids = (onlyWithSon === undefined)
```
aleksandra-prognimak, line 64:
```suggestion
  const kids = !onlyWithSon
```

aleksandra-prognimak (changes requested):
Хорошая работа, но нужно немного улучшить код)

aleksandra-prognimak (approved):
🔥🔥🔥

src/averageAges.js
```diff
@@ -15,11 +15,12 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = !century
+    ? people.filter(person => person.sex === 'm')
+    : people.filter(person => (
+      person.sex === 'm' && Math.ceil(person.died / 100) === century));
+
+  return men.reduce((sum, man) => sum + (man.died - man.born), 0) / men.length;
```
dellachaise, line 23: Here and in line 46 you have very similar code, try to create and use separate function for this purpose.

dellachaise (approved):
Great job 👍 

But please add changes according to comments 👇

src/averageAges.js
tiserett: function is a great idea, done it

PR #1160, opened 2022-08-13
Solution

VolodymyrChuyko:


src/averageAges.js
```diff
@@ -55,7 +68,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    [...child, mothers[i] !== undefined ? mothers[i][3] : undefined]
+  ));
+  const totalInfoFiltered = totalInfo.filter(years => (
+    years[4] !== undefined && (!onlyWithSon || years[2] === 'm')
+  ));
+  const sumOfDiff = totalInfoFiltered.reduce((accumulator, years) => (
+    accumulator + (years[3] - years[4])
+  ), 0);
+
+  return sumOfDiff / totalInfoFiltered.length;
```
AndreyJigijalkin, line 87: I think you can shorten this part, using 'find' method:)

AndreyJigijalkin (approved):

src/averageAges.js
VolodymyrChuyko: Thank you for your advice. Now I've implemented it

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
   // learn how to use array methods like .filter .map .some .every .find .reduce
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const men = people.filter(person => (
+    person.sex === 'm' && (!century || century === Math.ceil(person.died / 100))
```
AlinaMatuschak, line 23: hard to read If you a have a lot of conditions/comparing, you can create variable for `(!century || century === Math.ceil(person.died / 100))`

src/averageAges.js
```diff
@@ -37,7 +43,14 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothers = people.map(person => person.mother);
+  const women = people.filter(person => (
+    person.sex === 'f' && (!withChildren || mothers.includes(person.name))
+  ));
```
AlinaMatuschak, line 49: nice solution )
also can try use some  method

src/averageAges.js
```diff
@@ -37,7 +43,14 @@ function calculateMenAverageAge(people, century) {
...
+  const womenAges = women.map(person => person.died - person.born);
+  const womenAgesSum = womenAges.reduce((sum, age) => sum + age, 0);
```
AlinaMatuschak, line 51: you can do just one loop instead of two. It will be better )

src/averageAges.js
```diff
@@ -55,7 +68,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      people.find(mother => mother.name === child.mother))
+    )
+    : people.filter(child => (
+      people.find(mother => mother.name === child.mother) && child.sex === 'm')
+    );
+
+  const averageAgeDiff = children.reduce((sumOfAges, child) => (
+    sumOfAges
+    + child.born
+    - people.find(mother => mother.name === child.mother).born
```
AlinaMatuschak, line 82: Will be easier to read if you create a separate variable for finded mother. But on your choice )

AlinaMatuschak (changes requested):
Awesome. But use one loop instead of two when you can )

src/averageAges.js
```diff
@@ -15,11 +15,22 @@
...
+  );
+  const men = people.filter(
+    person => person.sex === 'm' && diedInСertainСentury(person)
+  );
+  const menAgesSum = men.reduce(
+    (sum, person) => sum + person.died - person.born,
+    0
+  );
+
+  return menAgesSum / men.length;
```
AlinaMatuschak, line 33: you do the same on calculateWomenAverageAge. You can create separate function and you it here and on calculateWomenAverageAge

AlinaMatuschak (approved):
Good job 👏

PR #1177, opened 2022-08-17
Done

andrii-dmitruk:


Anastasiia-Khmelovska (approved):

src/averageAges.js
```diff
@@ -20,6 +20,19 @@ function calculateMenAverageAge(people, century) {
...
+      const diedCentury = Math.ceil(person.died / 100);
+
+      return person.sex === 'm' && diedCentury === century;
+    });
+
+  const eligibleMenCount = eligibleMen.length;
+
+  return eligibleMen
+    .reduce((sum, man) => sum + (man.died - man.born), 0)
+       / eligibleMenCount;
```
AlinaMatuschak, line 35: you do the same on calculateWomenAverageAge. You can create separate function and you it here and on calculateWomenAverageAge

src/averageAges.js
```diff
@@ -55,7 +78,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    .filter(person => people.find(child => person.name === child.mother));
+
+  const children = !onlyWithSon
+    ? people.filter(person => person.mother !== null)
+    : people.filter(person => person.mother !== null && person.sex === 'm');
+
+  let childrenCount = children.length;
+
+  return children.reduce((sum, child) => {
+    const currentMother = mothers.find(mother => child.mother === mother.name);
```
AlinaMatuschak, line 91: Can you use here people array instead of mothers? There will be -1 loop, you will delete the loop on 81 
```suggestion
    const currentMother = people.find(mother => child.mother === mother.name);
```

AlinaMatuschak (approved):
Awesome 🤩

PR #1179, opened 2022-08-17
Solution

Ohiiko:


plemiannik (approved):
good job 👍

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const arrMan = (!century)
```
dellachaise, line 18: `men` because there more then 1 man.
```suggestion
  const menArray = (!century)
```

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
+    ? people.filter((human) => human.sex === 'm')
+    : people.filter((human) => human.sex === 'm'
+      && Math.ceil(human.died / 100) === century);
```
dellachaise, line 21: [checklist point 2](https://github.com/mate-academy/js_average-ages/blob/master/checklist.md)

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
+
+  const averageAgeMan = arrMan.map((person) => person.died - person.born)
+    .reduce((a, b) => a + b) / arrMan.length;
```
dellachaise, line 24: What does it mean `a`, `b`? Please create meaningful variables.

src/averageAges.js
```diff
@@ -37,7 +41,15 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const arrWoman = (!withChildren)
+    ? people.filter((human) => human.sex === 'f')
+    : people.filter((mother) => mother.sex === 'f'
+      && (people.find((person) => mother.name === person.mother)));
```
dellachaise, line 47: [checklist point 2](https://github.com/mate-academy/js_average-ages/blob/master/checklist.md)


Method `some` would be better use for this purpose instead of `find`, because this way as I understand you need only boolean, not entire person object.

src/averageAges.js
```diff
@@ -37,7 +41,15 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const arrWoman = (!withChildren)
```
dellachaise, line 44: Because in array more than one woman.
``suggestion
  const womenArray = (!withChildren)
```

File: src/averageAges.js [hunk 6]
```diff
@@ -37,7 +41,15 @@ function calculateMenAverageAge(people, century) {
...
+    ? people.filter((human) => human.sex === 'f')
+    : people.filter((mother) => mother.sex === 'f'
+      && (people.find((person) => mother.name === person.mother)));
+
+  const averageAgeWonam = arrWoman.map((person) => person.died - person.born)
+    .reduce((a, b) => a + b) / arrWoman.length;
+
```
dellachaise, line 51: You do the same on `calculateMenAverageAge`. You can create a separate function and use it here and in `calculateMenAverageAge`.

As an option(up to you) you could create 2 separate variables: `women` and `womenWithChildren` and pass them as arguments to call your new `getAverage` function. This way ternary operator would be in return statement and code would be more clear

src/averageAges.js
```diff
@@ -55,7 +67,17 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const arrDiffAge = (!onlyWithSon)
+    ? people.filter(child =>
+      people.find(mother => child.mother === mother.name))
+    : people.filter(child =>
+      people.find(mother => child.mother === mother.name) && child.sex === 'm');
```
dellachaise, line 74: The same, method `some` would be better use for this purpose instead of `find`, because this way as I understand you need only boolean, not entire person object.

src/averageAges.js
```diff
@@ -55,7 +67,17 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const averageAgeDiff = arrDiffAge.reduce((count, child) =>
+    count + (child.born - people.find(mother =>
+      child.mother === mother.name).born), 0) / arrDiffAge.length;
```
dellachaise, line 78: Better create variable `motherBorn` for example, this way code would be easier to read.

dellachaise (changes requested):
Great solution 👍 

Just add changes according to comments 👇

src/averageAges.js
```diff
@@ -15,13 +15,20 @@
...
+    ? people.filter((human) => human.sex === 'm')
+    : people.filter((human) => (human.sex === 'm'
+      && Math.ceil(human.died / 100) === century));
+
+  const averageAgeMan = averageAge(menArray);
+
+  return averageAgeMan;
 }
 
+function averageAge(array) {
```
dellachaise, line 28: Rename according to [checklist point 1](https://github.com/mate-academy/js_average-ages/blob/master/checklist.md)

src/averageAges.js
```diff
@@ -55,7 +70,23 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const arrDiffAge = (!onlyWithSon)
```
dellachaise, line 73: I think better rename this way. Because this is array of children or only sons.
```suggestion
  const childrenWithMother = (!onlyWithSon)
```

dellachaise (approved):
Well done 👍 

Just add a few obligatory changes 👇

PR #1180, opened 2022-08-17
solution

Alzay007:


zinchenko-anastasiia (approved):
good job!) 
but i think it would be more understandable if you used different names( i mean for averageAges)

danylo-onishchuk (approved):

PR #1187, opened 2022-08-18
Solution

anna-kolmakova:


Roman-Artiukh (approved):
good job!))

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // without nesting
+  const filteredMens = !century
+    ? people.filter(person => person.sex === 'm')
+    : people.filter(person => person.sex === 'm'
+      && century === Math.ceil(person.died / 100));
+
+  const mensCount = filteredMens.length;
+
+  return filteredMens.reduce((sumAge, person) => sumAge
+    + (person.died - person.born), 0) / mensCount;
```
AlinaMatuschak, line 26: Are you sure that you checked checklist? 🙃

AlinaMatuschak, line 26: you do the same on calculateWomenAverageAge. You can create separate function and you it here and on calculateWomenAverageAge

src/averageAges.js
```diff
@@ -55,8 +67,19 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
-}
+  const filteredChild = !onlyWithSon
+    ? people.filter(child => (people.find(mother =>
+      mother.name === child.mother)))
```
AlinaMatuschak, line 72: you can use `some` method to check if the child has a mother (*optional)

src/averageAges.js
AlinaMatuschak: I see more than 3 issues from checklist on your code 🥲

AlinaMatuschak (changes requested):
Good logic, but should fix code style )

danylo-onishchuk (approved):

PR #1193, opened 2022-08-18
Solution

SergeyPistelev:


ShmelyovOleg (approved):
great, very neat code!

src/averageAges.js
```diff
@@ -37,7 +43,14 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filterWomen = !withChildren
+    ? people.filter(person => person.sex === 'f')
+    : people.filter(person =>
+      (people.find(child => child.mother === person.name)));
```
dellachaise, line 49: [checklist point 2](https://github.com/mate-academy/js_average-ages/blob/master/checklist.md)

Method `some` would be better use for this purpose instead of `find`, because this way as I understand you need only boolean, not entire person object.

src/averageAges.js
```diff
@@ -37,7 +43,14 @@ function calculateMenAverageAge(people, century) {
...
+  const sumAges = filterWomen
+    .reduce((sum, woman) => sum + (woman.died - woman.born), 0);
+
+  return sumAges / filterWomen.length;
```
dellachaise, line 53: You do the same on `calculateMenAverageAge`. You can create a separate function and use it here and in `calculateMenAverageAge`.

Also as an option(up to you) you could create 2 separate variables: `women` and `womenWithChildren` and pass them as arguments to call your new `getAverage` function. This way ternary operator would be in return statement and code would be more clear.

src/averageAges.js
```diff
@@ -55,7 +68,18 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const filterKid = people
```
dellachaise, line 71: Better rename to `kids` or `children` and in line 74 - `filteredKids`.

src/averageAges.js
```diff
@@ -55,7 +68,18 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    .filter(person => people.find(mother => person.mother === mother.name));
```
dellachaise, line 72: Method `some` would be better use for this purpose instead of `find`, because this way as I understand you need only boolean, not entire person object.

src/averageAges.js
```diff
@@ -55,7 +68,18 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const sortKid = !onlyWithSon ? filterKid : filterKid
+    .filter(kid => kid.sex === 'm');
+
+  const difference = sortKid
+    .reduce((sum, kid) =>
+      sum + (kid.born - people.find(mother =>
```
dellachaise, line 79: [checklist point 2](https://github.com/mate-academy/js_average-ages/blob/master/checklist.md)

Better create variable `motherBorn` for example, this way code would be easier to read.

src/averageAges.js
```diff
@@ -55,7 +68,18 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const filterKid = people
+    .filter(person => people.find(mother => person.mother === mother.name));
+
+  const sortKid = !onlyWithSon ? filterKid : filterKid
+    .filter(kid => kid.sex === 'm');
+
+  const difference = sortKid
```
dellachaise, line 77:
```suggestion
  const ageDifference = sortKid
```

dellachaise (changes requested):
Great solution 👍 

Just need to add changes according to comments 👇 

And please be more attentive with creating variable names. It should be meaningful. And answered the question `what stored this variable?`.

danylo-onishchuk (approved):

PR #1196, opened 2022-08-18
Solution

vadym-zahazei:


src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menList = people.filter(person =>
+    person.sex === 'm'
+    && (!century || Math.ceil(person.died / 100) === century)
+  );
```
plemiannik, line 21: ✈️

src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
+
+  const menSumAge = menList.reduce((
+    sum, man) => sum + (man.died - man.born
+  ), 0);
+
+  const menAverageAge = +(menSumAge / menList.length).toFixed(2);
```
plemiannik, line 27: not sure, cos it shoud work without toFixed, but at another side if we must return number in such format and if i remember correctly we must, than maybe yes

plemiannik (approved):
✈️

src/averageAges.js
vadym-zahazei: Yes, it works without `toFixed`, even without a `+`

danylo-onishchuk (approved):

PR #1212, opened 2022-08-20
solution

Reptiloid044:
Hi!

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -15,11 +15,18 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filterMenAge = century
+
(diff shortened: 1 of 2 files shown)


Maxim-Gumeniuk (approved):
GJ!

src/averageAges.js
```diff
@@ -55,7 +73,18 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const filterAgeDiff = onlyWithSon
+    // eslint-disable-next-line
```
danylo-onishchuk, line 77: зачем это правило?

danylo-onishchuk (changes requested):

src/averageAges.js
Reptiloid044: Hi!
I had a problem with max-length, so I decided to disable eslint, for more readable code.
Why can I not use this rule? 

image

AndriiRut (approved):
🔥 
You need to write your code according to lint. In your future work, you can get some other rules of lint and shouldn't comment on lint errors and break projects architecture=)

PR #1217, opened 2022-08-21
Initial solution

MihailKostov:


src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+  const lifeYears = (!century)
+    ? men.map(man => man.died - man.born)
+    : men.filter(man => Math.ceil(man.died / 100) === century)
+      .map(man => man.died - man.born);
+
+  return lifeYears.reduce((sum, age) => sum + age, 0) / lifeYears.length;
```
alexandra-protyanova, line 24: u can move this logic to separate function and use everywhere to prevent code duplicating

alexandra-protyanova (approved):

PR #1239, opened 2022-09-06
Solution

illia-kots:


src/averageAges.js
```diff
@@ -15,6 +15,17 @@
...
 function calculateMenAverageAge(people, century) {
+  const men = century
+    ? people.filter(person => person.sex === 'm'
+    && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === 'm');
+
+  const menAgeDied = men.map(year =>
+    year.died - year.born).reduce((sum, x) => sum + x);
+
+  const menAverageAge = menAgeDied / men.length;
```
alexandra-protyanova, line 26: let us move logic of rounding to separate function and call it in all places where needed

alexandra-protyanova (changes requested):

src/averageAges.js
illia-kots: Thank's! Such kind of solution is more compact))

Virass (approved):

PR #1244, opened 2022-09-12
solution

Ivan-Nevmerzhytskyi:


src/averageAges.js
```diff
@@ -55,7 +74,38 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ageDiff = people
+    ? group.map(child => {
+      const mother = people.find(person => person.name === child.mother);
+
+      return child.born - mother.born;
+    })
+    : group.map(person => person.died - person.born);
+
+  const sumOfAgeDiff = ageDiff.reduce((sum, x) => sum + x, 0);
```
stanislav-topikha, line 105: what is `x`?

src/averageAges.js
```diff
@@ -37,7 +41,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothersList = people.map(person => person.mother);
```
ivanchuk-yurii, line 44: you can move this variable inside `if` statement to not do unnecessary calculation when `withChildren` is `false`

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let men = people.filter(person => person.sex === 'm');
+
+  if (century) {
+    men = men.filter(person => Math.ceil(person.died / 100) === century);
+  }
+
+  const menAverageAge = calculateAverageAge(men);
```
ivanchuk-yurii, line 24: you can instantly return result of this function, variable is redundant, same in other places

src/averageAges.js
```diff
@@ -55,7 +68,37 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+      return child.born - mother.born;
+    })
+    : sampleOfPeople.map(person => person.died - person.born);
+
+  const sumOfAgeDiff = ageDifference.reduce((sum, ageDiff) => sum + ageDiff, 0);
+  const averageAge = sumOfAgeDiff / sampleOfPeople.length;
+
+  return averageAge;
 }
```
ivanchuk-yurii, line 102: this function should do only one action - calculate average age, but not define mother for child, because it becomes tricky. take only `ageDifference` as param here and do mapping 
``` 
sampleOfPeople.map(child => {
      const mother = people.find(person => person.name === child.mother);

    })
``` 
there `calculateAverageAgeDiff`

ivanchuk-yurii (changes requested):

ivanchuk-yurii (changes requested):
replace `if ()` statement with &&, || or ?: without nesting

SanyaBratashchuk (approved):

PR #1245, opened 2022-09-12
Solution

MaxManis:


src/averageAges.js
```diff
@@ -14,14 +14,20 @@
...
-  // without nesting
-}
+// - Here is some crazy solutions
+// |-- to solve it with no loops, if\else statements etc
+// |-- ...but a lot of variables))
+function calculateMenAverageAge(people, century = true) {
+  const men = people.filter(person => person.sex === 'm');
+  const ages = men
+    .map((person) =>
+      Math.ceil(person.died / 100) === century || century === true
```
Denchepornyuk, line 24: century is a number, I believe you don't need `=== true` for checking century
```suggestion
      Math.ceil(person.died / 100) === century || century
```

src/averageAges.js
```diff
@@ -14,14 +14,20 @@
...
+        ? person.died - person.born
+        : 0)
+    .filter(age => age !== 0);
```
Denchepornyuk, line 27: you can do all of these only with two filters
if we don't have century then filter only by sex, if we have century then filter by sex and Math.ceil

Something like this
```
ages = century
 ? filter by sex
 : filter be sex and math.ceil

return calc average
```

src/averageAges.js
```diff
@@ -36,8 +42,20 @@ function calculateMenAverageAge(people, century) {
...
+    .filter(person => person.mother !== null)
+    .map((person) => person.mother))];
+  const women = withChildren
+    ? female.filter(person => mothers.includes(person.name))
+    : female;
+  const ages = women
+    .map((person) => person.died - person.born)
+    .filter(age => age !== 0);
+
+  return ages.reduce((prev, curr) => prev + curr) / ages.length;
```
Denchepornyuk, line 58: you have the same reduce in the first function let's create a separate function for this logic

src/averageAges.js
```diff
@@ -36,8 +42,20 @@ function calculateMenAverageAge(people, century) {
...
+  const female = people.filter(person => person.sex === 'f');
+  const mothers = [...new Set(people
+    .filter(person => person.mother !== null)
+    .map((person) => person.mother))];
+  const women = withChildren
+    ? female.filter(person => mothers.includes(person.name))
+    : female;
+  const ages = women
+    .map((person) => person.died - person.born)
+    .filter(age => age !== 0);
```
Denchepornyuk, line 56: here you can also make a logic simpler, like in the calculateMenAverageAge function

Denchepornyuk (changes requested):

src/averageAges.js
MaxManis: @Denchepornyuk I fixed everything like logic, separate function, less methods and vars, BUT it doesnt work with out this `century === true` check))

Denchepornyuk (approved):

PR #1250, opened 2022-09-17
Solution

TheVovchik:


iPanda177 (approved):
I really like your reworked version!
GJ 🔥

src/averageAges.js
```diff
@@ -1,61 +1,57 @@
...
- * @return {number}
- */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const whatWeFiler = people.filter(person => (
+    arguments.length === 1
```
Anna-Tkachenko, line 5: it's not clear, why did you check arguments length here
let's check the century property, it will be more understandable

src/averageAges.js
```diff
@@ -1,61 +1,57 @@
...
+      ? person.sex === 'm'
+      : person.sex === 'm' && Math.ceil(person.died / 100) === century
+  ));
+  const ageDiffArray = [];
+
+  whatWeFiler.forEach(person => {
+    ageDiffArray.push(person.died - person.born);
+  });
```
Anna-Tkachenko, line 13: you repeated this logic 2 times for different arrays
let's create separate function that will accept people array and will return the average people age

src/averageAges.js
```diff
@@ -1,61 +1,57 @@
...
+  const whatWeFiler = people.filter(person => (
+    arguments.length === 1
+      ? person.sex === 'f'
+      : person.sex === 'f' && people.some(some => some.mother === person.name)
+  ));
+  const ageDiffArray = [];
+
+  whatWeFiler.forEach(person => {
+    ageDiffArray.push(person.died - person.born);
+  });
```
Anna-Tkachenko, line 28: let's create separate function for this logic

src/averageAges.js
```diff
@@ -1,61 +1,57 @@
...
- *
- * @param {object[]} people
- * @param {boolean} withChildren - optional
- *
- * @return {number}
- */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const whatWeFiler = people.filter(person => (
+    arguments.length === 1
```
Anna-Tkachenko, line 20: same here about arguments property

Anna-Tkachenko (changes requested):

src/averageAges.js
```diff
@@ -1,61 +1,61 @@
...
- *
- * @return {number}
- */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const whatWeFiler = people.filter(person => (
```
Anna-Tkachenko, line 4: it's not clear, what this variable mean
let's create more understandable variable name

Anna-Tkachenko (approved):

src/averageAges.js
TheVovchik: okay will replace it with listOfPeople in main functions and list in side once's

PR #1251, opened 2022-09-17
Done

iPanda177:


TheVovchik (changes requested):
you give me some tougths with your decision on last average calculation, I also filtered it and map each time. I like your decision in general here, it's a quastion of tastes what's is beter or which's kung fu is better.

I left some comment about naming. 
Just make them if you understand what I mean.
You can indicate me once more if you whant I reread it once more or move it in mentors review instead

src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const filterMan = !century
```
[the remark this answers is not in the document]
OleksiiBlezniuk: agree with @TheVovchik : filter - action, filtered - result of this action

src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
...
+    ? people.filter(human => human.sex === 'm')
+    : people.filter(human =>
+      Math.ceil(human.died / 100) === century && human.sex === 'm');
```
OleksiiBlezniuk, line 26: In both cases you used filter, so it can be moved out of ternary operator (just use it inside for condition to filter)

src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
...
+
+  const menAge = filterMan.map(human => human.died - human.born);
+
+  const averageAge = menAge.reduce((a, b) => a + b) / menAge.length;
+
+  return averageAge;
```
OleksiiBlezniuk, line 32: You've used this logic in every function, let's move it to separate helper function

src/averageAges.js
```diff
@@ -37,7 +47,21 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const mothersList = people.map(human => human.mother);
+
+  const mothersWithChildrens = people.filter(human =>
+    mothersList.includes(human.name)
+  );
+
+  const filterWoman = !withChildren
+    ? people.filter(human => human.sex === 'f')
+    : mothersWithChildrens;
```
OleksiiBlezniuk, line 58: You've iterated array 3-4 times here, actually you can reduce this value to 1-2

OleksiiBlezniuk: Just filter people array to find all woman (or all woman which names were used in `mother` properties)

src/averageAges.js
```diff
@@ -55,7 +79,29 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const kidsHasMothers = !onlyWithSon
+    ? people.filter(human => mothersNames.includes(human.mother))
+    : people.filter(human => mothersNames.includes(human.mother)
+      && human.sex === 'm');
+
+  const mothers = {};
+
+  mothersWithChildrens.map(function(human) {
+    mothers[human.name] = human.born;
+  });
```
OleksiiBlezniuk, line 99: You can make it simpler
Find all children using filter and then by map convert it to array of age differences

src/averageAges.js
```diff
@@ -55,7 +79,29 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ageDiff = kidsHasMothers.map(human =>
+    human.born - mothers[human.mother]);
```
OleksiiBlezniuk, line 102:
```suggestion
  const ageDiff = kidsHasMothers.map((human) => (
    human.born - mothers[human.mother]
  ));
```

OleksiiBlezniuk: Bad wrapping style, it was in checklist

OleksiiBlezniuk (changes requested):

src/averageAges.js
```diff
@@ -14,12 +14,24 @@
  *
  * @return {number}
  */
+function findAverageAge(filteredList) {
+  const age = filteredList.map(human => human.died - human.born);
+
+  const averageAge = age.reduce((a, b) => a + b) / age.length;
```
OleksiiBlezniuk, line 20: Actually, you can use only reduce here

src/averageAges.js
```diff
@@ -14,12 +14,24 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredMan = people.filter(human => {
+    if (!century) {
+      return human.sex === 'm';
+    }
```
OleksiiBlezniuk, line 29: // replace `if ()` statement with &&, || or ?:

src/averageAges.js
```diff
@@ -37,7 +49,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothersList = people.map(human => human.mother);
+
+  const filteredWoman = !withChildren
+    ? people.filter(human => human.sex === 'f')
+    : people.filter(human => mothersList.includes(human.name));
```
OleksiiBlezniuk, line 56: So in worst case this array will be iterated 3 times, but this amount can be reduced to 2
No need to map people array, just use `find` or `some` instead of `includes` to get access to object

Also use only one filter and choose needed condition using ternary operator inside it

src/averageAges.js
```diff
@@ -55,7 +73,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+        ? kid.sex === 'm' && kid.mother === mother.name
+        : kid.mother === mother.name
+    ));
+  });
+
+  const ageDiff = kids.map(kid => {
+    const mother = people.find(woman => woman.name === kid.mother);
+
+    return kid.born - mother.born;
+  });
```
OleksiiBlezniuk, line 88: You can also use reduce instead of map to calculate sum of ages

OleksiiBlezniuk (changes requested):

src/averageAges.js
```diff
@@ -14,12 +14,22 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredMan = people.filter(human => (
+    !century
+      ? human.sex === 'm'
+      : Math.ceil(human.died / 100) === century && human.sex === 'm'
+  ));
```
OleksiiBlezniuk, line 30:
```suggestion
  const filteredMan = people.filter(human => !century
      ? human.sex === 'm'
      : Math.ceil(human.died / 100) === century && human.sex === 'm'
  );
```

src/averageAges.js
```diff
@@ -37,7 +47,20 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  // const women = people.filter(human => human.sex === 'f');
+
+  // const filteredWoman = !withChildren
+  //   ? women
+  //   : women.filter(human => human.find(
+  //     kid => kid.mother === human.name));
```
OleksiiBlezniuk, line 55: Remove commented code

src/averageAges.js
```diff
@@ -55,7 +78,16 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  let kids = people.filter(kid => kid.mother
+    && people.find(human => human.name === kid.mother));
+
+  kids = !onlyWithSon
+    ? kids
+    : kids.filter(kid => kid.sex === 'm');
```
OleksiiBlezniuk, line 86: No need to use filter twice if `onlyWithSon` is passed, use it once and write needed condition inside

src/averageAges.js
```diff
@@ -55,7 +78,16 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  return kids.reduce((acc, cur) =>
+    acc + cur.born - people.find(human =>
+      human.name === cur.mother).born, 0) / kids.length;
```
OleksiiBlezniuk, line 90:
```suggestion
  return kids.reduce((acc, cur) => (
    acc + cur.born - people.find(
      human => human.name === cur.mother
    ).born, 0)
  ) / kids.length;
```

OleksiiBlezniuk: But better to save result of find to variable first

OleksiiBlezniuk, line 90:
```suggestion
  const agesSum = kids.reduce((acc, cur) => {
    const mother = people.find(human => human.name === cur.mother);
    
    return acc + cur.born - mother.born;
  }, 0);
  
  return agesSum / kids.length
```

OleksiiBlezniuk: Don't forget about code readability

OleksiiBlezniuk (changes requested):

src/averageAges.js
iPanda177: image
lint arguing when i do like you suggest.
"Unexpected use of comma operator".
i will do almost like you suggest but without extra hooks.

OleksiiBlezniuk (approved):
👍

PR #1254, opened 2022-09-19
solution

ihorandrianov:


src/averageAges.js
```diff
@@ -38,6 +49,16 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const woman = withChildren
+    ? people
+      .filter((person, i, array) => person.sex === 'f'
+        && array.some(a => a.mother === person.name))
+    : people.filter(person => person.sex === 'f');
+  const womanAvgAge = woman
+    .reduce((memo, person) => memo + (person.died - person.born), 0)
+      / woman.length;
```
Dikey945, line 59: You use this iteration a few times, so you can make a separate function with this action.

Dikey945:
> _ No description provided. _

Good Job! 🔥

src/averageAges.js
```diff
@@ -20,6 +20,17 @@ function calculateMenAverageAge(people, century) {
...
   // without nesting
+  const man = people.filter(person => person.sex === 'm');
+
+  const manAvgAge = century
+    ? man
+      .filter(person => Math.ceil(person.died / 100) === century)
+      .reduce((memo, person) => memo + (person.died - person.born), 0)
+      / man.filter(person => Math.ceil(person.died / 100) === century).length
+    : man.reduce((memo, person) => memo + (person.died - person.born), 0)
+      / man.length;
```
OleksiiBlezniuk, line 31: Let's simplify this logic

In case if century is passed in this function - you'll iterate array 4 times (`filter`, another `filter`, `reduce`, and another `filter`) This amount can be less 🙂  Just let's filter array only once (and use ternary operator inside to get needed filter condition if century is passed or not)

And then calculate sum of ages with `reduce`

OleksiiBlezniuk: Also you've used average value calculating in next function, so better move this logic to helper function and reuse it

src/averageAges.js
```diff
@@ -38,6 +49,16 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const woman = withChildren
+    ? people
+      .filter((person, i, array) => person.sex === 'f'
+        && array.some(a => a.mother === person.name))
+    : people.filter(person => person.sex === 'f');
```
OleksiiBlezniuk, line 56: Same here, instead of writing two filter methods for both cases, just write one, and select needed condition by ternary operator

src/averageAges.js
```diff
@@ -56,6 +77,25 @@ function calculateWomenAverageAge(people, withChildren) {
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  const peopleWithAgeDiff = people.map((person, i, array) => {
+    const momIndex = array.findIndex(a => a.name === person.mother);
+
+    if (momIndex !== -1) {
+      person.ageDiff = person.born - array[momIndex].born;
+    }
```
OleksiiBlezniuk, line 85: Instead of searching of index - you can find object using `find`, then you can use it instead of using index to find needed object

OleksiiBlezniuk: Also - it's a bad practice to mutate incoming objects

src/averageAges.js
```diff
@@ -56,6 +77,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+    return person;
+  }).filter(a => a.ageDiff);
+
+  const childWithAgeDiff = onlyWithSon
+    ? peopleWithAgeDiff.filter(a => a.sex === 'm')
+    : peopleWithAgeDiff;
```
OleksiiBlezniuk, line 92: Use filter instead of map to get needed objects, and then calculate age difference in `reduce` method

OleksiiBlezniuk (changes requested):

PR #1256, opened 2022-09-19
Solution

Shepeha-Andriy:


src/averageAges.js
```diff
@@ -15,11 +15,20 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let filteredMen = people.filter(x => x.sex === 'm');
```
Veremchukk, line 18: What is this x??

Veremchukk (commented):
You need to correct the variable names

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let filteredMen = people.filter(person => person.sex === 'm');
```
naziamoff, line 18: you can make it `const` variable and assign it a value depending on the `century` condition.

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
+  if (arguments.length > 1) {
+    const filterByCentury = filteredMen.filter(x => Math
+      .ceil(x.died / 100) === century);
+
+    filteredMen = [...filterByCentury];
+  }
+
+  let avg = filteredMen.reduce((sum, x) => sum + (x.died - x.born), 0);
+
+  avg = (avg / filteredMen.length);
```
naziamoff, line 29: try to isolate this logic into a separate function that calculates average age

naziamoff (changes requested):
- You have to pay attention to the naming. Your variable names should perfectly describe **what is stored inside** so using variable names like `res` `x`, `avg` and so on is strictly prohibited.
- fix my comments below

Anna-Tkachenko (approved):
Good job!

PR #1257, opened 2022-09-19
solution

artem-nasadchuk:


src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
+  const onlyMen = people.filter(men => men.sex === 'm');
+  const ages = onlyMen.map(year => year.died - year.born);
+
+  const agesCentury = onlyMen
+    .filter(men => Math.ceil(men.died / 100) === century)
+    .map(year => year.died - year.born);
+
+  return typeof century === 'number'
+    ? agesCentury.reduce((sum, age) => sum + age, 0) / agesCentury.length
+    : ages.reduce((sum, age) => sum + age, 0) / ages.length;
```
grafit-off, line 27: Looks good, but you can just not calculate `agesCentury` if there is no `century` passed in function. In your case you calculate `agesCentury` every time, even a `century` was not passed in function. Maybe it's not critical, just IMHO.

src/averageAges.js
```diff
@@ -37,7 +42,16 @@ function calculateMenAverageAge(people, century) {
...
+  const onlyWomen = people.filter(person => person.sex === 'f');
+  const ages = onlyWomen.map(person => person.died - person.born);
+
+  const mothers = onlyWomen
+    .filter(person => people.find(child => child.mother === person.name))
+    .map(person => person.died - person.born);
+
+  return withChildren === true
+    ? mothers.reduce((sum, age) => sum + age) / mothers.length
+    : ages.reduce((sum, age) => sum + age) / ages.length;
```
grafit-off, line 54: Here is such a problem as in `calculateMenAverageAge`.

src/averageAges.js
```diff
@@ -55,7 +69,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    ? children.filter(child => child.sex === 'm')
+    : children;
+
+  const ageDiff = children.map(child => {
+    const mother = people.find(person => person.name === child.mother);
+
+    return child.born - mother.born;
+  });
+
+  return ageDiff.reduce((x, y) => x + y, 0) / ageDiff.length;
```
grafit-off, line 86: What is x, y? Use more meaningful names for variables

grafit-off (approved):
Good job!

src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const onlyMen = people.filter(men => men.sex === 'm');
+  const ages = onlyMen.map(year => year.died - year.born);
+
+  const agesCentury = onlyMen
+    .filter(men => Math.ceil(men.died / 100) === century)
+    .map(year => year.died - year.born);
```
Anna-Tkachenko, line 23: you did this logic (creating an array of ages 2 times)
let's process in next way: at the start if the function we'll define the correct array of people (depending on the century value), if it's present -> we need to find men with defined century, if no -> just all men

and then do average age calculation for one men array

src/averageAges.js
```diff
@@ -37,7 +42,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const onlyWomen = people.filter(person => person.sex === 'f');
+  const ages = onlyWomen.map(person => person.died - person.born);
+
+  const mothers = onlyWomen
+    .filter(person => people.find(child => child.mother === person.name))
+    .map(person => person.died - person.born);
```
Anna-Tkachenko, line 50: same here, you wrote redundant functionality for getting ages 2 times
let's define one general array at the start of the function depending on the `withChildren` argument

Anna-Tkachenko (changes requested):

src/averageAges.js
artem-nasadchuk: agree, stupid solution, will fix it! thx:)

src/averageAges.js
artem-nasadchuk: I will edit it now, thx)

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const onlyMen = century
+    ? people
+      .filter(men => men.sex === 'm' && Math.ceil(men.died / 100) === century)
+    : people.filter(men => men.sex === 'm');
+
+  const ages = onlyMen.map(year => year.died - year.born);
+
+  return ages.reduce((sum, age) => sum + age, 0) / ages.length;
```
Anna-Tkachenko, line 25: you repeated this logic 2 times, let's create separate function that will accept people array and will return average age value

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1259, opened 2022-09-19
first commit

viktoria-krasnyuk:


src/averageAges.js
```diff
@@ -15,7 +15,13 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
+  const onlyMen = people.filter(person => person.sex === 'm');
+  const men = century
+    ? onlyMen.filter(person => Math.ceil(person.died / 100) === century)
+    : onlyMen;
+  const averageAge = men.reduce((sum, i) => sum + (i.died - i.born), 0);
+
+  return averageAge / men.length;
```
SanyaBratashchuk, line 24: move logic of calculating an average age to the separate function so you will be able to reuse it in your function instead of duplicate the same code

src/averageAges.js
```diff
@@ -55,7 +67,18 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const findChildren = onlyWithSon
```
SanyaBratashchuk, line 70: findChildren sounds like a function

src/averageAges.js
```diff
@@ -37,7 +43,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const findMother = withChildren
```
SanyaBratashchuk, line 46: findMother sounds like a function

SanyaBratashchuk (changes requested):

src/averageAges.js
```diff
@@ -37,7 +48,12 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const onlyMother = withChildren
+    ? people.filter(mother => mother.sex === 'f'
+    && people.some(person => mother.name === person.mother))
+    : people.filter(mother => mother.sex === 'f');
```
Bohdan-Maksymiuk, line 54: you have filter in both cases, so you can move `? :` logic into callback

src/averageAges.js
```diff
@@ -55,7 +71,18 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = onlyWithSon
+    ? people.filter(child =>
+      (people.find(mother => child.mother === mother.name)
+      && child.sex === 'm'))
+    : people.filter(child =>
+      people.find(mother => child.mother === mother.name));
```
Bohdan-Maksymiuk, line 79: same here

Bohdan-Maksymiuk (changes requested):

ArturErmolenko (approved):

PR #1261, opened 2022-09-19
Solution

Anna-Borodaienko:


jojnajbude (approved):

src/averageAges.js
```diff
@@ -20,6 +20,19 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  const men = (Number.isInteger(century))
+    ? people.filter(person =>
+      (person.sex === 'm'
+      && Math.ceil(person.died / 100) === century))
+    : people.filter(person =>
+      person.sex === 'm');
```
OleksiiBlezniuk, line 29: In both cases you used `filter`, move it outside ternary operator (just select needed condition inside filter method)

src/averageAges.js
```diff
@@ -20,6 +20,19 @@ function calculateMenAverageAge(people, century) {
...
+
+  const agedSum = men.reduce((prev, { born, died }) =>
+    prev + died - born,
+  0);
+
+  return agedSum / men.length;
```
OleksiiBlezniuk, line 35: You've used this logic in next function too, let's move it to helper function

OleksiiBlezniuk: object destructuring 🔥

src/averageAges.js
```diff
@@ -38,6 +51,20 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+
+  const women = (withChildren)
+    ? people.filter(person => getWomenWithChildren(people, person))
+    : people.filter(person => person.sex === 'f');
```
OleksiiBlezniuk, line 57: Same here with filter

src/averageAges.js
```diff
@@ -38,6 +51,20 @@ function calculateMenAverageAge(people, century) {
...
+
+  const ageSum = women.reduce((prev, { born, died }) =>
+    prev + died - born,
+  0);
+
+  return ageSum / women.length;
```
OleksiiBlezniuk, line 63: Move to helper

src/averageAges.js
```diff
@@ -56,6 +83,26 @@ function calculateWomenAverageAge(people, withChildren) {
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  const children = (onlyWithSon)
+    ? people.filter(child =>
+      getChildren(people, child)
+      && child.sex === 'm')
+    : people.filter(child =>
+      getChildren(people, child));
```
OleksiiBlezniuk, line 91: Same here with filter

src/averageAges.js
```diff
@@ -56,6 +83,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ages = children.map(child => getDiffAge(children, people, child));
+
+  return ages.reduce((prev, age) => prev + age) / children.length;
```
OleksiiBlezniuk, line 95: Actually you can just use one reduce instead of using both map and reduce

OleksiiBlezniuk (changes requested):

src/averageAges.js
```diff
@@ -20,6 +20,22 @@ function calculateMenAverageAge(people, century) {
...
+  const agedSum = getAverageAge(men);
+
+  return agedSum / men.length;
+}
+
+function getAverageAge(people) {
+  return people.reduce((prev, { born, died }) =>
+    prev + died - born,
+  0);
 }
```
OleksiiBlezniuk, line 39: But this function calculates sum of ages 🙃

OleksiiBlezniuk: `getAgesSum` will be better

on real projects usually we using a lot of functions that was created by other people, for that case we'll expect that our function will return average value ('caused it called getAverage), but it's result - is sum, we'll spent a lot of time to understand whats wrong in it

It's another point why variables naming is important

OleksiiBlezniuk (approved):

PR #1265, opened 2022-09-20
Average ages

Vladosik1337:


myroslav-diiak (approved):

src/averageAges.js
```diff
@@ -14,12 +14,22 @@
...
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  const men = people.filter(person => person.sex === 'm');
+  const menInCentury = men
+    .filter((person) =>
+      Math.ceil(person.died / 100) === century || century === true);
+
+  const deathYears = menInCentury.reduce((a, b) => (a + b.died), 0);
+  const bornYears = menInCentury.reduce((a, b) => (a + b.born), 0);
```
Anna-Tkachenko, line 30: you can do it with one iteration instead of 2
just calculate died - born inside the reduce callback and get the sum of this value

Anna-Tkachenko (changes requested):
Good job!
Just small suggestions for code optimizing

Anna-Tkachenko (approved):
Good job! 🔥

PR #1268, opened 2022-09-20
solution

artem-musii:


Yevheniya-Deryka (approved):
GJ!

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
-  // without nesting
+  let men = people.filter((person) => person.sex === 'm');
+
+  men = century
+    ? men.filter((person) => Math.ceil(person.died / 100) === century)
+    : men;
+
+  const sumOfAge = men.reduce((prev, man) => (
+    man.died - man.born + prev
+  ), 0);
```
Anna-Tkachenko, line 26: You did it 2 times (the same logic) for different arrays
Let's create separate function for this functionality that will accept an array and will return the average people age

src/averageAges.js
```diff
@@ -37,7 +43,18 @@ function calculateMenAverageAge(people, century) {
...
+  let women = people.filter((person) => person.sex === 'f');
+
+  women = withChildren
+    ? women.filter((mother) => people.some((person) => (
+      mother.name === person.mother
+    )))
+    : women;
+
+  const sumOfAge = women.reduce((prev, woman) =>
+    woman.died - woman.born + prev, 0);
```
Anna-Tkachenko, line 55: don't repeat yourself: create separate function for calculating average age for people array

Anna-Tkachenko (changes requested):
Good job!
Just one small suggestion

Anna-Tkachenko (approved):
Good job!

PR #1270, opened 2022-09-20
solution

vladimir-dublya:


src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
...
+  const onlyMen = people.filter(men => men.sex === 'm');
+  const ages = onlyMen.map(year => year.died - year.born);
+
+  const agesCentury = onlyMen
+    .filter(men => Math.ceil(men.died / 100) === century)
+    .map(year => year.died - year.born);
+
+  return typeof century === 'number'
+    ? agesCentury.reduce((sum, age) => sum + age, 0) / agesCentury.length
+    : ages.reduce((sum, age) => sum + age, 0) / ages.length;
```
imaksweb, line 32:
```suggestion
  return century
    ? Math.round(agesCentury.reduce((sum, age) => sum + age, 0) / agesCentury.length * 100) / 100
    : Math.round(ages.reduce((sum, age) => sum + age, 0) / ages.length * 100) / 100;
```

src/averageAges.js
```diff
@@ -38,6 +48,16 @@ function calculateMenAverageAge(people, century) {
...
+  const onlyWomen = people.filter(person => person.sex === 'f');
+  const ages = onlyWomen.map(person => person.died - person.born);
+
+  const mothers = onlyWomen
+    .filter(person => people.find(child => child.mother === person.name))
+    .map(person => person.died - person.born);
+
+  return withChildren === true
+    ? mothers.reduce((sum, age) => sum + age) / mothers.length
+    : ages.reduce((sum, age) => sum + age) / ages.length;
```
imaksweb, line 60:
```suggestion
  return withChildren
    ? mothers.reduce((sum, age) => sum + age) / mothers.length
    : ages.reduce((sum, age) => sum + age) / ages.length;
```

imaksweb (approved):
Good job! Just few suggestions 🙂

src/averageAges.js
OleksiiBlezniuk, line 32: You've used `filter`, `map`, `filter`, `map` and `reduce` here, it means that you've iterated array 5 times🥲

It can be simplified just to using `filter` and `reduce` in this function 🙂 

Just use ternary operator inside filter method and select needed condition to filter people, and then use reduce to calculate average age

OleksiiBlezniuk: Also you've used logic to find average ages in next function, so it can be moved to separate helper function

src/averageAges.js
OleksiiBlezniuk, line 60: Same as in previous case, refactor it please

src/averageAges.js
```diff
@@ -56,6 +76,21 @@ function calculateWomenAverageAge(people, withChildren) {
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  let children = people.filter(child => {
+    return people.some(mother => child.mother === mother.name);
+  });
+
+  children = onlyWithSon
+    ? children.filter(child => child.sex === 'm')
+    : children;
```
OleksiiBlezniuk, line 85: Combine this to one filter method and use ternary operator inside it

src/averageAges.js
```diff
@@ -56,6 +76,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ageDiff = children.map(child => {
+    const mother = people.find(person => person.name === child.mother);
+
+    return child.born - mother.born;
+  });
+
+  return ageDiff.reduce((x, y) => x + y, 0) / ageDiff.length;
```
OleksiiBlezniuk, line 93: You can use reduce, and calculate difference inside it, then you'll make amount of iterations less

OleksiiBlezniuk (changes requested):

src/averageAges.js
```diff
@@ -16,46 +16,73 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredMen = century
+    ? people
+      .filter(({ sex, died }) => {
+        const isPersonCentury = Math.ceil(died / 100) === century;
+        const isMale = sex === 'm';
+
+        return isMale && isPersonCentury;
```
Andr1yK, line 25: 🔥

src/averageAges.js
```diff
@@ -16,46 +16,73 @@
...
- * Hint: To check if a woman has children you should find someone who mention
- * her as mother.
- *
- * @param {object[]} people
- * @param {boolean} withChildren - optional
- *
- * @return {number}
- */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
```
Andr1yK, line 34: remove needless comments

src/averageAges.js
```diff
@@ -16,46 +16,73 @@
...
+      const hasMotherData = people
+        .some(mother => mother.name === person.mother);
+      const isSon = person.sex === 'm';
+
+      return hasMother && hasMotherData && isSon;
+    })
+    : people.filter(person => {
+      const hasMother = person.mother !== null;
+      const hasMotherData = people
+        .some(mother => mother.name === person.mother);
```
Andr1yK, line 61: If we don't find data about the mother, that mother is not found too
```suggestion
      const hasMotherData = people
        .some(mother => mother.name === person.mother);
```

Andr1yK (approved):
Cool code style 🔥

PR #1273, opened 2022-09-20
solution

myroslav-diiak:


Vladosik1337 (approved):
Gj)

src/averageAges.js
```diff
@@ -56,6 +73,30 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  return ageDifference / filteredChildren.length;
+}
+
+function filterBySex(people, sex) {
+  return people.filter(person => person.sex === sex);
+}
+
+function averageAge(filteredArray) {
+  const sumOfAge = filteredArray.map(person => person.died - person.born)
+    .reduce((prev, curr) => prev + curr);
```
OleksiiBlezniuk, line 97: You can use only reduce here, just take object instead of number here, and add to accumulator difference between born and died properties

OleksiiBlezniuk (changes requested):
Avoid to use filter method twice if you able to do it only once

src/averageAges.js
```diff
@@ -56,6 +61,50 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  }
+
+  if (withChildren) {
+    filtered = filtered.filter(person => people.find(
+      child => child.mother === person.name
+    ));
+  }
+
+  return filtered;
+}
```
OleksiiBlezniuk, line 98: Again, here you've used filter method twice, but you can do it using it only once, just write needed condition inside it

src/averageAges.js
```diff
@@ -56,6 +61,50 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  const filteredChildren = (!onlyWithSon)
+    ? people.filter(person => people
+      .find(mother => mother.name === person.mother))
+    : peopleFilter(
+      people.filter(person => people
+        .find(mother => mother.name === person.mother)),
+      'm'
+    );
```
OleksiiBlezniuk, line 71: Better to write one filter method and write needed condition inside it instead of writing filter method for every case

OleksiiBlezniuk (changes requested):

src/averageAges.js
```diff
@@ -20,6 +20,12 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const filteredMales = (!century)
+    ? peopleFilter(people, person => person.sex === 'm')
+    : peopleFilter(people, person => person.sex === 'm'
+      && Math.ceil(person.died / 100) === century);
```
OleksiiBlezniuk, line 26: As for me moving filter method to separated function is not so good idea, but ok

OleksiiBlezniuk, line 26: Follow DRY

OleksiiBlezniuk, line 26:
```suggestion
  const filteredMales = peopleFilter(people, !century
    ? person => person.sex === 'm')
    : person => person.sex === 'm'
      && Math.ceil(person.died / 100) === century);
```

OleksiiBlezniuk: Or even

OleksiiBlezniuk, line 26:
```suggestion
  const filteredMales = people.filter(person => !century
    ? person.sex === 'm'
    : person.sex === 'm'
      && Math.ceil(person.died / 100) === century
  );
```

src/averageAges.js
```diff
@@ -55,7 +66,35 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const filteredChildren = (!onlyWithSon)
+    ? peopleFilter(people, person => people.find(
+      mother => mother.name === person.mother
+    ))
+    : peopleFilter(people, person => people.find(
+      mother => mother.name === person.mother && person.sex === 'm'
+    ));
```
OleksiiBlezniuk, line 75: DRY, select needed callback inside `peopleFilter` instead of writing function twice

OleksiiBlezniuk (changes requested):

src/averageAges.js
```diff
@@ -20,6 +20,13 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const filteredMales = peopleFilter(people, !century
+    ? person => person.sex === 'm'
```
OleksiiBlezniuk, line 24:
```suggestion
    ? (person) => person.sex === 'm'
```

src/averageAges.js
```diff
@@ -55,7 +68,36 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const filteredChildren = peopleFilter(people, !onlyWithSon
+    ? person => people.find(
+      mother => mother.name === person.mother
+    )
+    : person => people.find(
+      mother => mother.name === person.mother && person.sex === 'm'
+    )
+  );
```
OleksiiBlezniuk, line 78:
```suggestion
  const findMother = (motherName) => people
    .find((person) => name === motherName);
    
  const filterCallback = !onlyWithSon
    ? (person) => findMother(person.name)
    : (person) => findMother(person.name) && person.sex === 'm';

  const filteredChildren = peopleFilter(people, filterCallback);
```

OleksiiBlezniuk (approved):

PR #1280, opened 2022-09-20
result

AndriyZhabynets:


igor-1987 (approved):
Nice job and interesting solutions!

src/averageAges.js
```diff
@@ -36,8 +49,16 @@ function calculateMenAverageAge(people, century) {
...
  */
-function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+function calculateWomenAverageAge(people, withChildren = false) {
+  const allWomens = people.filter(human => human.sex === 'f');
+
+  const AgesWomens = withChildren === true
+    ? allWomens.filter(human => people
+      .find(child => child.mother === human.name))
+    : allWomens;
```
naziamoff, line 58: - You don't need to check whether some boolean var === true. You may use shortcut here
- think of better name for this variable. It should describe what is stored inside and should not be capitalizied.
```suggestion
  const AgesWomens = withChildren
    ? allWomens.filter(human => people
      .find(child => child.mother === human.name))
    : allWomens;
```

naziamoff (changes requested):

src/averageAges.js
```diff
@@ -7,19 +7,32 @@
...
+  const averageAgeTotal = sumAllAge / allAge.flat().length;
+
+  return averageAgeTotal;
+}
+
+function calculateMenAverageAge(people, century = true) {
+  const allMens = people.filter(human => human.sex === 'm');
+
+  const centuryAgesMens = allMens.filter(human =>
+    Math.ceil(human.died / 100) === century || century === true);
```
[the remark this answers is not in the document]
AndriyZhabynets: without it, the first test will not pass

src/averageAges.js
```diff
@@ -7,19 +7,31 @@
...
  * @return {number}
  */
-function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+
+function averageAge(people) {
```
naziamoff, line 18: function name should describe the action it does, so it should start with verb.
Examples: `get...`, `set...`, `calculate...` and so on

src/averageAges.js
```diff
@@ -7,19 +7,31 @@
...
+
+  const ages = people.map(person => person.died - person.born)
+    .reduce((sum, current) => sum + current, 0);
```
naziamoff, line 21: try to do it without map, just with reduce

src/averageAges.js
```diff
@@ -7,19 +7,31 @@
...
+  const averageAgeTotal = ages / people.length;
+
+  return averageAgeTotal;
+}
+
+function calculateMenAverageAge(people, century = true) {
+  const allMens = people.filter(human => human.sex === 'm');
+
+  const centuryAgesMens = allMens.filter(human =>
+    Math.ceil(human.died / 100) === century || century === true);
```
naziamoff, line 32: try to get proper men array with just one filter

src/averageAges.js
```diff
@@ -36,8 +48,15 @@ function calculateMenAverageAge(people, century) {
...
  */
-function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+function calculateWomenAverageAge(people, withChildren = false) {
+  const allWomens = people.filter(human => human.sex === 'f');
+
+  const ageWomens = withChildren
+    ? allWomens.filter(human => people
+      .find(child => child.mother === human.name))
+    : allWomens;
```
naziamoff, line 57: same here, do it with just one filter

naziamoff (changes requested):

Denchepornyuk (approved):

PR #1282, opened 2022-09-21
Solution

Oleksandra-Shevchenko:


src/averageAges.js
```diff
@@ -14,12 +14,31 @@
...
+
+  const srtMen = people.filter(human => human.sex === 'm');
+  const artMenCentury = srtMen.filter(person =>
+    Math.ceil(person.died / 100) === century);
+
+  if (century) {
+    return calculateAveragePeopleAge(artMenCentury);
+  } else {
+    return calculateAveragePeopleAge(srtMen);
+  }
```
Rina07, line 41: ``
```suggestion
  if (century) {
    return calculateAveragePeopleAge(artMenCentury);
  } 
    
  return calculateAveragePeopleAge(srtMen);
  
```

Rina07: Maybe omit extra else block? :)

src/averageAges.js
```diff
@@ -36,8 +55,22 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const srtWomen = people.filter(human => human.sex === 'f');
+  const womenWithChild = srtWomen.filter(woman =>
+    findLostChild(people, woman));
+
+  if (withChildren) {
+    return calculateAveragePeopleAge(womenWithChild);
+  } else {
+    return calculateAveragePeopleAge(srtWomen);
+  }
```
Rina07, line 73:
```suggestion
  if (withChildren) {
    return calculateAveragePeopleAge(womenWithChild);
  } 
    
  return calculateAveragePeopleAge(srtWomen);  
```

Rina07: and here? :)

Rina07 (approved):
Looks good :star: A few small suggestions if you are agree)

src/averageAges.js
```diff
@@ -14,12 +14,32 @@
  *
  * @return {number}
  */
+
+function calculateAveragePeopleAge(people) {
+  const ages = people.map(person => person.died - person.born);
+
+  const averageAge = ages.reduce((x, y) => x + y, 0) / ages.length;
+
+  return averageAge;
```
silchyn, line 23: There is no need to create a variable to return it on the following line. You can create a variable for sum, for example, to improve readability and return this sum divided by the length

src/averageAges.js
```diff
@@ -36,8 +56,23 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+function findLostChild(people, mother) {
```
silchyn, line 59: findLostChild suggests that the function returns the found child, but your function returns simply whether or not the mother has child, so it is better to name it smth like hasChild

silchyn (changes requested):

src/averageAges.js
```diff
@@ -14,12 +14,31 @@
...
   // write code here
   // learn how to use array methods like .filter .map .some .every .find .reduce
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  const srtMen = people.filter(human => human.sex === 'm');
+  const artMenCentury = srtMen.filter(person =>
+    Math.ceil(person.died / 100) === century);
+
```
Anna-Tkachenko, line 36: what is srtMen?

for now you create 2 arrays for each function call
let's create only one array (call filter method only one time with correct check depending on the century argument)

src/averageAges.js
```diff
@@ -36,8 +55,22 @@ function calculateMenAverageAge(people, century) {
...
+  return children.length !== 0;
+}
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const srtWomen = people.filter(human => human.sex === 'f');
+  const womenWithChild = srtWomen.filter(woman =>
+    hasChild(people, woman));
+
```
Anna-Tkachenko, line 68: same here, don't create 2 arrays each time, one of them is redundant

Anna-Tkachenko (changes requested):

src/averageAges.js
```diff
@@ -14,12 +14,33 @@
...
+}
+
 function calculateMenAverageAge(people, century) {
   // write code here
   // learn how to use array methods like .filter .map .some .every .find .reduce
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  const srtMen = people.filter(human => human.sex === 'm');
```
naziamoff, line 33: think of better name for this variable, it is totally unclear now

src/averageAges.js
```diff
@@ -14,12 +14,33 @@
...
+
+  const menYears = century
+    ? srtMen.filter(person => Math.ceil(person.died / 100) === century)
+    : srtMen;
+
+  if (century) {
+    return calculateAveragePeopleAge(menYears);
+  }
+
+  return calculateAveragePeopleAge(menYears);
```
naziamoff, line 43: looks like those 2 are the same, do you really need both?

src/averageAges.js
```diff
@@ -36,8 +57,24 @@ function calculateMenAverageAge(people, century) {
...
+
+  const womenYear = withChildren
+    ? srtWomen.filter(woman => hasChild(people, woman))
+    : srtWomen;
+
+  if (withChildren) {
+    return calculateAveragePeopleAge(womenYear);
+  }
+
+  return calculateAveragePeopleAge(womenYear);
```
naziamoff, line 77: again 2 totally same return statements. Try to remove one

src/averageAges.js
```diff
@@ -54,8 +91,31 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    const children = findChildren(people, woman, onlyWithSon);
+
+    children.forEach(child => {
+      ages[ages.length] = child.born - woman.born;
+    });
+  });
+
+  return ages.reduce((x, y) => {
+    return x + y;
+  }, 0) / ages.length;
```
naziamoff, line 109: no need to use curly brackets here, take a look at the checklist

src/averageAges.js
```diff
@@ -54,8 +91,31 @@ function calculateWomenAverageAge(people, withChildren) {
...
+}
+
+function findChildren(people, mother, onlyWithSon) {
+  return people.filter(child => {
+    const isChild = child.mother === mother.name;
+    const maleChild = child.sex === 'm';
+
+    return onlyWithSon ? isChild && maleChild : isChild;
```
naziamoff, line 117: You may simplify this a little
```suggestion
    return onlyWithSon
      ? child.mother === mother.name && child.sex === m
      : child.mother === mother.name
```

naziamoff (changes requested):

src/averageAges.js
```diff
@@ -36,8 +53,20 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+function hasChild(people, mother) {
+  const children = people.filter(child => child.mother === mother.name);
+
+  return children.length !== 0;
+}
```
naziamoff, line 60: looks like you implement method some here. Why don't we just use it?

src/averageAges.js
```diff
@@ -36,8 +53,20 @@ function calculateMenAverageAge(people, century) {
...
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const listOfWomen = people.filter(human => human.sex === 'f');
+
+  const womenYear = withChildren
```
naziamoff, line 65: same here, do we really store women years here?

naziamoff (changes requested):

src/averageAges.js
Oleksandra-Shevchenko: we can't use the method some because mothers can have more than 1 child in tests, so I use filter

Oleksandra-Shevchenko: but actually you are right here :)

src/averageAges.js
```diff
@@ -54,8 +82,30 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    x + y
+  ), 0) / ages.length;
+}
+
+function findChildren(people, mother, onlyWithSon) {
+  return people.filter(child => {
+    return onlyWithSon
+      ? child.mother === mother.name && child.sex === 'm'
+      : child.mother === mother.name;
+  });
```
naziamoff, line 108:
```suggestion
  return people.filter(child => (
    onlyWithSon
      ? child.mother === mother.name && child.sex === 'm'
      : child.mother === mother.name;
  ));
```

src/averageAges.js
```diff
@@ -36,8 +53,19 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+function hasChild(people, mother) {
+  return people.some(child => child.mother === mother.name);
+
```
naziamoff, line 58:
```suggestion
```

naziamoff (approved):

PR #1283, opened 2022-09-21
Solution

jojnajbude:


Anna-Borodaienko (approved):

src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menAges = people.filter(person => {
+    return person.sex === 'm'
+  && (century ? Math.ceil(person.died / 100) === century : true);
+  }).reduce((prev, { born, died }) => {
+    prev.push(died - born);
+
+    return prev;
+  }, []);
```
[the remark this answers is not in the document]
OleksiiBlezniuk: Why to use reduce to modify all elements of array? Use map for this purpose
But better just calculate sum of ages (no need to use reduce twice, when you can use it once)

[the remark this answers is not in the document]
OleksiiBlezniuk: Also wrap elements of ternary operator to new line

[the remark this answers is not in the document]
OleksiiBlezniuk: If you want to move repeated `person.sex === 'm'` out of ternary operator, you can save it to variable, separate you big code block to smaller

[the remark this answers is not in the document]
OleksiiBlezniuk: Also you'll calculate average age several times, move it to separate helper function

src/averageAges.js
```diff
@@ -37,7 +42,16 @@ function calculateMenAverageAge(people, century) {
...
+  const womenAges = people.filter((person, idx, arr) => {
+    return person.sex === 'f'
+    && (withChildren ? arr.some(child => child.mother === person.name) : true);
+  }).reduce((prev, { born, died }) => {
+    prev.push(died - born);
+
+    return prev;
+  }, []);
+
+  return womenAges.reduce((a, b) => a + b, 0) / womenAges.length;
```
OleksiiBlezniuk, line 54: Same here

src/averageAges.js
```diff
@@ -55,7 +69,24 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const hasMother = (group, person) => {
+    const mother = group.find(mthr => mthr.name === person.mother);
+
+    return mother;
+  };
```
OleksiiBlezniuk, line 76:
```suggestion
  const hasMother = (group, person) => {
    return group.find(mother => mother.name === person.mother);
  };
```

OleksiiBlezniuk: Also better to move helper function outside parent function

src/averageAges.js
```diff
@@ -55,7 +69,24 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ageDiff = [...people].filter((person, idx, arr) => {
```
OleksiiBlezniuk, line 78: filter is non-mutating method

OleksiiBlezniuk: where have you used index? why do you need this array from filter if you have `people`?

src/averageAges.js
```diff
@@ -15,11 +15,26 @@
...
+  };
+
+  const suitableMen = people
+    .filter(person => {
+      const isMan = person.sex === 'm';
+
+      return isMan
+        && (
+          century ? rigthCentury(person) : true
+        );
```
OleksiiBlezniuk, line 29:
```suggestion
      const isMan = person.sex === 'm';
      const isRightCentury = century 
        ? rigthCentury(person)
        : true;

      return isMan && isRightCentury;
```

OleksiiBlezniuk: In that case is better to move it to constant too

src/averageAges.js
```diff
@@ -15,11 +15,26 @@
...
+    });
+
+  const menAgesSum = suitableMen
+    .reduce((prev, { born, died }) => {
+      return prev + (died - born);
+    }, 0);
+
+  return calculateAverage(menAgesSum, suitableMen.length) || 0;
```
OleksiiBlezniuk, line 37: But this reduce is used twice too 🙂  Move it to calculateAverage

src/averageAges.js
OleksiiBlezniuk: Fix it everywhere

src/averageAges.js
OleksiiBlezniuk: Fix it everywhere

src/averageAges.js
```diff
@@ -55,7 +90,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const findMother = (group, person) => {
+    const mother = group.find(mthr => mthr.name === person.mother);
+
+    return mother;
+  };
+
+  const children = people.filter((person) => {
+    const isWoman = person.sex === 'm';
```
OleksiiBlezniuk, line 100:
```suggestion
    const isMan = person.sex === 'm';
```

src/averageAges.js
```diff
@@ -55,7 +90,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    const hasMother = findMother(people, person) !== undefined;
+
+    return hasMother && (
+      onlyWithSon ? isWoman : true
+    );
```
OleksiiBlezniuk, line 105:
```suggestion
    return hasMother && onlyWithSon
      ? isWoman
      : true;
```

OleksiiBlezniuk: Or move to constant

src/averageAges.js
```diff
@@ -55,7 +90,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      const ageDiff = person.born - findMother(people, person).born;
+
+      return prev + ageDiff;
+    }, 0);
+
+  return calculateAverage(ageDiffSum, children.length);
+}
+
+function calculateAverage(array, length) {
+  return array / length;
```
OleksiiBlezniuk, line 119: Array divided by length?😮 Use better naming

OleksiiBlezniuk (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,21 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const rigthCentury = (person) => {
```
OleksiiBlezniuk, line 18: Function name should have verb in name, to understand that it's an action

OleksiiBlezniuk, line 18:
```suggestion
  const checkCentury = (person) => {
```

src/averageAges.js
```diff
@@ -37,7 +47,20 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const haveChild = (person, arr) => {
+    return arr.some(child => child.mother === person.name);
+  };
+
+  const suitableWoman = people.filter((person, idx, arr) => {
+    const isWoman = person.sex === 'f';
+    const isHasChild = (
```
OleksiiBlezniuk, line 56:
```suggestion
    const hasChild = (
```

src/averageAges.js
```diff
@@ -37,7 +47,20 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const haveChild = (person, arr) => {
```
OleksiiBlezniuk, line 50:
```suggestion
  const checkChild = (person, arr) => {
```

src/averageAges.js
```diff
@@ -55,7 +78,41 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const findMother = (group, person) => {
+    const mother = group.find(mthr => mthr.name === person.mother);
```
OleksiiBlezniuk, line 82: Well, it will be more correct here to use something like `person` instead of mother and `potentialChild` (or something like that) instead of `person` 

But best option is destructuring

OleksiiBlezniuk, line 82:
```suggestion
  const findMother = (group, { mother }) => group
    .find(({ name })=> name === mother);
```

OleksiiBlezniuk (approved):
GJ!

PR #1286, opened 2022-09-21
Solution

nikitaovch99:


Andrew-Skoryk (approved):
GJ! :+1:
I would recommend changing the naming of variables.  For example: `const avarageAge` but it's only sumOfAges. Otherwise, great work!

src/averageAges.js
```diff
@@ -15,6 +15,26 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  let manAvarageAge = people.filter((man) => {
```
Anna-Tkachenko, line 18: it's not an array with average ages, change variable name

src/averageAges.js
```diff
@@ -15,6 +15,26 @@
...
+      return Math.ceil(man.died / 100) === century;
+    });
+  }
+
+  const age = manAvarageAge.map(man => {
+    return man.died - man.born;
+  });
+  const avarageAge = age.reduce((man, prev) => {
+    return man + prev;
+  });
```
Anna-Tkachenko, line 33: you did this logic 2 times for 2 different arrays
let's create separate function that will accept an array of people and will return the average age value

src/averageAges.js
```diff
@@ -37,7 +57,25 @@ function calculateMenAverageAge(people, century) {
...
+    });
+  }
+
+  const age = women.map(woman => {
+    return woman.died - woman.born;
+  });
+
+  const sum = age.reduce((woman, prev) => {
+    return woman + prev;
+  });
```
Anna-Tkachenko, line 76: let's create separate function with this functionality

PR #1287, opened 2022-09-21
Solution

Yeliseiev:


src/averageAges.js
```diff
@@ -20,6 +20,14 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  let men;
+
+  century
+    ? men = people.filter(person => person.sex === 'm'
+      && Math.ceil(person.died / 100) === century)
+    : men = people.filter(person => person.sex === 'm');
```
Maksym845, line 28: I think you can make this assignment more beautiful, 
`men = century
   ?...
   :...`

Yeliseiev: Yep you rigth. Nice propose)

src/averageAges.js
```diff
@@ -14,12 +14,19 @@
  *
  * @return {number}
  */
+
+function calculateAverageAge(fiteredList) {
+  return fiteredList.reduce((sum, person) =>
+    sum + (person.died - person.born), 0) / fiteredList.length;
```
silchyn, line 20:
```suggestion
  return fiteredList.reduce((sum, person) => (
    sum + (person.died - person.born)
  ), 0) / fiteredList.length;
```

silchyn (changes requested):
You said that you followed the checklist, but you did not. Why?

src/averageAges.js
Yeliseiev: Yep, 2.[CODE STYLE]. 
The code was too attractive in its compact form. And I accidentally missed it like that

silchyn (approved):
Cool!

PR #1288, opened 2022-09-21
Solution

Apollonia5:


src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const males = people.filter(person => century
+    ? (Math.ceil(person.died / 100) === century && person.sex === 'm')
+    : (person.sex === 'm')
+  );
+
+  return Math.round(males.map(man => man.died - man.born)
+    .reduce((man1, man2) => man1 + man2) / males.length * 100) / 100;
```
anastasiia-nahaiska, line 24: It looks  too complicated for me, but maybe it's the good solution)

anastasiia-nahaiska (approved):

anastasiia-nahaiska (commented):
GJ!)

silchyn (changes requested):
Good points above

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const males = people.filter(person =>
+    century
+      ? (Math.ceil(person.died / 100) === century && person.sex === 'm')
+      : (person.sex === 'm')
+  );
```
silchyn, line 22:
```suggestion
  const males = people.filter(person => (
    century
      ? (Math.ceil(person.died / 100) === century && person.sex === 'm')
      : (person.sex === 'm')
  ));
```

silchyn (changes requested):
You violated the checklist rule but said the opposite. Why?

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const males = [...people].filter(person => (
```
naziamoff, line 18: is method `filter` mutable?

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
+    century
+      ? (Math.ceil(person.died / 100) === century && person.sex === 'm')
+      : (person.sex === 'm')
+  ));
+
+  return Math.round(males.map(man => man.died - man.born)
+    .reduce((man1, man2) => man1 + man2) / males.length * 100) / 100;
```
naziamoff, line 25: Actually first arg of the `reduce` cb is not the taken value but the accumulator
Proper naming helps to keep your code simpler
```suggestion
    .reduce((acc, age) => acc + age) / males.length * 100) / 100;
```

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
-  // without nesting
+  const males = people.filter(person => (
+    century
+      ? (Math.ceil(person.died / 100) === century && person.sex === 'm')
+      : (person.sex === 'm')
+  ));
+
+  return Math.round(males.map(man => man.died - man.born)
+    .reduce((acc, man2) => acc + man2) / males.length * 100) / 100;
 }
```
naziamoff, line 26: you may try to destructure man2 and in that case you won't need to use `map` here

naziamoff (changes requested):
and take a look at my previous comment, you didn't fix it everywhere
Btw woman1/woman2 are not really good options for naming.

naziamoff (approved):
🔥

PR #1289, opened 2022-09-21
solution was added

Maksym845:


src/averageAges.js
```diff
@@ -14,12 +13,24 @@
...
+    return prev + person;
+  }, 0) / agesArr.length;
+
+  return avg;
+}
+
+function calculateMenAverageAge(people, century = false) {
+  let mens = people.filter(person => person.sex === 'm');
+
+  mens = century !== false
```
Yeliseiev, line 29: I think you don`t need to check is century !== false and you don`t need to equalize it to 'false' in the function declaring

Yeliseiev (approved):
Good work!

src/averageAges.js
silchyn, line 29:
```suggestion
  mens = century
```

silchyn (changes requested):

src/averageAges.js
```diff
@@ -14,12 +13,24 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+
+function calculateAvarageAge(people) {
+  const agesArr = people.map(person => person.died - person.born);
+  const avg = agesArr.reduce((prev, person) => {
+    return prev + person;
+  }, 0) / agesArr.length;
+
+  return avg;
```
silchyn, line 23:
```suggestion
  const ages = people.map(person => person.died - person.born);
  
  return agesArr.reduce((prev, person) => {
    return prev + person;
  }, 0) / agesArr.length;
```

silchyn (changes requested):
Good job

src/averageAges.js
```diff
@@ -14,12 +13,24 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+
+function calculateAvarageAge(people) {
+  const ages = people.map(person => person.died - person.born);
+  const agesArr = ages.reduce((prev, person) => {
+    return prev + person;
+  }, 0) / ages.length;
```
naziamoff, line 21: You may do the return statement right away. And pay attention to the naming, it should perfectly describe what is stored inside
```suggestion
  const averageAge = ages.reduce((prev, person) => prev + person, 0) / ages.length;
```

naziamoff (changes requested):

naziamoff (changes requested):
I think you should pay attention to use proper names for variables. 
Those mistakes more about English, but I highly recommend to fix them
- men instead of mens
- women instead of womens
- children instead of childrens
- average instead of avarage

naziamoff (approved):
Nice 🔥

PR #1290, opened 2022-09-21
Solution

ArtemHulkoDev:


src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
+    return century
+      ? man.sex === 'm' && Math.ceil(man.died / 100) === century
+      : man.sex === 'm';
+  });
+
+  const sumOfAges = filteredMen
+    .map(man => man.died - man.born)
+    .reduce((sum, currentAge) => sum + currentAge);
+
+  return sumOfAges / filteredMen.length;
```
glugovsky, line 28: Would be nice to move it to a separate function, considering that the same operation is done in `calculateWomenAverageAge`

src/averageAges.js
```diff
@@ -37,7 +43,17 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filteredWomen = people.filter(woman => {
+    return withChildren
+      ? woman.sex === 'f' && people.some(person => person.mother === woman.name)
```
glugovsky, line 48: It is not necessary to check if the person is female, because their name is written in child's `mother` property

src/averageAges.js
Anna-Tkachenko: fix @glugovsky comment

Anna-Tkachenko (changes requested):

src/averageAges.js
ArtemHulkoDev: yeah, fixed it already, thx)

src/averageAges.js
ArtemHulkoDev: added, thx)

Anna-Tkachenko (approved):
Well Done!

PR #1291, opened 2022-09-21
Solution

korniienkokarina:


src/averageAges.js
```diff
@@ -1,61 +1,63 @@
...
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+  const menInCentury = men.filter(man => Math.ceil(man.died / 100) === century);
```
DenysKleimenov, line 12: I think it would be more accurate if you would create one array according to the condition instead of creating two anyway.

src/averageAges.js
```diff
@@ -1,61 +1,63 @@
...
  *
  * @param {object[]} people
  * @param {boolean} withChildren - optional
  *
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => person.sex === 'f');
+  const mothers = women.filter(woman => people
```
DenysKleimenov, line 28: Same as in the previous function

src/averageAges.js
```diff
@@ -1,61 +1,63 @@
...
+
+  childrenCalculated.map(person => {
+    const mother = people.find(mom => mom.name === person.mother);
+
+    if (mother) {
+      const diff = person.born - mother.born;
+
+      ageDifference.push(diff);
+    };
+  });
```
DenysKleimenov, line 58: I think you could declare an "ageDifference" as a result of **map** method, and in the map method return **diff** variable, so it would create an array and push data into array.

DenysKleimenov (approved):
LGTM, but I would do some fixes.

src/averageAges.js
Anna-Tkachenko: Agree, fix @DenysKleimenov comment

src/averageAges.js
```diff
@@ -1,61 +1,63 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+  const menInCentury = men.filter(man => Math.ceil(man.died / 100) === century);
+  const menCalculated = century ? menInCentury : men;
+  const menAge = menCalculated.map(man => man.died - man.born);
+
+  return menAge.reduce((sum, n) => sum + n) / menAge.length;
```
Anna-Tkachenko, line 16: let's create separate function that will accept people array and will return average age value

src/averageAges.js
Anna-Tkachenko: fix @DenysKleimenov comment

src/averageAges.js
```diff
@@ -1,61 +1,63 @@
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => person.sex === 'f');
+  const mothers = women.filter(woman => people
+    .find(child => child.mother === woman.name));
+  const womenCalculated = withChildren ? mothers : women;
+  const womenAge = womenCalculated.map(woman => woman.died - woman.born);
+
+  return womenAge.reduce((sum, n) => sum + n) / womenAge.length;
```
Anna-Tkachenko, line 33: you wrote the same logic 2 times, let's create separate function for it

src/averageAges.js
Anna-Tkachenko: agree, fix @DenysKleimenov comment

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1292, opened 2022-09-21
averageAges

Arsen-Shpak:


src/averageAges.js
```diff
@@ -20,6 +20,30 @@ function calculateMenAverageAge(people, century) {
...
+      break;
+
+    case 'undefined':
+      lifeOfAges = men.map((man) => man.died - man.born);
+      break;
+  }
+
+  const averageAges
+    = lifeOfAges.reduce((sumOfAges, ages) => sumOfAges + ages)
+    / lifeOfAges.length;
```
dmytristepanov, line 44: This logic was used several times, let's move it to the helper function

src/averageAges.js
```diff
@@ -20,6 +20,30 @@ function calculateMenAverageAge(people, century) {
...
+          ? man.died - man.born
+          : false;
+      });
+
+      lifeOfAges = lifeOfAges.filter((age) => age !== false);
+      break;
+
+    case 'undefined':
+      lifeOfAges = men.map((man) => man.died - man.born);
+      break;
```
dmytristepanov, line 39: I think u can do it simpler as you did in women case

dmytristepanov (approved):

silchyn (changes requested):
Good points above

Anna-Tkachenko (approved):
🔥

PR #1293, opened 2022-09-21
Solution v01

anton-iskryk:


artem-nasadchuk (approved):
as for me looks good, keep going!

src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = century
+    ? people.filter(person => person.sex === 'm')
+      .filter(man => Math.ceil(man.died / 100) === century)
```
Anna-Tkachenko, line 20: combine 2 conditions in one filter call

src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
+
+    : people.filter(person => person.sex === 'm');
+
+  const age = men.map((man) => man.died - man.born);
+  const sum = age.reduce((acc, year) => acc + year, 0);
```
Anna-Tkachenko, line 25: you did this functionality 2 times for different arrays
let's create separate function that will accept people array and will return average age value

src/averageAges.js
```diff
@@ -37,7 +42,15 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = withChildren
+    ? people.filter(person => person.sex === 'f'
+      && people.some(woman => woman.mother === person.name))
+    : people.filter(person => person.sex === 'f');
+
+  const age = women.map((woman) => woman.died - woman.born);
+  const sum = age.reduce((acc, year) => acc + year, 0);
```
Anna-Tkachenko, line 51: let's create separate function for this functionality

Anna-Tkachenko (changes requested):

src/averageAges.js
```diff
@@ -14,12 +14,20 @@
  *
  * @return {number}
  */
+function getAverageAgeValue(people) {
+  const age = people.map((person) => person.died - person.born);
```
Anna-Tkachenko, line 18: it's not age, it's array of ages, let's create more understandable variable name

src/averageAges.js
```diff
@@ -14,12 +14,20 @@
...
+  const sum = age.reduce((acc, year) => acc + year, 0);
+
+  return sum / age.length || 0;
```
Anna-Tkachenko, line 21: || 0 is redundant, you'll return 0 only if age.length will be also 0

Anna-Tkachenko (approved):

PR #1296, opened 2022-09-21
solution to average ages

dsydorenko534:


src/averageAges.js
```diff
@@ -15,11 +15,24 @@
...
+  }
+  );
+
+  const menAge = men.map((man) => {
+    return {
+      age: man.died - man.born,
+    };
+  });
+
+  const totalAge = menAge.reduce((sum, { age }) => sum + age, 0);
```
iPanda177, line 31: there's no need to create separate variable, just put condition inside reduce method

src/averageAges.js
```diff
@@ -37,7 +50,31 @@ function calculateMenAverageAge(people, century) {
...
+      child: people.find((kid) => kid.mother === human.name),
+    };
+  });
+
+  const women = peopleWithChild.filter((person) => {
+    return withChildren
+      ? person.sex === 'f' && person.child
+      : person.sex === 'f';
+  }
+  );
```
iPanda177, line 65: you can replace with just one filter method, where find all females or females with children

src/averageAges.js
```diff
@@ -55,7 +92,19 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = onlyWithSon
+    ? people.filter(child =>
+      (people.find(mother => child.mother === mother.name)
+      && child.sex === 'm'))
+    : people.filter(child =>
+      (people.find(mother => child.mother === mother.name)));
```
iPanda177, line 100: In both cases you used filter, so it can be moved out of ternary operator (just use it inside for condition to filter)

iPanda177 (changes requested):

iPanda177 (approved):
GJ

src/averageAges.js
```diff
@@ -55,7 +63,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    const mother = people.find(
+      person => person.name === child.mother);
+
+    return sum + child.born - mother.born;
+  }, 0);
+
+  return sumOfAge / children.length;
+}
+
+function findAverageAge(list) {
```
naziamoff, line 81: I don't think `list` is a proper name for a variable here. The name should describe what is stored inside. I would come up with sth like that:
```suggestion
function findAverageAge(people) {
```

PR #1299, opened 2022-09-21
functions implemented

ArturNagaycev:


src/averageAges.js
```diff
@@ -15,11 +15,16 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
```
OleksandrGlulhovskiy, line 21: you need replace if () and use another methods

OleksandrGlulhovskiy (commented):
Good job, but you need some changes))

src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
+      person =>
+        Math.ceil(person.died / 100) === century
+        && person.sex === 'm'
+    )
+    : people.filter(person => person.sex === 'm');
+
+  return men.reduce(
+    (sumAges, man) =>
+      (sumAges + (man.died - man.born)), 0
+  ) / men.length;
```
Anna-Tkachenko, line 29: let's create separate function that will accept people array and will return the average age value

src/averageAges.js
```diff
@@ -37,7 +44,17 @@ function calculateMenAverageAge(people, century) {
...
+    ? people.filter(
+      woman =>
+        people.find(person => person.mother === woman.name)
+    )
+    : people.filter(person => person.sex === 'f');
+
+  return women.reduce(
+    (sumAges, woman) =>
+      (sumAges + (woman.died - woman.born)), 0
+  ) / women.length;
```
Anna-Tkachenko, line 57: you repeated this logic 2 times, let's create separate function

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1303, opened 2022-09-21
Develop

volodymyrklymkiv:


src/averageAges.js
```diff
@@ -55,7 +76,23 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const parentFilter = people.filter(function(person) {
```
Yuriy-Komolov, line 79: Misha recommends to use arrow function instead. You used arrow function everywere, why here not?)

Yuriy-Komolov (approved):

src/averageAges.js
volodymyrklymkiv: last function was the hardest, so i wrote for better understanding for myself)
but i will fix

src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
...
+  const peopleFilter = people.filter(element =>
+    (century)
+      ? element.sex === 'm'
+      && Math.ceil(element.died / 100) === century
+      : element.sex === 'm'
+  );
+
+  return peopleFilter.reduce((totalAge, person) => {
+    return totalAge + (person.died - person.born);
+  }, 0) / peopleFilter.length;
```
Denchepornyuk, line 32: you have the same code twice, let's create a separate function for this

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1305, opened 2022-09-21
solutions

MAKSOti:


src/averageAges.js
```diff
@@ -15,11 +15,24 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let manAvarageAge = people.filter((man) => {
+    return man.sex === 'm';
+  });
```
OleksiiBlezniuk, line 20: no need to use `return` keyword here, just return value after arrow

src/averageAges.js
```diff
@@ -15,11 +15,24 @@
...
+
+  if (century) {
+    manAvarageAge = manAvarageAge.filter(man => {
+      return Math.ceil(man.died / 100) === century;
+    });
+  }
```
OleksiiBlezniuk, line 26: So if we have century - we need to filter array twice
Let's do it only once, just choose needed condition inside filter callback

src/averageAges.js
```diff
@@ -15,11 +15,24 @@
...
+
+  const age = manAvarageAge.map(man => {
+    return man.died - man.born;
+  });
+  const avarageAge = age.reduce((man, prev) => {
+    return man + prev;
+  });
```
OleksiiBlezniuk, line 33: You can use just reduce and calculate age here, let's reduce amount of array iterations

src/averageAges.js
```diff
@@ -55,7 +86,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // write code here
+  const children = people.filter(child => {
+    return people.some(mother => {
+      if (onlyWithSon) {
+        return child.sex === 'm' && child.mother === mother.name;
+      } else {
+        return child.mother === mother.name;
+      }
+    });
+  });
```
OleksiiBlezniuk, line 97: Let's make this code shorter, use ternary operator here and return value after arrow (without `return` keyword)

src/averageAges.js
```diff
@@ -55,7 +86,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const ageDifference = children.map(child => {
+    const mother = people.find(woman => woman.name === child.mother);
+
+    return child.born - mother.born;
+  });
+
+  return ageDifference.reduce((a, b) => a + b) / ageDifference.length;
```
OleksiiBlezniuk, line 104: Combine this methods

OleksiiBlezniuk (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = century
+    ? people.filter(
+      person =>
+        Math.ceil(person.died / 100) === century
+        && person.sex === 'm'
+    )
+    : people.filter(person => person.sex === 'm');
```
OleksiiBlezniuk, line 24:
```suggestion
  const men = people.filter(
      person => century
        ? Math.ceil(person.died / 100) === century && person.sex === 'm'
        : person.sex === 'm'
      );
```

OleksiiBlezniuk: no need to write filter method every time, just move ternary operator inside this method

OleksiiBlezniuk (approved):

PR #1307, opened 2022-09-21
Solution

AniaHna:


oleksandr-korobko (approved):
Goos job!)

src/averageAges.js
```diff
@@ -1,5 +1,13 @@
 'use strict';
 
+function avarageAge(people) {
+  const agesPeople = people.map(person => person.died - person.born);
+  const acc = 0;
+
+  return agesPeople.reduce((prev, current) => prev + current, acc)
+   / agesPeople.length;
```
OleksiiBlezniuk, line 8: Let's combine this methods, we can calculate age in `reduce` method, why to iterate methods twice 🙂

src/averageAges.js
```diff
@@ -37,7 +46,12 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => person.sex === 'f');
+  const womenWithChildren = women.filter(person => people.find(
+    child => child.mother === person.name
+  ));
```
OleksiiBlezniuk, line 52: Let's use one filter instead of two, we can select needed condition to return inside one filter method

src/averageAges.js
```diff
@@ -55,7 +69,17 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const womenWithChildren = people.filter(child => child.mother
+    && people.find(person => person.name === child.mother));
+
+  const withSonOnly = womenWithChildren.filter(person => person.sex === 'm');
```
OleksiiBlezniuk, line 75: Same here

src/averageAges.js
```diff
@@ -14,12 +22,13 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+  const menCentury = men
+    .filter(person => Math.ceil(person.died / 100) === century);
```
OleksiiBlezniuk, line 29: Let's use one filter instead of two, we can select needed condition to return inside one filter method

src/averageAges.js
```diff
@@ -55,7 +69,17 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    && people.find(person => person.name === child.mother));
+
+  const withSonOnly = womenWithChildren.filter(person => person.sex === 'm');
+
+  const ageDifference = arr => arr.reduce((prev, curr) =>
+    prev + curr.born - people.find(person =>
+      person.name === curr.mother).born, 0) / arr.length;
+
+  return !onlyWithSon
+    ? ageDifference(womenWithChildren) : ageDifference(withSonOnly);
```
OleksiiBlezniuk, line 82:
```suggestion
  return !onlyWithSon
    ? ageDifference(womenWithChildren)
    : ageDifference(withSonOnly);
```

OleksiiBlezniuk (changes requested):

src/averageAges.js
AniaHna: I have two different variables which need for task

src/averageAges.js
```diff
@@ -1,5 +1,11 @@
 'use strict';
 
+function avarageAge(people) {
+  return people.reduce((prev, current) => prev
+  + (current.died - current.born), 0)
+   / people.length;
```
Andr1yK, line 6: In addition, u can use variables 🙂

src/averageAges.js
```diff
@@ -14,12 +20,14 @@
...
  */
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+  const menCentury = men
```
Andr1yK, line 26: Do u really need two filters?

Andr1yK (changes requested):

src/averageAges.js
```diff
@@ -1,5 +1,13 @@
 'use strict';
 
+function avarageAge(people) {
+  const reducerPeople = people.reduce((prev, current) => prev
+  + (current.died - current.born), 0)
+   / people.length;
```
Andr1yK, line 6:
```suggestion
  return people.reduce((prev, current) => prev
    + (current.died - current.born), 0) / people.length;
```

src/averageAges.js
Andr1yK: So you've filtered array twice, but you can do it only once

Andr1yK: @AniaHna fix it

Andr1yK (changes requested):

src/averageAges.js
AniaHna: Hey, you earlier  said  i can use variables, i did it, but now you write to delete it

Andr1yK: If u create var, and in the next line return one... it is not required :)

Andr1yK (approved):

PR #1310, opened 2022-09-22
Task was completed

DenysKleimenov:


src/averageAges.js
```diff
@@ -15,11 +15,21 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menList = [...people].filter(person => (
```
korniienkokarina, line 18: you don't have to copy array, because filter returns new array
`const menList = people.filter(person => (`

korniienkokarina (commented):
everything looks nice, GJ👍

src/averageAges.js
```diff
@@ -15,11 +15,21 @@
...
+      : person.sex === 'm'
+  ));
+
+  const sumOfAge = menList.reduce((sum, currentMan) => {
+    const currentManAge = currentMan.died - currentMan.born;
+
+    return sum + currentManAge;
+  }, 0);
+
+  const averageAge = sumOfAge / menList.length;
```
Anna-Tkachenko, line 30: let's create separate function that will accept people array and will return the average age value

src/averageAges.js
```diff
@@ -55,7 +79,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ageDifferenceList = childrenList.map(child => {
+    const ageDifference = child.born - people.find(woman => (
+      woman.name === child.mother
+    )).born;
+
+    return ageDifference;
+  });
+
+  const AgeDifference = ageDifferenceList.reduce((sum, currentDiff) => (
```
Anna-Tkachenko, line 97:
```suggestion
  const ageDifference = ageDifferenceList.reduce((sum, currentDiff) => (
```

Anna-Tkachenko (changes requested):

src/averageAges.js
```diff
@@ -14,12 +14,33 @@
...
+
+// const calcAverageAgeValue = function(sum, currentPerson) {
+//   const currentPersonAge = currentPerson.died - currentPerson.born;
+
+//   return sum + currentPersonAge;
+// };
+
+const calculateAverageAge = (peopleList) => {
+  const sumOfAge = peopleList.reduce((sum, person) => {
+    const currentManAge = person.died - person.born;
```
Anna-Tkachenko, line 26: maybe person instead of man?

Anna-Tkachenko (approved):
Good job!

PR #1313, opened 2022-09-22
solution

Naveed-Shoukat:
There is some precession difference in my solution which I shown to @Ihor Nahnybida  and as advised, I changed the test from 2 decimals to Zero Decimals. Please look into my code. If it's Okay than approve it otherwise advise me fix for that. Thanks

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -15,11 +15,16 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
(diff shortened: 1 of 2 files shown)


kamal-tufekcic (approved):
I'm approving because Ihor allowed it, I tried to run your code but I was struggling to follow the logic, you have a lot of variables that could be avoided IMO, and I'm sure that your results are wrong because your many filter and map functions have an error somewhere and they are including more or less than should be included, I'd recommend using find, it was a big help for me, but that'd involve rewriting a lot of your code

PR #1314, opened 2022-09-22
implement averageAge function

SerhiiRomaniukWD:
rebooted task (git crash)
st review already was, and i made changes
now it's version for mentor review

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -46,7 +46,7 @@ function calculateMenAverageAge(people, century) {
 function calculateWomenAverageAge(people, withChildren) {
   const everyWomen = people.filter(person => person.sex === 'f');
 
-  return (withChildren === undefined)
+  return !withChildren
     ? averageAge(everyWomen)
     : averageAge(everyWomen.filter(woman => (
       people.some(person => person.mother === woman.name))));
@@ -68,10 +68,10 @@ function calculateWomenAverageAge(people, withChildren) {
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
   const children = people.filter(person => (
-    (onlyWithSon !== undefined)
-      ? people.find(woman => woman.name === person.mother)
+    onlyWithSon
+      ? people.some(woman => woman.name === person.mother)
         && person.sex === 'm'
-      : people.find(woman => woman.name === person.mother)
+      : people.some(woman => woman.name === person.mother)
   ));
 
   return children.reduce((prev, current) => (
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -7,19 +7,34 @@
  * function calculates average age only for men who died in this century
  *
  * To calculate century:
- * Divide year of person's death by 100: Math.ceil(person.died / 100)
+ * Divide year of personson's death by 100: Math.ceil(personson.died / 100)
  *
  * @param {object[]} people
  * @param {number} century - optional
  *
  * @return {number}
  */
+
+function averageAge(persons) {
+  return persons
+    .map(person => person.died - person.born)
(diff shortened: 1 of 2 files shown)


src/averageAges.js
```diff
@@ -7,19 +7,34 @@
...
 function calculateMenAverageAge(people, century) {
   // write code here
   // learn how to use array methods like .filter .map .some .every .find .reduce
   // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
+  // replace `if ()` statement with &&, || or ?:`
   // without nesting
+  const everyMan = people.filter(person => person.sex === 'm');
+  const thisCenturyPeople = everyMan.filter(person => (
+    Math.ceil(person.died / 100) === century));
```
Anna-Tkachenko, line 33: for now you create 2 people arrays for each method call
let's call correct filter method depending on century argument, create only one correct array and will work with it

src/averageAges.js
```diff
@@ -37,7 +52,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const everyWomen = people.filter(person => person.sex === 'f');
+  const womenWithChildren = people.filter(woman => (
+    people.some(person => person.mother === woman.name)));
+
```
Anna-Tkachenko, line 58: same here

Anna-Tkachenko (changes requested):

src/averageAges.js
```diff
@@ -7,19 +7,28 @@
...
  * @param {object[]} people
  * @param {number} century - optional
  *
  * @return {number}
  */
+
+function averageAge(persons) {
+  return persons
+    .map(person => person.died - person.born)
+    .reduce((firstPerson, secondPerson) => (
```
naziamoff, line 21: - you may try destructuring here, in that case you won't need to use .map method
- first arg of reduce callback is not actually its first element, but the accumulator
```suggestion
    .reduce((acc, { died, born }) => (
```

src/averageAges.js
```diff
@@ -7,19 +7,28 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const everyMan = people.filter(person => person.sex === 'm');
+
+  return (century === undefined)
```
naziamoff, line 28: You may use shortcut for boolean variables.
```suggestion
  return !century
```

naziamoff (changes requested):

naziamoff (changes requested):
take a look at my comments again pls. Looks like you haven't fixed those for all the occurrences

src/averageAges.js
```diff
@@ -7,19 +7,26 @@
...
  * @param {number} century - optional
  *
  * @return {number}
  */
+
+function averageAge(persons) {
+  return persons.reduce((acc, { died, born }) => (
+    acc + (died - born)), 0) / persons.length;
+};
+
```
naziamoff, line 22: - use proper naming. Name of the function usually starts with verb that describes kind of action it is used for
- It looks clearer this way, you should pay better attention to the checklist. Code readability really matters
```suggestion
function getAverageAge(persons) {
  return persons.reduce((acc, { died, born }) => (
    acc + (died - born)
  ), 0) / persons.length;
};

```

naziamoff (approved):
Good job 🔥 
But there are some comments, take a look at them

PR #1315, opened 2022-09-22
v1: solution

igor-1987:


AndriyZhabynets (approved):
Nice solution!

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(human => (
+    century
+      ? human.sex === 'm'
+        && Math.ceil(human.died / 100) === century
+      : human.sex === 'm'
+  ));
```
OleksiiBlezniuk, line 23:
```suggestion
  const men = people.filter(human => century
    ? human.sex === 'm'
      && Math.ceil(human.died / 100) === century
    : human.sex === 'm'
  );
```

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
+
+  const menAges = men.map(man => (man.died - man.born));
+  const menAgesSum = menAges.reduce((prev, curr) => prev + curr);
```
OleksiiBlezniuk, line 26: 1 - you can just use reduce, no need to use map too (just calculate age in reduce method)
2 - you've used same logic in next function, so better to move it to separate function and reuse it

src/averageAges.js
```diff
@@ -37,7 +43,17 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(
+    withChildren
+      ? human => human.sex === 'f'
+        && people.some(person => person.mother === human.name)
+      : human => human.sex === 'f'
+  );
```
OleksiiBlezniuk, line 51:
```suggestion
  const women = people.filter(human => withChildren
    ? human.sex === 'f'
      && people.some(person => person.mother === human.name)
    : human.sex === 'f'
  );
```

OleksiiBlezniuk: Instead of selecting callback - select condition

src/averageAges.js
```diff
@@ -55,7 +71,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(
+    onlyWithSon
+      ? child => child.mother && child.sex === 'm'
+        && people.some(woman => woman.name === child.mother)
+      : child => child.mother
+        && people.some(woman => woman.name === child.mother)
+  );
```
OleksiiBlezniuk, line 80:
```suggestion
  const children = people.filter(child => onlyWithSon
    ? child.mother && child.sex === 'm'
      && people.some(woman => woman.name === child.mother)
    : child.mother
      && people.some(woman => woman.name === child.mother)
  );
```

src/averageAges.js
OleksiiBlezniuk: fix it everywhere

OleksiiBlezniuk (changes requested):

src/averageAges.js
igor-1987: Thank you for suggestions and remarks. I guess, this is it.

src/averageAges.js
```diff
@@ -55,7 +69,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    : child.mother
+      && people.some(woman => woman.name === child.mother)
+  );
+
+  const differencesInAges = children.map(child => (
+    child.born - people.find(mother => mother.name === child.mother).born
+  ));
+
+  const differencesInAgesSum = differencesInAges.reduce(
+    (prev, curr) => prev + curr, 0
```
OleksiiBlezniuk, line 84: Ok, but this methods could be combined

OleksiiBlezniuk (approved):

PR #1318, opened 2022-09-23
Average ages solved

sergey-vorobei:


src/averageAges.js
```diff
@@ -14,30 +14,41 @@
  *
  * @return {number}
  */
+const middleAge = (arr) => arr.reduce((sum, person) =>
+  sum + (person.died - person.born), 0) / arr.length;
```
SanyaBratashchuk, line 18:
```suggestion
const getAverageAge = (arr) => arr.reduce((sum, person) =>
  sum + (person.died - person.born), 0) / arr.length;
```

SanyaBratashchuk (approved):

src/averageAges.js
sergey-vorobei: Oк. Thanks for suggest

AlisaVlasova (approved):

PR #1319, opened 2022-09-23
Solution

tatyana22:


src/averageAges.js
```diff
@@ -16,10 +16,19 @@
...
+      && Math.ceil(person.died / 100) === century)
+    : person.sex === 'm');
+
+  return getAverageAge(men);
+}
+
+function getAverageAge(peopleSort) {
+  const averageAge = peopleSort.map(person => person.died - person.born)
+    .reduce((sum, x) => sum + x, 0) / peopleSort.length;
+
```
myroslav-diiak, line 30: You can do it without map method

myroslav-diiak (approved):

src/averageAges.js
```diff
@@ -56,6 +76,24 @@ function calculateWomenAverageAge(people, withChildren) {
...
+        && child.sex === 'm'
+      : getChildren(people, child));
+
+  return children.reduce((sum, child) =>
+    sum + getDiffAge(children, people, child), 0) / children.length;
+}
+
+function getChildren(people, child) {
+  return people.find(mother => mother.name === child.mother);
+}
```
naziamoff, line 91: why do you need this function?

src/averageAges.js
```diff
@@ -56,6 +76,24 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+function getDiffAge(children, people, child) {
+  const motherBirth = people.find(mother => mother.name === child.mother);
+
+  return child.born - motherBirth.born;
```
naziamoff, line 96: You better do it inside reduce method in the return statement. Now it just overcomplicates your code.

src/averageAges.js
```diff
@@ -36,8 +45,19 @@ function calculateMenAverageAge(people, century) {
...
+    ? (person.sex === 'f'
+      && isMother(people, person))
+    : person.sex === 'f');
+
+  return getAverageAge(women);
+}
+
+function isMother(people, person) {
+  return people.find(child => child.mother === person.name);
 }
```
naziamoff, line 61: you should create separate functions only in order to not repeat yourself.
In this case you use this logic only once, so there's no need to create a separate function for that

naziamoff (changes requested):
My comments + checklist p.2

src/averageAges.js
tatyana22: we find children with this function

src/averageAges.js
tatyana22: OK!! Let it be so!

src/averageAges.js
naziamoff: I agree with Myroslav, you can make it more simple
_Hint:_ you may try destructuring inside reduce args

naziamoff (changes requested):
checklist p.2

src/averageAges.js
tatyana22: I think my code was more readable, but if you want. OK!

naziamoff (approved):
GJ

PR #1320, opened 2022-09-23
Solution

rstarovoit:


src/averageAges.js
```diff
@@ -14,12 +14,21 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people
+    .filter(person => person.sex === 'm');
+  const menCentury = men
+    .filter(person => Math.ceil(person.died / 100) === century);
```
grafit-off, line 29: Looks good, but you can just not calculate `menCentury` if there is no century passed in function. In your case you calculate `menCentury` every time, even a century was not passed in function. Maybe it's not critical, just IMHO.

src/averageAges.js
```diff
@@ -37,7 +46,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people
+    .filter(person => person.sex === 'f');
+  const womenWithChildren = women
+    .filter(person => people
+      .find(child => child.mother === person.name));
```
grafit-off, line 53: Same here, you wrote redundant functionality for getting ages 2 times

src/averageAges.js
```diff
@@ -55,7 +70,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const womenWithChil = people
+    .filter(child => child.mother && people
+      .find(person => person.name === child.mother));
+
+  const withSonOnly = womenWithChil
+    .filter(person => person.sex === 'm');
+
+  const ageDifference = array => array.reduce((sum, x) =>
+    sum + x.born - people
+      .find(person => person.name === x.mother).born, 0) / array.length;
```
grafit-off, line 82: What is `x`?

grafit-off (approved):
Good job. I've left some comments below!

src/averageAges.js
naziamoff: totally agree with Mykola, you should definitely fix this

src/averageAges.js
naziamoff: Mykola is right

src/averageAges.js
naziamoff: again, Mykola is right. Your variable names should be maximally clear. The variable name gives us a clear understanding of what is stored inside
You should never use names like `x`, they don't give any info

src/averageAges.js
rstarovoit: Absolutely agree! Many thanks for the help. Fixed it!

src/averageAges.js
rstarovoit: Thanks! Fixed it!

src/averageAges.js
rstarovoit: Thanks. Probably I left it from my draft code... Fixed it!

src/averageAges.js
```diff
@@ -14,12 +14,22 @@
  *
  * @return {number}
  */
+
+function averageAge(people) {
+  return people
+    .map(person => person.died - person.born)
+    .reduce((sum, x) => sum + x)
+    / people.length;
+}
```
naziamoff, line 23: you can avoid map here, try to do it just with reduce
And again, no `x` pls, it hurts me so bad 🥲

src/averageAges.js
```diff
@@ -55,7 +73,19 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const womenWithChil = people
```
naziamoff, line 76: I th u shouldn mak shortc li tha. Not easy to read, right? 😄 

What I meant is that you shouldn't make shortcuts like that. You should make your variable names maximally clear. What about trying to do it like that:
```suggestion
  const mothers = people
```

src/averageAges.js
```diff
@@ -55,7 +73,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const withSonOnly = womenWithChil
+    .filter(person => person.sex === 'm');
+
+  const ageDifference = array => array.reduce((sum, child) =>
+    sum + child.born - people
+      .find(person => person.name === child.mother).born, 0) / array.length;
+
+  return onlyWithSon
+    ? ageDifference(withSonOnly) : ageDifference(womenWithChil);
```
naziamoff, line 88: Looks more readable this way
```suggestion
  return onlyWithSon
    ? ageDifference(withSonOnly) 
    : ageDifference(womenWithChil);
```

naziamoff (changes requested):
Well done 🔥 
There are just some small things to make it better

naziamoff (approved):
🔥

PR #1328, opened 2022-09-24
Solution

oleksandr-korobko:


src/averageAges.js
```diff
@@ -14,12 +14,21 @@
  *
  * @return {number}
  */
+function calculateAverageAge(people) {
+  const peopleAgeMap = people.map(person => person.died - person.born);
+  const sumOfAges = peopleAgeMap
+    .reduce((accumulator, age) => accumulator + age, 0);
+
+  return sumOfAges / peopleAgeMap.length || 0;
+}
```
myroslav-diiak, line 23: You can do it without map method

myroslav-diiak (approved):

src/averageAges.js
Denchepornyuk: agree, you can do it without map

src/averageAges.js
```diff
@@ -55,7 +69,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ageDifference = children.map(child => {
+    const mother = people.find(woman => woman.name === child.mother);
+
+    return child.born - mother.born;
+  });
+
+  return ageDifference
+    .reduce((accumulator, difference) => accumulator + difference, 0)
+    / ageDifference.length;
```
Denchepornyuk, line 87: use calculateAverageAge function here

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1338, opened 2022-09-27
Solution

SanyaKovalenko:


MAKSOti (approved):
GJ
but u need read cheklist
BAD EXAMPLE:

const humans = people.map((a, b) =>
a + b);

const babies = people.map(person => children
? a
: b

GOOD EXAMPLE:

const humans = people.reduce((a, b) => (
a + b
), 0);

src/averageAges.js
```diff
@@ -15,6 +15,16 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  const men
+    = arguments.length > 1
```
OleksiiBlezniuk, line 19:
```suggestion
  const men = arguments.length > 1
```

src/averageAges.js
```diff
@@ -15,6 +15,16 @@
...
+      ? people.filter(person => Math.ceil(person.died / 100) === century
+        && person.sex === 'm')
+      : people.filter(person => person.sex === 'm');
```
OleksiiBlezniuk, line 22: No need to duplicate `filter`, just choose needed filter condition inside `filter` callback

src/averageAges.js
```diff
@@ -15,6 +15,16 @@
...
+
+  return men.reduce(
+    (prev, curr) => prev + (curr.died - curr.born), 0
+  ) / men.length;
```
OleksiiBlezniuk, line 26: You'll use same logic in next function, better to move it to separate helper function

src/averageAges.js
```diff
@@ -36,8 +46,20 @@ function calculateMenAverageAge(people, century) {
...
+//  person.mother === withChildren
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women
+    = arguments.length > 1
+      ? people.filter(
+        person =>
+          people.some(child => child.mother === person.name) === withChildren
+          && person.sex === 'f')
+      : people.filter(person => person.sex === 'f');
```
OleksiiBlezniuk, line 58: Same here, no need to duplicate filter, also pay attention to code style

OleksiiBlezniuk, line 58:
```suggestion
  const women = people.filter(person => withChildren
    ? person.sex === 'f'
    : person.sex === 'f' && people.some(child => child.mother === person.name);
```

OleksiiBlezniuk: Also why do you need to check arguments length here? Just check second (conditional) argument value

src/averageAges.js
```diff
@@ -55,7 +77,18 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  let children = people.filter(child => child.mother
+    && people.find(person => person.name === child.mother));
+
+  children = onlyWithSon
+    ? children.filter(child => child.sex === 'm')
+    : children;
```
OleksiiBlezniuk, line 85: Use filter once, just choose needed condition inside callback

OleksiiBlezniuk (changes requested):

PR #1342, opened 2022-09-27
solution

oleksii-bieliaiev:


src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => century
+    ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+    : person.sex === 'm'
+  );
+  const menLife = men.map(person => person.died - person.born);
+
+  return menLife.reduce((sum, age) => sum + age, 0) / menLife.length;
```
SanyaBratashchuk, line 24: Move this logic to separate function so you will be able to reuse it

SanyaBratashchuk (changes requested):

src/averageAges.js
oleksii-bieliaiev: tried to do it

Bohdan-Maksymiuk (approved):

PR #1350, opened 2022-09-30
Solution

vallerii:


MAKSOti (approved):
GJ
but u need read cheklist 
BAD EXAMPLE:

const humans = people.map((a, b) =>
    a + b);

const babies = people.map(person => children 
 ? a
 : b
 
GOOD EXAMPLE:

const humans = people.reduce((a, b) => (
  a + b
), 0);

src/averageAges.js
```diff
@@ -14,8 +14,20 @@
  *
  * @return {number}
  */
+function calculatePeopleAverageAge(people) {
+  const ages = people.map(person => person.died - person.born);
+
+  return ages.reduce((sum, age) => sum + age, 0) / ages.length;
```
OleksiiBlezniuk, line 20: Combine this methods, you can calculate age inside reduce method

src/averageAges.js
```diff
@@ -14,8 +14,20 @@
...
+}
+
 function calculateMenAverageAge(people, century) {
-  // write code here
+  const onlyMen = century
+    ? people.filter(men => men.sex === 'm'
+      && Math.ceil(men.died / 100) === century)
+    : people.filter(men => men.sex === 'm');
```
OleksiiBlezniuk, line 27: No need to duplicate filter methods, just choose needed condition to return inside it

src/averageAges.js
```diff
@@ -37,7 +49,12 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const onlyWomen = withChildren
+    ? people.filter(person => person.sex === 'f'
+      && people.some(woman => woman.mother === person.name))
+    : people.filter(person => person.sex === 'f');
```
OleksiiBlezniuk, line 55: Same here

src/averageAges.js
```diff
@@ -55,7 +72,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  let children = people.filter(child => {
+    return people.some(mother => child.mother === mother.name);
+  });
+
+  children = onlyWithSon
+    ? children.filter(child => child.sex === 'm')
+    : children;
```
OleksiiBlezniuk, line 81: Let's use only one filter here, no need to filter twice if `onlyWithSon` is passed

src/averageAges.js
```diff
@@ -55,7 +72,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ageDiff = children.map(child => {
+    const mother = people.find(person => person.name === child.mother);
+
+    return child.born - mother.born;
+  });
+
+  return ageDiff.reduce((sum, age) => sum + age, 0) / ageDiff.length;
```
OleksiiBlezniuk, line 89: Let's calculate age inside reduce

OleksiiBlezniuk (changes requested):

Andr1yK (approved):

PR #1361, opened 2022-10-10
Solution

Ivanenko1402:


src/averageAges.js
```diff
@@ -15,11 +15,23 @@
...
+
+  if (century) {
+    manList = people.filter(
+      person => person.sex === 'm' && Math.ceil(person.died / 100) === century,
+    );
+  } else {
+    manList = people.filter(
+      person => person.sex === 'm',
+    );
+  }
```
polosanya, line 28: `if else` is forbidden in this task. Try to use [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator#syntax)

src/averageAges.js
```diff
@@ -15,11 +15,23 @@
...
+
+  const totalAge = manList.reduce(
+    (sum, person) => sum + (person.died - person.born), 0
```
polosanya, line 31: 🔥

src/averageAges.js
```diff
@@ -37,7 +49,26 @@ function calculateMenAverageAge(people, century) {
...
+      person => person.sex === 'f'
+      && people.find(
+        child => child.mother === person.name,
+      )
+    );
+  } else {
+    womanList = people.filter(
+      person => person.sex === 'f',
+    );
+  }
```
polosanya, line 65: Same problem

src/averageAges.js
```diff
@@ -55,7 +86,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // write code here
+  let children = people.filter(
+    child => child.mother && people.find(
+      person => person.name === child.mother
+    )
+  );
+
+  if (onlyWithSon) {
+    children = children.filter(child => child.sex === 'm');
+  }
```
polosanya, line 97: please fix all cases

polosanya (changes requested):
Nice start! Please check my comments

stanislav-topikha (changes requested):

qoa11a (approved):

PR #1362, opened 2022-10-11
Solution

bagibest:


src/averageAges.js
```diff
@@ -14,17 +14,24 @@
...
+    const maleGender = man.sex === 'm';
+
+    return century ? manDiedCentury === century && maleGender : maleGender;
+  }
 
+  const men = people.filter(filteredMen);
+
+  const averageAge = men.reduce((acc, man) =>
+    acc + (man.died - man.born), 0
+  ) / men.length;
```
marina-tilniak, line 30: Extract average calculation into a function

marina-tilniak (changes requested):

PR #1366, opened 2022-10-14
averages

metinbicaksiz:


src/averageAges.js
```diff
@@ -14,12 +14,34 @@
...
  * @return {number}
  */
+
 function calculateMenAverageAge(people, century) {
   // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
```
amitpatiljc, line 20: Part of this line has code that is repeated. Could you possibly turn it into a variable or a global variable?

src/averageAges.js
```diff
@@ -14,12 +14,34 @@
...
+
+  const agesWith = [];
+
+  withCent.map(function(person) {
+    agesWith.push(person.died - person.born);
+  });
+
+  const avWithCent = agesWith.reduce((a, b) => a + b, 0) / agesWith.length;
+
+  const result = (!century) ? avWithoutCent : avWithCent;
```
amitpatiljc, line 42: by these variable names i am not able to understand what these variables are all about
and also instead of calculating all the values in different variables may be we can use a ternary operator and get the solution on 2 lines of code

amitpatiljc (changes requested):
try to optimise the code by using ternary operator at the start only and calculate variables in that only

src/averageAges.js
metinbicaksiz: I have applied the same changes to the ret of the code, too

src/averageAges.js
```diff
@@ -56,6 +103,34 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  people.filter((a) => {
+    for (let i = 0; i < people.length; i++) {
+      if (a.sex === 'f'
+          && a.name === people[i].mother
+          && people[i].sex === 'm') {
+        ageDiffSon.push(people[i].born - a.born);
+      }
+    }
+  });
+
```
amitpatiljc, line 128: avoid if condition

amitpatiljc (approved):
Well done !

PR #1373, opened 2022-10-17
Solution

alexxxstep:


src/averageAges.js
```diff
@@ -15,11 +15,26 @@
...
+  const men = people.filter((p) => p.sex === 'm');
+
+  const menCentury = men.reduce((acc, m) => {
+    const centuryManDied
+      = m.died % 100 === 0 ? m.died / 100 : Math.floor(m.died / 100 + 1);
+
+    const isCorrectCentury = century ? centuryManDied === century : false;
+
+    return isCorrectCentury ? [...acc, m] : acc;
+  }, []);
```
polosanya, line 27: You can just use [Math.ceil](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Math/ceil) to avoid all these checks

src/averageAges.js
```diff
@@ -15,11 +15,26 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter((p) => p.sex === 'm');
+
+  const menCentury = men.reduce((acc, m) => {
```
polosanya, line 20: Please use more informative names for your variables

src/averageAges.js
```diff
@@ -55,7 +92,34 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    return mother
+      ? [
+        ...acc,
+        {
+          ...child,
+          ageMotherDiff: diff,
+        },
+      ]
+      : acc;
+  }, []);
```
polosanya, line 113: Looks very complicated. This task can be solved much easier. Try to find another solution

polosanya (changes requested):
Good job! But so far your solution looks a bit confusing. Try to find a simpler solution and feel free to ask questions in your fe_chat if you have any difficulties.

liliia-livandovskaya (approved):

PR #1374, opened 2022-10-18
Solution

bokorzita:


src/averageAges.js
```diff
@@ -55,7 +75,23 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  let filteredByChildren = people.filter(
+    x => people.some(c => c.name === x.mother));
```
jstmpelowycz, line 79: I'd rather create such a function (I'll need it later on)
```
const getMother = (mother) => people.find(
  ({ name }) => name === mother,
);
```
if `getMother` return `undefined`, it becomes `false`, otherwise – `true`
```suggestion
  let children = people.filter(({ mother }) => getMother(mother));
```

src/averageAges.js
```diff
@@ -55,7 +75,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  filteredByChildren = onlyWithSon
+  ? filteredByChildren.filter(
+    x => x.sex === 'm')
+  : filteredByChildren;
```
jstmpelowycz, line 84: let's use the same approach as we did with men

src/averageAges.js
```diff
@@ -55,7 +75,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ageDifferencesArray
+    = filteredByChildren.map(
+      x => x.born - people.find(
+        c => c.name === x.mother
+      ).born
+    );
```
jstmpelowycz, line 91: 1) no need to add `...Array` postfix at the end, `...s` post refers to array itself
2) also, here we need our `getMother` function
```suggestion
  const ageDifferences = children.map(child => {
    const mother = getMother(mother);
    
    return child.born - mother.born;   
  });
```

src/averageAges.js
```diff
@@ -55,7 +75,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  return ageDifferencesArray.reduce((p, c) => p + c)
+  / filteredByChildren.length;
```
jstmpelowycz, line 94: let's use our dedicated function I've suggested using up above

jstmpelowycz (changes requested):
nicely done! let's optimize our code

src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
...
+
+  const newPeople
+    = people.filter(
+      x =>
+        century
+          ? Math.ceil(x.died / 100) === century && x.sex === 'm'
+          : x.sex === 'm');
+  const agesArray = newPeople.map(x => x.died - x.born);
+
+  return agesArray.reduce((p, c) => p + c) / newPeople.length;
```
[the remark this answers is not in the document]
amitpatiljc: what do you mean by p and c?
you can name them so that it can be self explainatory

src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  const newPeople
+    = people.filter(
+      x =>
+        century
+          ? Math.ceil(x.died / 100) === century && x.sex === 'm'
+          : x.sex === 'm');
```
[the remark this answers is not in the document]
amitpatiljc: for example
`const isMale = (person) => person.sex === 'm';`
you can use it like this 
`const men = people.filter(person => isMale(person));`

amitpatiljc (changes requested):
Please go through the examples provided

src/averageAges.js
```diff
@@ -15,47 +15,62 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let males = people.filter(({ sex }) => sex === 'm');
```
amitpatiljc, line 18: Part of this line has code that is repeated. Could you possibly turn it into a variable or a global variable?

`const isMale = (person) => person.sex === 'm';`
this can be used as a global variable outside any particular function, so that you can use it everywhere

src/averageAges.js
```diff
@@ -15,47 +15,62 @@
...
- *
- * @param {object[]} people
- * @param {boolean} withChildren - optional
- *
- * @return {number}
- */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let women = people.filter(({ sex }) => sex === 'f');
+
```
amitpatiljc, line 38: Part of this line has code that is repeated. Could you possibly turn it into a variable or a global variable?

`const isFemale = (person) => person.sex === 'f';`

src/averageAges.js
```diff
@@ -15,47 +15,62 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let males = people.filter(({ sex }) => sex === 'm');
+
+  if (century) {
+    males = males.filter(male => Math.ceil(male.died / 100) === century);
+  }
+
```
amitpatiljc, line 23: in the checklist point it is mentioned that 
` replace `if ()` statement with &&, || or ?:`
so remove all if conditions and replace with `turnery` condition

src/averageAges.js
```diff
@@ -15,47 +15,62 @@
...
+  const averageAges = males.map(
+    male => {
+      return male.died - male.born;
```
amitpatiljc, line 26: this can also be global variable 
`const calculateAge = (person) => person.died - person.born;`

src/averageAges.js
```diff
@@ -15,47 +15,62 @@
...
+    })
+    .reduce(
+      (ageOfPreviousMale, ageOfCurrentMale) =>
+        ageOfPreviousMale + ageOfCurrentMale
+    ) / males.length;
```
amitpatiljc, line 31: this reduce method as well can be used as a global variable
`const averageAgeCalculate = (ages) =>
  ages.reduce((sum, age) => sum + age, 0) / ages.length;`

src/averageAges.js
```diff
@@ -15,47 +15,62 @@
...
+  }
+
+  const ageOfWomen = women.map(
+    woman => woman.died - woman.born)
+    .reduce(
+      (ageOfPreviousWoman, ageOfCurrentWoman) =>
+        ageOfPreviousWoman + ageOfCurrentWoman
+    ) / women.length;
+
+  return ageOfWomen;
```
amitpatiljc, line 51: and while returning you need to round age in all 3 instances

amitpatiljc (changes requested):

src/averageAges.js
```diff
@@ -15,47 +15,62 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
```
amitpatiljc, line 21: you need to replace if () as per this checklist point , please do that and all good

amitpatiljc (changes requested):
replace `if ()` statement with &&, || or ?:

src/averageAges.js
jstmpelowycz: let's do that – you can put `century` validation into `people.filter(...)`

src/averageAges.js
jstmpelowycz: yes, let's create a dedicated function up above to use whenever we'd like to

src/averageAges.js
jstmpelowycz: agree, let's fix it, you'll need the same function ahead in your code here

src/averageAges.js
```diff
@@ -15,47 +15,62 @@
...
- * @return {number}
- */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let women = people.filter(({ sex }) => sex === 'f');
+
+  if (withChildren) {
+    women = women.filter(woman => people.some(
+      person => person.mother === woman.name));
+  }
```
jstmpelowycz, line 42: let's use ternary alongside filtering (as in the previous case with men)

src/averageAges.js
jstmpelowycz, line 51: here we need that function I mentioned to calculate average age of a group of people (male, female – whatever

src/averageAges.js
```diff
@@ -15,47 +15,62 @@
...
-  // write code here
+  const getMother = (mother) => people.find(
+    ({ name }) => name === mother,
+  );
+
+  let children = people.filter(({ mother }) => getMother(mother));
+
+  if (onlyWithSon) {
+    children = children.filter(child => child.sex === 'm');
+  }
```
jstmpelowycz, line 63: let's use ternary alongside filtering

src/averageAges.js
```diff
@@ -15,47 +15,62 @@
...
+
+  const ageDifferences = children.map(child => {
+    return child.born - getMother(child.mother).born;
+  });
+
+  return ageDifferences.reduce(
+    (ageOfPreviousChild, ageOfCurrentChild) =>
+      ageOfPreviousChild + ageOfCurrentChild
+  )
+  / children.length;
```
jstmpelowycz, line 73: we can use a dedicated function here to calculate our ages

jstmpelowycz (changes requested):
nicely done! let's fix all comments and all good

jstmpelowycz (approved):
great job, Zita!

PR #1391, opened 2022-10-28
Solution

taraselnik:


src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const getListMen = people.filter(person => person.sex === 'm', 0);
```
iryna-lelyk, line 23: You don't have to specify the index (as a second parameter) in filter method if index = 0. Also you can create helper functions to get people by their sex (males, females) and to get average age. Because you will be able to use them in another functions in this task.

src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
...
+
+  const listToCalculate = century
+    ? getListMen.filter(
+      person => Math.ceil(person.died / 100) === century)
+    : getListMen;
+
+  const getMenAges = listToCalculate.map((person) => person.died - person.born);
+
+  return getMenAges.reduce((total, item) => total + item) / getMenAges.length;
```
iryna-lelyk, line 32: You can do this in one iteration. Only using method reduce.
```suggestion
  const getMenAges = listToCalculate.map((person) => person.died - person.born);

  return getMenAges.reduce((person, item) => total + (person.died - person.born)) / getMenAges.length;
```

src/averageAges.js
```diff
@@ -37,7 +47,17 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const listToCalculate = withChildren
+    ? people.filter(
+      person => person.sex === 'f' && people.some(
+        child => child.mother === person.name))
+    : people.filter(person => person.sex === 'f', 0);
```
iryna-lelyk, line 54: Here you will be able to use your helper function ;)

iryna-lelyk (approved):
Good job!

src/averageAges.js
taraselnik: 1. with 0 agree 2. helper function - i thought about, but we have to create several different functions in this task

src/averageAges.js
taraselnik: this code can't work? but in common you are right? it can be only reduce )

src/averageAges.js
taraselnik: because of human rights i cant divide people by sex before ) Besides, I coded separate functions as they asked in task (with strong wish to make "Helper"). 
and if it will be super needed i will redo.

illiarerun, line 54: you don't need second param for filter
```suggestion
    : people.filter(person => person.sex === 'f');
```

src/averageAges.js
```diff
@@ -37,7 +47,17 @@ function calculateMenAverageAge(people, century) {
...
+
+  const getWomenAges = listToCalculate.map(
+    (person) => person.died - person.born);
```
illiarerun, line 57: you can do it in reduce function

illiarerun (approved):
Left few comments:

PR #1392, opened 2022-10-29
solution

Vasyliy22:


src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
+
+  const ageInEighteen = onlyMen.filter(item =>
+    Math.ceil(item.died / 100) === century)
+    .map(item => item.died - item.born);
+
+  return ageInEighteen.reduce((sum, age) =>
+    sum + age / ageInEighteen.length, 0)
+    || onlyMen.map(item => item.died - item.born)
+      .reduce((sum, age) =>
+        sum + age / onlyMen.length, 0);
```
Anastasiia-Svintsova, line 30: It hard to read
```suggestion
  return (
    ageInEighteen.reduce((sum, age) => sum + age / ageInEighteen.length, 0) ||
    onlyMen
      .map((item) => item.died - item.born)
      .reduce((sum, age) => sum + age / onlyMen.length, 0)
  );
```
or better write it to a variable and return this variable

src/averageAges.js
```diff
@@ -37,7 +45,12 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = withChildren
+    ? people.filter(woman => people.some(kid => kid.mother === woman.name))
+    : people.filter(woman => woman.sex === 'f');
+
+  return women.map(item => item.died - item.born)
+    .reduce((sum, age) => sum + age / women.length, 0);
```
Anastasiia-Svintsova, line 53: create a separate method to calculate an average age, you use it in different places

Anastasiia-Svintsova (changes requested):
Fix comments

Anastasiia-Svintsova (approved):

PR #1415, opened 2022-10-31
solution

savchuksavchuk:


src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let men = [...people];
+
+  century === undefined
+    ? men = men.filter(person => person.sex === 'm')
+    : men = men.filter(person => person.sex === 'm'
+      && Math.ceil(person.died / 100) === century);
```
polosanya, line 23:
```suggestion
  const men = century === undefined
    ? people.filter(person => person.sex === 'm')
    : people.filter(person => person.sex === 'm'
      && Math.ceil(person.died / 100) === century);
```
Btw, you do filter in both cases, so you can just implement different logic inside filter callback

src/averageAges.js
```diff
@@ -37,7 +44,19 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let women = people.filter(person => person.sex === 'f');
+
+  if (withChildren) {
```
polosanya, line 49: replace `if ()` statement with &&, || or ?:

src/averageAges.js
```diff
@@ -55,7 +74,28 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  let children = [...people];
```
polosanya, line 77: You are not using mutable methods, so you can use people directly

src/averageAges.js
```diff
@@ -55,7 +74,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  if (onlyWithSon) {
```
polosanya, line 79: Same here and below. `If` is forbidden in this task

polosanya (changes requested):
Good job! Please check my comments

src/averageAges.js
```diff
@@ -37,7 +42,19 @@ function calculateMenAverageAge(people, century) {
...
+
+  const women = withChildren === undefined
+    ? people.filter(person => person.sex === 'f')
+    : people.filter(person => (
+      person.sex === 'f' && mothers.includes(person.name)
+    ));
+
+  const averageAge = women.map(person => person.died - person.born)
+    .reduce((a, b) => a + b)
+    / women.length;
```
polosanya, line 55:
```suggestion
  const averageAge = women
    .map(person => person.died - person.born)
    .reduce((a, b) => a + b)
      / women.length;
```
I think this formatting looks better.

src/averageAges.js
```diff
@@ -55,7 +72,24 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  let ageDifference = children.map(child => getAgeDifference(child));
+
+  function getAgeDifference(child) {
+    const mother = people.find(person => person.name === child.mother);
+
+    return mother && child.born - mother.born;
+  }
+
+  ageDifference = ageDifference.filter(el => el);
```
polosanya, line 87: What's the purpose of this filtering? Do you really need it?

polosanya (changes requested):
Almost perfect!

src/averageAges.js
savchuksavchuk: The getGeDifference function returns age difference (number) or 'undefined'. So i need to get rid of 'undefined' in the array to avoid mistakes calculating averageAgeDifference.

src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = century === undefined
+    ? people.filter(person => person.sex === 'm')
+    : people.filter(person => person.sex === 'm'
+      && Math.ceil(person.died / 100) === century);
+
+  const averageAge = men.map(person => +person.died - +person.born)
+    .reduce((a, b) => a + b)
+    / men.length;
```
loralevitska, line 25: Move to helper function to avoid code duplicate

loralevitska (changes requested):
Almost done)

PR #1418, opened 2022-11-01
Average ages solution

shtht24:


MikhailKucherina (approved):
Nice job!

src/averageAges.js
```diff
@@ -37,7 +51,23 @@ function calculateMenAverageAge(people, century) {
...
+
+  people.forEach(person => mothers.push(person.mother));
+
+  let women = [];
+
+  withChildren
+    ? women = people.filter(
+      person =>
+        person.sex === 'f'
+        && mothers.includes(person.name)
```
illiarerun, line 64: use find or some to check if woman has children

src/averageAges.js
```diff
@@ -55,7 +85,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = !onlyWithSon
+    ? people.filter(child =>
+      (people.find(person => child.mother === person.name))
+    )
+    : people.filter(child =>
+      (people.find(person => child.mother === person.name && child.sex === 'm'))
+    );
```
illiarerun, line 94: you basically do the same in these 2 filters, better move your ternary inside
```suggestion
  const children = people.filter(child => onlyWithSon
    ? (people.find(person => child.mother === person.name  && child.sex === 'm'))
    : (people.find(person => child.mother === person.name)
  )
```

src/averageAges.js
```diff
@@ -55,7 +85,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ageDiff = children.map(child => (
+    child.born - people.find(person => (
+      person.name === child.mother)).born));
+
+  return ageDiff.reduce((a, b) => a + b) / ageDiff.length;
```
illiarerun, line 100: would be totally great to have helper here

illiarerun (changes requested):

src/averageAges.js
shtht24: ty for review and great advices. it`s clear for me now :)

src/averageAges.js
```diff
@@ -20,6 +20,18 @@ function calculateMenAverageAge(people, century) {
...
+
+  const men = century
+    ? people.filter(
+      person =>
+        person.sex === 'm'
+        && Math.ceil(person.died / 100) === century
+    )
+    : people.filter(person => person.sex === 'm');
+
+  const menAges = men.map(person => (person.age = person.died - person.born));
```
denys-danyliuk, line 32: you can create helper function for that

denys-danyliuk, line 32:
```suggestion
  const menAges = men.map(person => (person.died - person.born));
```

PR #1421, opened 2022-11-01
Solution

Esceype:


src/averageAges.js
```diff
@@ -15,6 +15,25 @@
...
+  }
+
+  const getCenturyOfDied = getMen.filter(person =>
+    century === Math.ceil(person.died / 100));
+  const calculateAge = getCenturyOfDied.map(person =>
+    person.died - person.born);
+  const calculateAvgAge = calculateAge.reduce((a, b) =>
+    a + b, 0) / calculateAge.length;
+
+  return calculateAvgAge;
```
Largoxoxo, line 35: Here you basically do the same calculations for average age twice. And you need to do them even more times. I think it's better to create a function for this.

src/averageAges.js
```diff
@@ -37,7 +56,25 @@ function calculateMenAverageAge(people, century) {
...
+    const calculateAge2 = getWomen.map(person =>
+      person.died - person.born);
+    const calculateAvgAge2 = calculateAge2.reduce((a, b) =>
+      a + b, 0) / calculateAge2.length;
+
+    return calculateAvgAge2;
+  }
+
+  const getWomenWithChildren = getWomen.filter(person =>
+    people.some(child => child.mother === person.name));
```
Largoxoxo, line 71: I don't understand it fully. I think some is an incorrect method to use here. Some returns true if at least one element of array pass the check. So, it won't cycle through the whole array it will return true even if only one item in the array is truthy. You are using it just to create an additional iteration over the array, but you can use a better method here, like map or find.

src/averageAges.js
```diff
@@ -15,6 +15,25 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  const getMen = people.filter(person => person.sex === 'm');
+
+  if (arguments.length === 1) {
+    const calculateAge1 = getMen.map(person => person.died - person.born);
+    const calculateAvgAge1 = calculateAge1.reduce((a, b) =>
```
Largoxoxo, line 22: Bad variable name.

src/averageAges.js
```diff
@@ -55,7 +92,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const womenAndChildren = onlyWithSon
+    ? people.filter(person => person.sex === 'm'
+      && people.find(mother => person.mother === mother.name))
+    : people.filter(person =>
+      people.find(mother => person.mother === mother.name));
+
+  const difference = womenAndChildren.map(child => (
```
Largoxoxo, line 101:
```suggestion
  const ageDifference = womenAndChildren.map(child => (
```

src/averageAges.js
```diff
@@ -55,7 +92,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    child.born - people.find(mother => child.mother === mother.name).born)
+  );
+
+  const calcDiff = difference.reduce((a, b) => a + b, 0) / difference.length;
```
Largoxoxo, line 105:
```suggestion
  const calcAgeDifference = difference.reduce((a, b) => a + b, 0) / difference.length;
```

Largoxoxo (approved):

Largoxoxo (approved):

Olexandr6 (approved):

PR #1422, opened 2022-11-01
Solution

Oleksii-Siviolov:


src/averageAges.js
```diff
@@ -14,12 +14,23 @@
...
+
+const listPeopleBySex = (people, sex) => {
+  return people.filter(person => person.sex === sex);
+};
+
+const calculateAverageAge = (people) => {
+  return people.reduce((sumOfAges, person) =>
+    (person.died - person.born) + sumOfAges, 0) / people.length;
+};
+
```
kryvets-vadym, line 26: good approach to the solution, but in this case it may be redundant

src/averageAges.js
```diff
@@ -55,7 +73,19 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const listOfChildren = people.filter(child =>
+    people.find(person => person.name === child.mother));
+
+  const pickChildren = onlyWithSon
+    ? listPeopleBySex(listOfChildren, 'm')
+    : listOfChildren;
```
kryvets-vadym, line 81: here you can use one variable and check for `onlyWithSon` using the conditional operator `?`

kryvets-vadym (approved):
Good job, keep up the good work!

src/averageAges.js
```diff
@@ -14,12 +14,23 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const pickMen = century
+    ? listPeopleBySex(people, 'm').filter(men =>
+      Math.ceil(men.died / 100) === century)
+    : listPeopleBySex(people, 'm');
```
illiarerun, line 31: let's create separate variable for `listPeopleBySex` result and then filter/not filter it here so you won't repeat yourself

src/averageAges.js
```diff
@@ -37,7 +48,14 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const getMothersList = people.map(person => person.mother);
+
+  const pickWomen = withChildren
+    ? listPeopleBySex(people, 'f').filter(women =>
+      getMothersList.includes(women.name))
+    : listPeopleBySex(people, 'f');
```
illiarerun, line 56: same here

src/averageAges.js
```diff
@@ -55,7 +73,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const pickChildren = onlyWithSon
+    ? listPeopleBySex(listOfChildren, 'm')
+    : listOfChildren;
+
+  const ageDifference = pickChildren.reduce(
+    (sum, child) => sum + child.born - people.find(
+      mother => mother.name === child.mother).born, 0)
+    / pickChildren.length;
+
```
illiarerun, line 87:
```suggestion
  const ageDifference = pickChildren.reduce((sum, child) => (
    sum + child.born - people.find(mother => (
      mother.name === child.mother
    )).born
  ), 0) / pickChildren.length;

```

illiarerun (changes requested):

denys-danyliuk (approved):

PR #1423, opened 2022-11-01
done

MXMUK:
js_average-ages

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -14,12 +14,29 @@
  *
  * @return {number}
  */
+
+const calculateAverageAge = (people) => {
+  return (
+    people.reduce((prev, person) => person.died - person.born + prev, 0)
+    / people.length
+  );
+};
+
 function calculateMenAverageAge(people, century) {
   // write code here
   // learn how to use array methods like .filter .map .some .every .find .reduce
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  let filteredPeople = people.filter((person) => person.sex === 'm');
+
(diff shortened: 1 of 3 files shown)


src/averageAges.js
```diff
@@ -14,12 +14,29 @@
...
+  );
+};
+
 function calculateMenAverageAge(people, century) {
   // write code here
   // learn how to use array methods like .filter .map .some .every .find .reduce
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  let filteredPeople = people.filter((person) => person.sex === 'm');
```
Stepico, line 31: I guess it's not really good name, what if we have to create an additional variable for women?

Stepico (approved):
GJ, clear code, I like it, just one minor piece of advice from me about naming

src/averageAges.js
mykola-koval: agree, bad variable name

src/averageAges.js
```diff
@@ -14,12 +14,29 @@
...
+
+  if (century) {
+    filteredPeople = filteredPeople.filter(
+      (person) => Math.ceil(person.died / 100) === century
+    );
```
mykola-koval, line 36: let's use a ternary operator for this logic, instead of using let

src/averageAges.js
mykola-koval: and let's create a helper function for filtering people by gender

mykola-koval (changes requested):

mykola-koval (approved):

PR #1430, opened 2022-11-02
done

AndreyChernysh:


src/averageAges.js
```diff
@@ -14,12 +14,25 @@
  *
  * @return {number}
  */
+
+const filterByGender = (gender, people) => {
+  return people.filter(({ sex }) => sex === gender);
```
yaro-bey, line 19: here you could just make variable - each element of people (lets say person) and since its an object check the value of the key 'sex'

yaro-bey (approved):
GT!

mykola-koval (approved):

PR #1432, opened 2022-11-02
Solution

Bohdan-Kovalenko1992:


src/averageAges.js
```diff
@@ -37,7 +42,17 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let womenList = people.filter(person => person.sex === 'f');
+
+  womenList = withChildren
+    ? womenList.filter(women => people.some(person =>
```
Largoxoxo, line 48: I just don't get it how using some is a valid check here. I think some is an incorrect method to use here. Some returns true if at least one element of array pass the check. So, it won't cycle through the whole array it will return true even if only one item in the array is truthy. You are using it just to create an additional iteration over the array, but you can use a better method here, like map.

Largoxoxo (approved):

Andriy-Shymkiv (changes requested):

Olexandr6 (approved):

Largoxoxo (approved):

PR #1438, opened 2022-11-06
average-ages

SashaKornieiev:


src/averageAges.js
```diff
@@ -55,9 +62,35 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = !onlyWithSon
+    ? people.filter(child =>
+      people.find(mother => child.mother === mother.name))
```
vlad-demchuk, line 67: image

vlad-demchuk: or better use syntax  with braces  for better readability

vlad-demchuk (changes requested):

liliia-livandovskaya (approved):

PR #1441, opened 2022-11-09
Solution

webdevw:


src/averageAges.js
```diff
@@ -14,8 +14,23 @@
...
+  return Math.round((sumAges / arrayOfReduce.length) * 100) / 100;
+}
+
 function calculateMenAverageAge(people, century) {
-  // write code here
+  const arrayPeople = century
+    ? people
+      .filter(item => (Math.ceil(item.died / 100)) === century)
+      .filter(({ sex }) => sex === 'm')
+      .map((item) => item.died - item.born)
```
polosanya, line 28: I think it's a bad idea to call a person an item)
Also, do you really need to call filter couple of times? Can't you do all the checks in one iteration?
And if you use map, then it's not arrayPeople anymore, but array of ages

src/averageAges.js
```diff
@@ -55,7 +78,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = onlyWithSon
+    ? people
+      .filter((item) => item.mother != null)
+      .filter(item => item.sex === 'm')
+      .filter(item => people.some(({ name }) => name === item.mother))
+    : people
+      .filter((item) => item.mother != null)
+      .filter(item => people.some(({ name }) => name === item.mother));
```
polosanya, line 88: Same here

src/averageAges.js
```diff
@@ -55,7 +78,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const arrayYears = children.map((item) => {
+    const mother = people.find((element) => element.name === item.mother);
```
polosanya, line 91: Stop calling people by such strange names)

polosanya (changes requested):
Good job! Please pay attention to my comments

qoa11a (approved):

PR #1444, opened 2022-11-10
Solution

Bohdan-Hutsuliak:


src/averageAges.js
```diff
@@ -20,6 +20,12 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const men = people.filter(person => (!century)
+    ? person.sex === 'm'
```
marina-tilniak, line 24: good job 👍  you can also move to variable all men. Code will be simpler in this way

marina-tilniak (approved):

PR #1451, opened 2022-11-14
solution

olenitut:


src/averageAges.js
```diff
@@ -37,7 +40,17 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let women = people.filter(person => person.sex === 'f');
+
+  if (withChildren) {
+    women = women.filter(woman => {
+      return people.some(person => person.mother === woman.name);
+    });
+  }
```
Bohdan-Maksymiuk, line 49: use ternary here instead of filtering twice

Bohdan-Maksymiuk (approved):

PR #1455, opened 2022-11-16
Implemented all 3 solutions with the use of array methods

zynkro:


src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter((person) =>
+    century
+      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+      : person.sex === 'm'
```
amitpatiljc, line 21: Part of this line has code that is repeated. Could you possibly turn it into a global variable?
somthing like this 
`const isMale = (person) => person.sex === 'm';`

src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
+  );
+
+  const sumAges = men.reduce(
+    (sum, person) => sum + (person.died - person.born),
+    0
+  );
+
+  return Math.round((sumAges / men.length) * 100) / 100;
```
amitpatiljc, line 29: Possible variable as well?

amitpatiljc (changes requested):

src/averageAges.js
zynkd: Hi Amit, thank you for the suggestions. I tried to implement them. Let me know if all is good. Cheers!

amitpatiljc (approved):
Well Done!

PR #1457, opened 2022-11-19
Solution

Nazar-Brunarskyi:


romanblyashuk94 (approved):
Good Sollutions!

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const callBack = century
+    ? person => person.sex === 'm' && Math.ceil(person.died / 100) === century
+    : person => person.sex === 'm';
```
solaryasha, line 20:
```suggestion
  const checkIfTheMan = century
    ? person => person.sex === 'm' && Math.ceil(person.died / 100) === century
    : person => person.sex === 'm';
```

src/averageAges.js
```diff
@@ -37,7 +41,21 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothers = withChildren
+    ? people
+      .filter(person => person.mother)
+      .map(person => person.mother)
```
solaryasha, line 47: Could we do this with the one`reduce` usage?

src/averageAges.js
```diff
@@ -37,7 +41,21 @@ function calculateMenAverageAge(people, century) {
...
+    : null;
+
+  const callBack = withChildren
```
solaryasha, line 50:
```suggestion
  const checkIfHasTheChild = withChildren
```

src/averageAges.js
```diff
@@ -55,7 +73,32 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    .map(person => {
+      const mother = people.find(per => per.name === person.mother);
+
+      const difference = mother
+        ? person.born - mother.born
+        : false;
+
+      return difference;
+    })
+    .filter(age => age);
```
solaryasha, line 90: Why do you need to filter it?

src/averageAges.js
```diff
@@ -55,7 +73,32 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const peopleCopy = onlyWithSon
+    ? people.filter(person => person.sex === 'm')
+    : [...people];
+
+  const ageDifferenceArr = peopleCopy
+    .map(person => {
```
solaryasha, line 81: Maybe, let's use `filter` here ?

solaryasha (changes requested):

src/averageAges.js
Nazar-Brunarskyi: I can't get your point why do I need filter here

src/averageAges.js
solaryasha: But why do you need to map the array with a mother object and possibly a false value,
if you could create an array with mothers with one filter?

solaryasha (changes requested):

solaryasha (approved):

PR #1459, opened 2022-11-20
Solution

Oleksandr-Filo:


shyptia (approved):
For me everything is OK

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => (
+    century === undefined
+      ? person.sex === 'm'
+      : person.sex === 'm' && Math.ceil(person.died / 100) === century));
```
valerii-derkach, line 21: better inverse this condition, like u did with  withChildren (45 line)

src/averageAges.js
```diff
@@ -37,7 +41,16 @@ function calculateMenAverageAge(people, century) {
...
+  const women = people.filter(person => (
+    withChildren
+      ? people.find(child => person.name === child.mother)
+      : person.sex === 'f'
+  ));
+  const womenAges = women.map(person => person.died - person.born);
+  const totalWomenAge = womenAges.reduce((sum, age) => sum + age, 0);
+  const averageWomenAge = totalWomenAge / women.length;
+
+  return averageWomenAge;
```
valerii-derkach, line 53: DRY
let's create a callback function and move this logic there, you repeat the same calculations thee times
```suggestion
  const totalWomenAge = womenAges.reduce((sum, age) => sum + age, 0);
  const averageWomenAge = totalWomenAge / women.length;

  return averageWomenAge;
```

valerii-derkach (changes requested):

src/averageAges.js
Oleksandr-Filo: done

src/averageAges.js
Oleksandr-Filo: thanks for advice, done

valerii-derkach (approved):
🚀

PR #1461, opened 2022-11-20
Solution

Dmytro-Kravchenko:


src/averageAges.js
```diff
@@ -37,7 +42,16 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filteredObj = people
+    .filter(person => withChildren
+      ? people.find(kid => kid.mother === person.name) && person.sex === 'f'
+      : person.sex === 'f');
+
+  const womenAverageAge = filteredObj
+    .map(person => person.died - person.born)
+    .reduce((sum, current) => sum + current) / filteredObj.length;
```
tetiana-ivanchenko, line 52:
```suggestion
    .reduce((sum, person) => sum + person.died - person.born, 0) / filteredObj.length;
```

tetiana-ivanchenko: and also move it to separate method, you use it more that once

tetiana-ivanchenko (changes requested):

src/averageAges.js
Dmytro-Kravchenko: The function will run faster, but code will be harder to understand

tetiana-ivanchenko: what did you mean? why it's harder to understand?

tetiana-ivanchenko (approved):

src/averageAges.js
Dmytro-Kravchenko: Sorry, I was wrong, it's not hard to understand. In the 3rd commit, I made the same change to the "calculateAverageAgeDiff" function.

PR #1463, opened 2022-11-23
Solution

Khabailiuk:


src/averageAges.js
```diff
@@ -36,8 +49,29 @@ function calculateMenAverageAge(people, century) {
...
+    / ageWomenWithChildren.length;
+
+    return avarageAgeWomenWithChildren;
+  } else {
+    const ageWomen = filteredWomen.map((person) => person.died - person.born);
+    const sumAgeWomen = ageWomen.reduce((age1, age2) => age1 + age2);
+    const avarageAgeWomen = sumAgeWomen / ageWomen.length;
+
+    return avarageAgeWomen;
+  }
```
tetiana-ivanchenko, line 74: your algorithm is not so concise, that can be. you can use only
const filteredWomen =  // and here with ternary operator create condition for filtering. the same with prev function
and then, move code that cal average age into separate method, and use it here and in prev function.

tetiana-ivanchenko (changes requested):

src/averageAges.js
```diff
@@ -14,14 +14,15 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
-}
+  const arrayOfMan = people.filter(person => century ? person.sex
+    === 'm' && Math.ceil(person.died / 100) === century : person.sex === 'm');
 
+  return arrayOfMan.reduce((sum, person) =>
+    sum + (person.died - person.born), 0)
+  / arrayOfMan.length;
```
tetiana-ivanchenko, line 24: move it to the separate fuction, because you use it twice

tetiana-ivanchenko (changes requested):

src/averageAges.js
```diff
@@ -14,14 +14,20 @@
...
+  return totalAge / arrayOfPeople.length;
 }
 
+function calculateMenAverageAge(people, century) {
+  const arrayOfMan = people.filter(person => century ? person.sex
+    === 'm' && Math.ceil(person.died / 100) === century : person.sex === 'm');
+
+  return calculateAverageAge(arrayOfMan);
+}
 /**
```
OleziO, line 31: Delete all comments.

OleziO (approved):

PR #1465, opened 2022-11-24
Solution

UmizDemud:


src/averageAges.js
```diff
@@ -15,11 +15,33 @@
...
+    }, 0);
+  } else {
+    result = people.reduce((a, b) => {
+      if (b.sex === 'm') { // Sex is male
+        count++;
+
+        return a + b.died - b.born;
+      } else {
+        return a;
+      }
```
mgrinko, line 40: if is not a good idea to have nested ifs because readability is not so good
here you need only filter and map methods

src/averageAges.js
```diff
@@ -15,11 +15,33 @@
...
+            && (Math.ceil(b.died / 100) === century)) {
+        count++;
+
+        return a + b.died - b.born;
+      } else {
+        return a;
+      }
+    }, 0);
+  } else {
+    result = people.reduce((a, b) => {
```
mgrinko, line 33: please remove the duplication

mgrinko (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const selectedPeople = people
+    .filter(p => p.sex === 'm'
+      && (century ? Math.ceil(p.died / 100) === century : true));
+  const count = selectedPeople.length;
+
+  return selectedPeople
+    .reduce((a, b) => a + b.died - b.born, 0) / (count > 0 ? count : 1);
```
meforoma, line 24: introduce a helper function that will calculate average

src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const selectedPeople = people
+    .filter(p => p.sex === 'm'
+      && (century ? Math.ceil(p.died / 100) === century : true));
```
meforoma, line 20:
```suggestion
      && (century
        ? Math.ceil(p.died / 100) === century
        : true)
      );
```

meforoma (changes requested):
relevant logic - but you totally need helper functions - at least to calculate average

src/averageAges.js
```diff
@@ -55,7 +74,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const selectedPeople = onlyWithSon
+    ? people.filter(p => p.sex === 'm')
+    : people;
+
+  let count = 0;
+  let mother;
+
+  const total = selectedPeople.reduce((a, b) => {
```
naziamoff, line 84: let's think of a better name for this variable

naziamoff: and for the parameters as well

src/averageAges.js
```diff
@@ -15,11 +26,12 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const selectedPeople = people
+    .filter(p => p.sex === 'm'
+      && (!century || Math.ceil(p.died / 100) === century)
+    );
```
naziamoff, line 32: https://github.com/mate-academy/js_average-ages/pull/1465#discussion_r1032533046

naziamoff (changes requested):
Well done, just a little bit of polishing needed

src/averageAges.js
UmizDemud: This is because I wrote all these codes in a single night. Also been learning since then

naziamoff (approved):
🔥

PR #1468, opened 2022-11-26
average-ages

mykolahuryn:


src/averageAges.js
```diff
@@ -55,7 +68,16 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const sort = people.filter(person => onlyWithSon
```
polosanya, line 71:
```suggestion
  const mothers = people.filter(person => onlyWithSon
```
Why sort? Let's make it clear

polosanya (changes requested):
Almost perfect! Please fix your naming

loralevitska (approved):
good job!

PR #1474, opened 2022-11-28
solution

lefteris-L:


nurdanokcu (approved):
I like the helper functions.

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const getMales = (person) => {
+    return person.sex === 'm'
+      && (arguments.length > 1
```
naziamoff, line 20: what if some day we decide to add some extra parameter to this function and this will not be century? This will break the logic

naziamoff (approved):
Impressive 👍

src/averageAges.js
lefteris-L: `century !== undefined`

PR #1477, opened 2022-11-28
Solution

romanblyashuk94:


Nazar-Brunarskyi (approved):
looks complicated) but interesting

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(({ sex, died }) => {
+    return century
+      ? Math.ceil(died / 100) === century && sex === 'm'
+      : sex === 'm';
+  });
+
+  return men.reduce((sum, { born, died }) => sum + died - born, 0)
+    / men.length;
```
solaryasha, line 25: You could create the `average` function which will sum up the numbers. But it's not required

src/averageAges.js
```diff
@@ -55,7 +65,20 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(({ mother: mothersName, sex }) => {
+    return onlyWithSon
+      ? people.some(({ name }) => name === mothersName) && sex === 'm'
+      : people.some(({ name }) => name === mothersName);
```
solaryasha, line 71: Just, for your information, you could do this search without the nested loop

solaryasha (approved):
Good job!

PR #1480, opened 2022-11-30
Solution

KrisMakarovska:


src/averageAges.js
```diff
@@ -20,6 +20,17 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  const men = century
+    ? people.filter(person => ((Math.ceil(person.died / 100))
+      === century && person.sex === 'm'))
+    : people.filter(person => person.sex === 'm');
+
+  const averegeAges = men
```
polosanya, line 29:
```suggestion
  const averageAges = men
```

src/averageAges.js
```diff
@@ -55,7 +74,20 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = onlyWithSon
+    ? people.filter(child => people.some(mom => child.mother === mom.name
+      && child.sex === 'm'))
+    : people.filter(child => people.some(mom => child.mother === mom.name));
```
polosanya, line 80: Try not to duplicate yourself. In both cases, you have almost the same logic. 
Think about using the ternary operator inside the filter callback.
It will help you to improve performance and make your code cleaner
You can fix it in all functions

polosanya (changes requested):
Almost perfect! Please check the comments

src/averageAges.js
KrisMakarovska: +

src/averageAges.js
KrisMakarovska: +

polosanya (approved):
Well done!

PR #1482, opened 2022-11-30
solution

nikachu404:


src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => (
+    century === undefined
+      ? person.sex === 'm'
+      : person.sex === 'm' && Math.ceil(person.died / 100) === century
```
4ebupel, line 21: You could just erase === undefined and swap the expressions, also you are using person.sex === 'm' expresion twice, so it's better to create a variable for them

src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
+  ));
+  const calculateAge = men.map(person => person.died - person.born);
+  const calculateAverageAge = calculateAge.reduce((sum, age) => (
+    (sum + age))) / men.length;
```
4ebupel, line 25: It's better to make a separate function for that purpose in the end of a file, before exports

src/averageAges.js
```diff
@@ -37,7 +42,16 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => (
+    withChildren
+      ? people.find(child => child.mother === person.name)
+      : person.sex === 'f'
+  ));
+  const calculateAge = women.map(person => person.died - person.born);
+  const calculateAverageAge = calculateAge.reduce((sum, age) => (
+    (sum + age))) / women.length;
```
4ebupel, line 52: The same comment about function)

src/averageAges.js
```diff
@@ -55,7 +69,18 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const children = people.filter(person => (
+    onlyWithSon
+      ? people.find(mother => mother.name === person.mother)
+        && person.sex === 'm'
+      : people.find(mother => mother.name === person.mother)
+  ));
+  const calculateAge = children.map(person =>
+    (person.born - people.find(mother => mother.name === person.mother).born));
+  const calculateAverageAge = calculateAge.reduce((sum, age) => (
+    (sum + age))) / children.length;
```
4ebupel, line 81: I think u already understand what I want))

4ebupel (changes requested):
Good work, only few small issues)

src/averageAges.js
nikachu404: Thanks for the review! I'll fix it now

ivan-kadykalo (approved):
Good job 🔥

PR #1484, opened 2022-12-01
Solution

CharAnagn:


src/averageAges.js
```diff
@@ -56,8 +67,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    return child.born - mother.born;
+  });
+
+  return ageDifference.reduce((a, b) => a + b, 0) / ageDifference.length;
 }
 
+const calculateAverageHelper = (array) => {
+  const len = array.length;
+
+  return array.map(a => a.died - a.born).reduce((x, y) => x + y, 0) / len;
```
nurdanokcu, line 88: For better readability. 
 ```suggestion
   return array
     .map(person => a.died - a.born)
     .reduce((x, y) => x + y, 0) / len;
```
You can even change the 'a' with 'person'. But I think it is fine.

File: src/averageAges.js [hunk 2]
```diff
@@ -38,6 +44,11 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const filteredWomen = people.filter(person => withChildren
+    ? person.sex === 'f' && people.find(any => any.mother === person.name)
```
nurdanokcu, line 48: Here 
 _**some  method**_
 is also useful.  Because to determine if that person is a mother, if we find anybody that mentions her name, it is enough. We don't need to find who her child is. It works perfectly here ofc. Just wanted to mention a different way also. 
Same for determining if that person is a child of somebody.

nurdanokcu (approved):
Very nice and clear solution. 🚀

src/averageAges.js
```diff
@@ -20,6 +20,12 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  const filteredMen = people.filter(person => century
+    ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+    : person.sex === 'm');
+
+  return calculateAverageHelper(filteredMen);
```
meforoma, line 28: i'm surprised this even worked since you introduce the `calculateAverageHelper` function later with a const special word, but using it before declaration 🤔

src/averageAges.js
meforoma: totally agree with @nurdanokcu
also variable naming `any` is, you know 🙂 
```suggestion
    ? person.sex === 'f'
      && people.some(child => child.mother === person.name)
```

src/averageAges.js
meforoma, line 88: upvote @nurdanokcu's comment
this helper would not work with any kind of array, right?
So better make it more specific.
And multi-line.
Also i'd declare this method before using it (see above).
```suggestion
const calculateAverageHelper = (peopleArray) => {
  const peopleCount = peopleArray.length;

  return peopleArray
    .map(person => person.died - person.born)
    .reduce((sum, age) => sum + age, 0) / peopleCount;
```

meforoma (changes requested):
please address few comments

meforoma (approved):
👍

PR #1491, opened 2022-12-03
Develop

SalehBUD:


UmizDemud (approved):
Readable 👍

src/averageAges.js
```diff
@@ -1,5 +1,16 @@
...
 
+/**
+ * Helper Functions
+ */
+const filterByGender = (people, gender) => {
+  return people.filter((person) => person.sex === gender);
+};
+
+const calculateAverage = (array) => {
+  return array.reduce((a, b) => b.died - b.born + a, 0) / array.length;
```
meforoma, line 11: one-letter variables are not great
```suggestion
  return array.reduce((ageSum, {died, born}) => (
    died - born + ageSum
  ), 0) / array.length;
```

src/averageAges.js
```diff
@@ -37,7 +50,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = withChildren
+    ? filterByGender(people, 'f').filter((mother) =>
+      people.find((child) => child.mother === mother.name)
```
meforoma, line 55: you just need to evaluate, right - so receive a boolean output `.some` would do a faster job in this regard.
`.find` returns the entire object - but you don't need an object.

src/averageAges.js
```diff
@@ -55,7 +74,22 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = onlyWithSon
+    ? people.filter(
+      (person) =>
+        people.find((mother) => person.mother === mother.name)
```
meforoma, line 80: '.some'

src/averageAges.js
```diff
@@ -55,7 +74,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+          && person.sex === 'm'
+    )
+    : people.filter((person) =>
+      people.find((mother) => person.mother === mother.name)
```
meforoma, line 84: `.some`

meforoma (changes requested):
please see the comments

meforoma (approved):
🎉

PR #1496, opened 2022-12-04
first commit

haduigon:


src/averageAges.js
```diff
@@ -20,6 +20,29 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const totalAge = people.map(human => human.sex === 'm'
```
TarasHoliuk, line 23: Instead of calculating both cases and only at the end of the function returning the correct case, first of all, better to filter `people` based on the second argument (use ternary inside filter callback).
For example, in the current function, filter people by `sex` and if `century` is defined - by century (when the person died) as well.

src/averageAges.js
```diff
@@ -20,6 +20,29 @@ function calculateMenAverageAge(people, century) {
...
+  const filteredCentury = totalAgeCentury.filter(year => year > 0);
+
+  const avarageMaleAge = womenFiltered.reduce((a, b) => a + b, 0);
+
+  const avarageMaleAgeCentury = filteredCentury.reduce((a, b) => a + b, 0);
+
+  let final = 0;
+
+  century > 0 ? final
+    = avarageMaleAgeCentury / filteredCentury.length
```
TarasHoliuk, line 42: You are doing almost the same in each function (reduce and result of reduce / length). please, move it to a separate function

TarasHoliuk (changes requested):

src/averageAges.js
haduigon: done

src/averageAges.js
haduigon: done

src/averageAges.js
```diff
@@ -14,12 +14,29 @@
...
 function calculateMenAverageAge(people, century) {
   // write code here
   // learn how to use array methods like .filter .map .some .every .find .reduce
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const filteredPeople = people.filter(human =>
+    century !== undefined
+      ? human.sex === 'm' && century === Math.ceil(human.died / 100)
+      : human.sex === 'm');
```
tetiana-ivanchenko, line 31:
```suggestion
    century
      ? human.sex === 'm' && century === Math.ceil(human.died / 100)
      : human.sex === 'm');
```

src/averageAges.js
```diff
@@ -14,12 +14,29 @@
...
+
+  ;
+
+  const totalAgeCentury = filteredPeople.reduce(
+    (a, b) => a + (b.died - b.born), 0
+  );
```
tetiana-ivanchenko, line 37: you use it twice, so you can move it in the separate function

src/averageAges.js
```diff
@@ -36,8 +53,19 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const filteredPeople = people.filter(human => people.find(mother =>
+    withChildren === true
+      ? human.name === mother.mother
+      : human.sex === 'f'
```
tetiana-ivanchenko, line 62:
```suggestion
    withChildren
      ? human.name === mother.mother
      : human.sex === 'f'
```

src/averageAges.js
```diff
@@ -56,6 +84,28 @@ function calculateWomenAverageAge(people, withChildren) {
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  const womenMess = people.map(woman => woman.sex === 'f' ? woman : 0);
+
+  const women = womenMess.filter(woman => woman !== 0);
```
tetiana-ivanchenko, line 89: combine it, you dont need map, use just filter

src/averageAges.js
```diff
@@ -56,6 +84,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const children = people.filter(
+    child => women.find(mother => onlyWithSon === true
```
tetiana-ivanchenko, line 92:
```suggestion
    child => women.find(mother => onlyWithSon
```

tetiana-ivanchenko (changes requested):

src/averageAges.js
haduigon: it is done in new commit )

src/averageAges.js
haduigon: it is done in new commit )

src/averageAges.js
haduigon: it is done in new commit )

src/averageAges.js
haduigon: it is done in new commit )

src/averageAges.js
haduigon: it is done in new commit )

nazarmatsevych (approved):

PR #1497, opened 2022-12-04
Solution

oleh-yablunovskyi:


src/averageAges.js
```diff
@@ -15,13 +17,21 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(man => (
```
vadiimvooo, line 20: Maybe it will be better to rename this variable to something like filterByMen. The same situation for other functions.

vadiimvooo (approved):

src/averageAges.js
```diff
@@ -15,13 +17,21 @@
...
+  ));
+
+  const menAgesSum = menList.reduce((totalAge, man) => (
+    totalAge + (man.died - man.born)
+  ), 0);
+
+  return menAgesSum / menList.length;
 }
 
+// console.log(calculateMenAverageAge(people, 18));
```
solaryasha, line 33: Don't leave comments

src/averageAges.js
```diff
@@ -15,13 +17,21 @@
...
+    !century
+      ? man.sex === 'm'
+      : man.sex === 'm' && Math.ceil(man.died / 100) === century
+  ));
+
+  const menAgesSum = menList.reduce((totalAge, man) => (
+    totalAge + (man.died - man.born)
+  ), 0);
+
+  return menAgesSum / menList.length;
```
solaryasha, line 30: You are duplicating code three times, let's create a function

src/averageAges.js
```diff
@@ -15,13 +17,21 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menList = people.filter(man => (
+    !century
+      ? man.sex === 'm'
```
solaryasha, line 22: `man.sex === 'm'` - you are using this expression twice. So, let's create a variable for it.

solaryasha (changes requested):

PR #1498, opened 2022-12-04
solution v1

Lun4trick:


src/averageAges.js
```diff
@@ -1,61 +1,74 @@
...
  *
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothers = people.map(eachPerson => eachPerson.mother);
+  let womenOnly = people.filter(woman => woman.sex === 'f');
+
+  if (withChildren) { // checks if person is a mother
+    womenOnly = womenOnly.filter(woman => mothers.includes(woman.name));
```
meforoma, line 39: really thoughtful approach 👏

src/averageAges.js
```diff
@@ -1,61 +1,74 @@
...
+  }
+
+  function calcMotherChildDiff(obj) {
+    const mother = people.find(person => person.name === obj.mother);
+
+    return obj.born - mother.born;
+  }
+
+  const AgeDiff = children.map(person =>
+    calcMotherChildDiff(person));
```
meforoma, line 68: just a hint - this syntax would work just fine and shortens the code
```suggestion
  const AgeDiff = children.map(calcMotherChildDiff);
```

PR #1501, opened 2022-12-04
Solution

4ebupel:


oleksandrnikora (approved):
It looks good

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // without nesting
+  const men = people.filter(person => (
+    century
+      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+      : person.sex === 'm'
+  ));
+  const menAges = men.map(person => person.died - person.born);
+  const averageMenAge = getAverageValue(menAges, men.length);
+
+  return averageMenAge;
```
solaryasha, line 26:
```suggestion
  return getAverageValue(menAges, men.length);;
```

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => (
+    century
+      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+      : person.sex === 'm'
+  ));
+  const menAges = men.map(person => person.died - person.born);
+  const averageMenAge = getAverageValue(menAges, men.length);
```
solaryasha, line 24: Why do you need to pass length as an argument? You could calculate it in the function body

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => (
+    century
+      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+      : person.sex === 'm'
```
solaryasha, line 21: Maybe, let's create the variable for `person.sex === 'm'` expression, cause you are using that value twice.

src/averageAges.js
```diff
@@ -37,7 +41,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => (
+    withChildren
+      ? people.find(child => child.mother === person.name)
+      : person.sex === 'f'
```
solaryasha, line 47: Here as well, let's create the variable for `person.sex === 'f'` expression, cause you are using that value twice.

solaryasha (changes requested):

src/averageAges.js
4ebupel: +

src/averageAges.js
4ebupel: +

src/averageAges.js
4ebupel: didn't get it, maybe I'm some sort of blind, but I think that  person.sex === 'f' expression used only once in the whole task, isn't it?

solaryasha (approved):

PR #1508, opened 2022-12-05
solution

Alireza-Mohammadhossein:


src/averageAges.js
```diff
@@ -15,6 +15,15 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  const man = people.filter(person => person.sex === 'm');
+  const men = century
+    ? man.filter(
+      (person) => Math.ceil(person.died / 100) === century
+    )
+    : man;
```
Lun4trick, line 23:
```suggestion
```
I Think here its better to use a simple 'if' statement, what filters the original woman array when some parameter is valid, if we want to add more params in the future, its easier to do with this method. U can use the same logic further.

src/averageAges.js
```diff
@@ -37,7 +46,14 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const woman = people.filter(person => person.sex === 'f');
+  const women = withChildren
```
meforoma, line 50: same thing
```suggestion
  const women = people.filter(person => person.sex === 'f');
  const womenWithChildren = withChildren
```

src/averageAges.js
```diff
@@ -55,8 +71,35 @@ function calculateWomenAverageAge(people, withChildren) {
...
+        people.find((mother) => person.mother === mother.name)
+          && person.sex === 'm'
+    )
+    : people.filter((person) =>
+      people.find((mother) => person.mother === mother.name)
+    );
+
+  const different = children.map(
+    (child) =>
+      child.born - people.find((mother) => child.mother === mother.name).born
```
meforoma, line 86: apply de-structuring of arguments for practice sake
`(child)` -> `({ born, mother })`
`(mother)` -> `({ name })`

src/averageAges.js
```diff
@@ -55,8 +71,35 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  );
+
+  return different.reduce((acc, el) => acc + el) / different.length;
 }
+//
+//
+//
+//
+//
```
meforoma, line 95: remove unnecessary empty lines

PR #1524, opened 2022-12-05
solution

daryna-zinchenko:


src/averageAges.js
```diff
@@ -15,11 +15,12 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people
+    .filter(person => person.sex === 'm')
+    .filter(person => !century || century === Math.ceil(person.died / 100))
+    .map(person => person.died - person.born);
+
+  return men.reduce((a, b) => a + b) / men.length;
```
Anastasia4sik, line 23: Checklist p2, correct in whole work
```suggestion
  return men.reduce((a, b) => (a + b)) / men.length;
```

src/averageAges.js
```diff
@@ -37,7 +38,16 @@ function calculateMenAverageAge(people, century) {
...
+  const mothersNames = people
+    .filter(person => person.mother)
+    .map(person => '' + person.mother);
+
+  const women = people
+    .filter(person => person.sex === 'f')
+    .filter(person => !withChildren || mothersNames.includes(person.name))
+    .map(person => person.died - person.born);
+
+  return women.reduce((a, b) => a + b) / women.length;
```
Anastasia4sik, line 50: You`re using this reduce method everywhere, you can put it in a separate function and pass the necessary parameters, but it is optional

Anastasia4sik (changes requested):
Good job, but fix some points

PR #1536, opened 2022-12-06
Average ages

OlexandraKorol:


src/averageAges.js
```diff
@@ -37,7 +41,15 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenCheck = people.filter(person =>
+    withChildren
+      ? person.sex === 'f' && people.find(child => person.name === child.mother)
+      : person.sex === 'f'
```
OlggaJul, line 47: 'f' repeated twice
if that person is someone's mom, it's definitely a woman

OlggaJul (approved):

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const sexFilter = people.filter((person) => person.sex === 'm');
+
+  const centuryCheck = (century)
+    ? sexFilter.filter((man) => century === Math.ceil(man.died / 100))
+    : sexFilter;
```
solaryasha, line 22: Could we combine the conditions and use one filter?

src/averageAges.js
solaryasha: Good comment, let's fix it

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
+
+  const manAges = getAges(centuryCheck);
+
+  return manAges.reduce((sum, value) => sum + value) / manAges.length;
```
solaryasha, line 26: Let's create a function for calculating the average

solaryasha (changes requested):

solaryasha (approved):
👍

PR #1537, opened 2022-12-06
Solution

sercansimsek:


src/averageAges.js
```diff
@@ -55,7 +69,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  let getChildren = people.filter((child) => {
+    return people.some((person) => person.name === child.mother);
+  });
+
+  getChildren = onlyWithSon
+    ? getChildren.filter((child) => child.sex === 'm')
+    : getChildren;
+
+  const sumAge = getChildren.reduce((finalValue, child) => {
+    const mother = people.find(p => p.name === child.mother);
```
Iksich, line 81: It would be better maybe to use something more describing than p.

Iksich (approved):

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const getMen = people.filter((person) => person.sex === 'm');
```
meforoma, line 18: the name is not great - `getMen` would fit a helper function that for expample processes array `people`
but this is just an array of `men`
```suggestion
  const men = people.filter((person) => person.sex === 'm');
```

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
+  const currentMan = century
+    ? getMen.filter((person) => Math.ceil(person.died / 100) === century)
```
meforoma, line 20:
```suggestion
    ? men.filter(({ died }) => Math.ceil(died / 100) === century)
```

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const getMen = people.filter((person) => person.sex === 'm');
+  const currentMan = century
```
meforoma, line 19: but it's not a single `man` right?

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
+    ? getMen.filter((person) => Math.ceil(person.died / 100) === century)
+    : getMen;
+
+  return currentMan
+    .map((person) => person.died - person.born)
+    .reduce((a, b) => a + b) / currentMan.length;
```
meforoma, line 25: introduce a helper function that will calculate the average age from a passed array of people

meforoma (changes requested):
please see the comments

src/averageAges.js
```diff
@@ -14,12 +21,15 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  return people
+    .filter(person => (person.sex === 'm')
+      && (century === undefined || century === Math.ceil(person.died / 100))
```
naziamoff, line 28: in this case, you only check whether the century is explicitly undefined, but what if we have `null` there?
So it means that it is better to check not for undefined value, but to check whether it is `falsy`
```suggestion
      && (!century || century === Math.ceil(person.died / 100))
```

src/averageAges.js
```diff
@@ -1,5 +1,12 @@
 'use strict';
 
+function hasChildren(people, woman) {
+  return people.some((person) => person.mother === woman.name);
+}
+
+function getMother(people, child) {
+  return people.find(mother => mother.name === child.mother);
+}
```
naziamoff, line 9: better to make those arrow ones. There's a difference in the hoisting process. Take a look at [the article](https://www.educative.io/answers/function-declaration-vs-function-expression-in-javascript)

naziamoff: not critical though

src/averageAges.js
```diff
@@ -37,7 +47,12 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  return people
+    .filter(person => (person.sex === 'f'))
+    .filter((person) => withChildren ? hasChildren(people, person) : true)
+    .reduce((average, person, index, men) => {
```
naziamoff, line 53: why are you accepting `men` as parameter?

naziamoff (approved):
Looks good. I have a bunch of comments though

src/averageAges.js
sercansimsek: Because I need to get the array length here to calculate average but it would be better to name it women here :D

PR #1541, opened 2022-12-06
Solution

Iksich:


Edriss572 (commented):
nice done

src/averageAges.js
```diff
@@ -14,12 +14,30 @@
...
  */
+const male = person => person.sex === 'm';
+const female = person => person.sex === 'f';
+
+const calculateAverage = (ages) => {
+  const sumOfAge = ages
+    .map(person => person.died - person.born)
+    .reduce((sum, age) => sum + age, 0) / ages.length;
+
+  return sumOfAge;
```
meforoma, line 25: name is misleading - it's not a sum, but an actual average, right?
```suggestion
  const averageAge = ages
    .map(person => person.died - person.born)
    .reduce((sum, age) => sum + age, 0) / ages.length;

  return averageAge;
```
or 
```suggestion
  const sumOfAge = ages
    .map(person => person.died - person.born)
    .reduce((sum, age) => sum + age, 0);

  return sumOfAge / ages.length;
```

src/averageAges.js
```diff
@@ -56,6 +81,19 @@ function calculateWomenAverageAge(people, withChildren) {
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  const kids = people.filter(kid =>
+    onlyWithSon
+      ? people.some(person => kid.mother === person.name) && male(kid)
+      : people.some(person => kid.mother === person.name)
+  );
```
meforoma, line 88: could be bit shorter
```suggestion
  const kids = people.filter(kid =>
    people.some(person => kid.mother === person.name)
    && (onlyWithSon ? male(kid) : true)
  );
```

meforoma (approved):
relevant logic - just pay attention to few comments i've wrote you

PR #1544, opened 2022-12-06
Done

JzSol:


src/averageAges.js
```diff
@@ -55,7 +68,31 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const getPeople = onlyWithSon
+    ? people.filter((person) => person.sex === 'm')
+    : people;
```
Grayfox147, line 73:
```suggestion
  const getPeople = onlyWithSon
    ? people.filter((person) => person.sex === 'm')
    : people.filter(child =>
      people.find(person => person.name === child.mother));
```
You could easily filter inmeadiatly for your the element you are looking to work with we can call it child and then you find the child by looking that person that has a mother, by comparing each person name, with every element mother . and you could save some coding afterwards.

src/averageAges.js
```diff
@@ -55,7 +68,31 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+    if (mother) {
+      len++;
+
+      return ttl + person.born - mother.born;
+    }
+
+    return ttl;
+  }, 0);
+
```
Grayfox147, line 88:
```suggestion
 const ageDif = children.map(child =>
    child.born - people.find(person => person.name === child.mother).born
  );
```

You could implement map to replace this statement and inside inmeadiately compare your element child .born  - its mother =  people.find(person => person.name === child.mother) . born

Grayfox147 (commented):
Overall really good implementation of the array methods

src/averageAges.js
```diff
@@ -37,10 +39,21 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const getPeople = people.filter(
```
meforoma, line 42: semantically these are `women`, right
`getPeople` is a good name for a function that would somehow process people array
```suggestion
  const women = people.filter(
```

src/averageAges.js
meforoma: upvote

meforoma (changes requested):
please see the comments

meforoma (approved):

PR #1545, opened 2022-12-06
finished

jackelashkar:


alaaashkar (approved):
amazing !

src/averageAges.js
```diff
@@ -55,7 +72,18 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(person => (
+    onlyWithSon
+      ? people.find(mother => mother.name === person.mother)
+        && person.sex === 'm'
+      : people.find(mother => mother.name === person.mother)
```
meforoma, line 79: people.find twice - i think you could work it around like this
```suggestion
    people.find(mother => mother.name === person.mother)
    && (onlyWithSon
      ? person.sex === 'm'
      : true)
```

meforoma (changes requested):
please pair up with @alaaashkar - same comments apply )

src/averageAges.js
jackelashkar: you're right Roman

meforoma (approved):
🎉

PR #1547, opened 2022-12-06
done

KareemAchkar:


jackelashkar (approved):
well done!

meforoma (changes requested):
@KareemAchkar you're pairing up with Brothers, right?
please apply the same comments 😉

src/averageAges.js
```diff
@@ -14,14 +13,20 @@
...
+    const isMale = person.sex === 'm';
+
+    return century
+      ? isMale && Math.ceil(person.died / 100) === century
+      : isMale;
+  });
 
+  const calculateAge = men.map(person => person.died - person.born);
+
+  return calculateAverage(calculateAge);
```
meforoma, line 28: `calculateAverage` is a proper name for a function that do calculate it.
But `calculateAge` is not a function, but an array of ages instead.
It might seem minor, but when the codebase grows big, reasonable names make difference.

```suggestion
  const menAges = men.map(person => person.died - person.born);

  return calculateAverage(menAges);
```

src/averageAges.js
```diff
@@ -36,10 +41,18 @@ function calculateMenAverageAge(people, century) {
...
-}
+  const women = people.filter(person => (
+    withChildren
+      ? people.find(child => child.mother === person.name)
+      : person.sex === 'f'
+  ));
+
+  const calculateAge = women.map(person => person.died - person.born);
 
+  return calculateAverage(calculateAge);
```
meforoma, line 54:
```suggestion
  const womenAges = women.map(person => person.died - person.born);

  return calculateAverage(womenAges);
```

src/averageAges.js
```diff
@@ -54,10 +67,24 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    onlyWithSon
+      ? people.find(mother => mother.name === person.mother)
+      && person.sex === 'm'
+      : people.find(mother => mother.name === person.mother)
+  ));
+
+  const calculateAge = children.map(person =>
+    (person.born - people.find(mother => mother.name === person.mother).born));
+
+  return calculateAverage(calculateAge);
```
meforoma, line 82:
```suggestion
  const ageDiffs = children.map(person =>
    (person.born - people.find(mother => mother.name === person.mother).born));

  return calculateAverage(ageDiffs);
```

meforoma (approved):
✅ approve - but please do consider the comments i've left for the future.

PR #1549, opened 2022-12-07
Develop

TheOneThatKnowsNothing:


src/averageAges.js
```diff
@@ -15,11 +15,14 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  return century === undefined || century === false
```
Oleksandr-Filo, line 18: You could write it shorter
```suggestion
  return !century
```

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
+    ? people.filter(person => person.sex === 'm')
+      .map(person => person.died - person.born)
+      .reduce((a, b, i) => +(a + (b - a) / (i + 1)).toFixed(3), 0)
+    : people.filter(person =>
+      Math.ceil(person.died / 100) === century && person.sex === 'm')
+      .map(person => person.died - person.born)
+      .reduce((a, b, i) => +(a + (b - a) / (i + 1)).toFixed(3), 0);
```
Oleksandr-Filo, line 25: 1. DRY
You repeat the same calculations six times, probably you should create function for making these calculations? 
2. You should use more understandable names for variables (not a, b etc.)

Oleksandr-Filo (changes requested):

ivan-kadykalo (approved):
Good job 🔥

PR #1554, opened 2022-12-07
Develop chupiak

Chupiak:


Chupiak:
Conditions of method 'calculateAverageAgeDiff' not enough. Some mothers can has 2 or more children. One mother can be the daughter of another mother at the same time, and so on. This makes it difficult to find the right solution.

Chupiak:
Some part of the code can be moved to separate methods, but will it be correct in this case?

Chupiak:
I solved this task, but solution is right only for 'people.js' case.

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  return (century)
+    ? (people
+      .filter((x) => (Math.ceil(x.died / 100) === century && x.sex === 'm'))
```
qoa11a, line 20: Let's move filtered people to a variable. Than you can use it everywhere you need to. DRY - don't repeat yourself.

src/averageAges.js
```diff
@@ -37,7 +45,27 @@ function calculateMenAverageAge(people, century) {
...
+      .reduce((x, y) => (x + y)) / arrayWithChildren.length;
+  }
+
+  return (people
+    .filter((x) => x.sex === 'f')
+    .map((x) => (x.died - x.born))
+    .reduce((x, y) => (x + y)))
+      / (people
+        .filter((x) => x.sex === 'f')
+        .map((x) => (x.died - x.born)).length);
```
qoa11a, line 68: This logic is hard to understand. Try to find out a simpler solution. KISS - keep it simple.

src/averageAges.js
```diff
@@ -55,7 +83,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    if (mother) {
+      len++;
+
+      return obj + person.born - mother.born;
+    }
+
+    return obj;
+  }, 0);
+
+  return sum / len;
```
qoa11a, line 105: Again, it's very hard to read. Create variables, move some logic to separate functions etc.

qoa11a (changes requested):
Good job! 👍 
Just few comments on code style. Don't hesitate to move some logic to separate functions and create variables for values that you calculate more than one time.

DarkMistyRoom (approved):

PR #1557, opened 2022-12-07
First commit by Edriss Miakhel

Edriss572:


src/averageAges.js
```diff
@@ -56,6 +73,18 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    .filter(child => getMother(people, child))
+    .reduce((average, child, index, children) => {
+      const mother = getMother(people, child);
+
+      return average + (child.born - mother.born) / children.length;
+    }, 0);
+}
+
+function getMother(people, child) {
+  return people.find(mother => mother.name === child.mother);
```
meforoma, line 87: you don't need an actual object - but just confirm it exists - use `.some` method

src/averageAges.js
```diff
@@ -56,6 +73,18 @@ function calculateWomenAverageAge(people, withChildren) {
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  return people
+    .filter(child => onlyWithSon ? child.sex === 'm' : true)
```
meforoma, line 77: putting too much inside `return` - better keep it more explanatory
```
let mothers = people.filter(child => getMother(people, child));

if (onlyWithSom) {
  mothers = mothers.filter(child => child.sex === 'm')
}

return getAverageAge(mothers)
```

getAverageAge(mothers) - it's another comment - create a helper function to calculate avg ages from the array passed

src/averageAges.js
```diff
@@ -38,6 +49,12 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  return people
+    .filter(person => person.sex === 'f')
+    .filter(person => withChildren ? hasChildren(people, person) : true)
+    .reduce((average, person, index, men) => {
+      return average + (person.died - person.born) / men.length;
```
meforoma, line 56: see the similar comment below

meforoma (changes requested):
please see the comments

PR #1559, opened 2022-12-07
Solution

NgineZZ:


src/averageAges.js
```diff
@@ -1,61 +1,50 @@
...
-  // write code here
+  let children = people.filter(child => {
+    return people.some(person => person.name === child.mother);
+  });
+
+  if (onlyWithSon) {
+    children = children.filter(child => child.sex === 'm');
+  }
+
+  const ageDiff = children.reduce((sum, child) => {
```
nurdanokcu, line 41: Nice usage of reduce .

nurdanokcu (approved):
Neat.

src/averageAges.js
```diff
@@ -1,61 +1,50 @@
...
- * @param {number} century - optional
- *
- * @return {number}
- */
+function calculateAverageAge(persons) {
+  const ages = persons.map(person => person.died - person.born);
+  const averageAge = ages.reduce((sum, age) =>
+    (sum + age)) / ages.length;
+
+  return averageAge;
```
naziamoff, line 8: this way we shorten your code without losing any readability

and if you move the callback to a new line - don't forget to use round braces
```suggestion
  return ages.reduce((sum, age) => (
    sum + age
  )) / ages.length;
```

src/averageAges.js
```diff
@@ -1,61 +1,50 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let men = people.filter(person => person.sex === 'm');
+
+  if (century) {
+    men = men.filter(man => (Math.ceil(man.died / 100) === century));
+  }
```
naziamoff, line 16: it is possible to calculate it inside one filter callback, let's do it

src/averageAges.js
```diff
@@ -1,61 +1,50 @@
...
- *
- * @return {number}
- */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothers = people.map(person => person.mother);
+  let women = people.filter(person => person.sex === 'f');
+
+  if (withChildren) {
+    women = women.filter(woman => mothers.includes(woman.name));
```
naziamoff, line 26: same here, use ternary

naziamoff (changes requested):
Already looks good, although a little bit of polishing required

PR #1565, opened 2022-12-10
 js_average-ages

rafmamedov:


src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const malePerson = people.filter(person => {
+    return century
+      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+      : person.sex === 'm';
```
ivanchuk-yurii, line 21:
```suggestion
    return person.sex === 'm' && (
      !century
      || Math.ceil(person.died / 100) === century
    );
```
imho it's better. change also in other functions

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
+  });
+
+  const menAges = malePerson.map(person => person.died - person.born);
+  const sumOfAges = menAges.reduce((sum, age) => sum + age, 0);
+  const averageAge = sumOfAges / menAges.length;
```
ivanchuk-yurii, line 26: you also have this in `calculateWomenAverageAge`, move this into separate function

src/averageAges.js
rafmamedov: +

rafmamedov: got it, thanks!

SanyaBratashchuk (approved):

PR #1567, opened 2022-12-11
Solution

skomskiy:


src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
+      return person.sex === 'm' && Math.ceil(person.died / 100) === century;
+    }
+
+    return person.sex === 'm';
+  });
+
+  const ages = men.reduce((sum, age) => sum + (age.died - age.born), 0);
+  const agesMiddle = ages / men.length;
+
+  return (agesMiddle);
```
SanyaBratashchuk, line 29: Move this logic to the separate function and reuse it

src/averageAges.js
```diff
@@ -55,7 +76,20 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  let children = people.filter(
+    person => people.find(mother => mother.name === person.mother));
+
+  if (onlyWithSon) {
+    children = children.filter(child => child.sex === 'm');
+  }
```
SanyaBratashchuk, line 84: image

SanyaBratashchuk: Fix it everywhere

SanyaBratashchuk (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
+      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+      : person.sex === 'm'
+  );
+
+  function functionAge() {
+    const ages = men.reduce((sum, age) => sum + (age.died - age.born), 0);
+    const agesMiddle = ages / men.length;
+
+    return agesMiddle;
+  }
```
SanyaBratashchuk, line 29: move it out of this function so other functions have accsess to it. And avoid duplication

SanyaBratashchuk (approved):

PR #1578, opened 2022-12-17
Solution

Ukrainiane-panda:


src/averageAges.js
```diff
@@ -55,7 +62,16 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const identifyChildren = people.filter(child =>
+    people.find(person => person.name === child.mother)
+    && (onlyWithSon
+      ? child.sex === 'm' : true));
+
+  const mothers = identifyChildren.map((child) =>
+    people.find(mother => child.mother === mother.name));
+
+  return identifyChildren.reduce((sum, child, index) =>
+    sum + child.born - mothers[index].born, 0) / identifyChildren.length;
```
OlEzhA0, line 74: it's interesting idea to use `mothers[index]` to reduce your algorithm complexity 
but you can prettify your solution, and do something like 

(I wrote it's just in this textarea but it should work well, if you'll have any issues - ping me in chat)
```suggestion
  const childrenWithMotherParams = identifyChildren.map((child) => ({ 
    ...child, 
    motherBorn: people.find(mother => child.mother === mother.name)?.born || 0
  }));

  return childrenWithMotherParams.reduce((sum, child) =>
    sum + child.born - child.motherBorn, 0) / identifyChildren.length;
```

it's just for your understanding because `[index]` usually is not obvious for a quick code reading

OlEzhA0 (approved):
GJ

PR #1585, opened 2022-12-23
Solution

Vadivel-Babu:


src/averageAges.js
```diff
@@ -20,6 +20,17 @@ function calculateMenAverageAge(people, century) {
...
   // without nesting
+  const men = people.filter(
+    (person) =>
+      person.sex === 'm'
+      && (century === Math.ceil(person.died / 100) || !century)
+  );
+
+  const averageAge
+    = men.map((man) => man.died - man.born).reduce((a, b) => a + b, 0)
+    / men.length;
```
qoa11a, line 31: Create a function to calculate averageAge as you use it more than one time.

qoa11a (changes requested):

src/averageAges.js
```diff
@@ -14,12 +14,30 @@
  *
  * @return {number}
  */
+
+function callBack(arr, len) {
```
polosanya, line 18:
```suggestion
function calculateAverageAge(ages) {
```
Let's use more descriptive and meaningful names. 
Also, why do you need to pass the array length? You can get it inside the function

src/averageAges.js
```diff
@@ -38,6 +56,19 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const women = people
+    .filter(
+      (person) =>
+        person.sex === 'f'
+        && (withChildren
```
polosanya, line 63:
```suggestion
        person.sex === 'f'
          && (withChildren
```
Fix all cases

src/averageAges.js
```diff
@@ -38,6 +56,19 @@ function calculateMenAverageAge(people, century) {
...
+          ? people.find((child) => child.mother === person.name)
+          : !withChildren)
+    )
+    .map((woman) => woman.died - woman.born);
+
+  const result = callBack(women, women.length);
```
polosanya, line 69: Result is not the best name for variable. Try to specify what you get from this function call.
Fix all cases

polosanya (changes requested):
Almost perfect!

src/averageAges.js
polosanya: Not fixed

src/averageAges.js
```diff
@@ -56,6 +87,21 @@ function calculateWomenAverageAge(people, withChildren) {
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  const children = people
+    .filter(
+      (person) =>
+        people.some((mom) => mom.name === person.mother)
+        && (onlyWithSon ? person.sex === 'm' : !onlyWithSon)
```
polosanya, line 94:
```suggestion
        people.some((mom) => mom.name === person.mother)
          && (onlyWithSon ? person.sex === 'm' : !onlyWithSon)
```

polosanya (approved):
Good job!

PR #1590, opened 2022-12-26
Solution

Malva37:


NVyrsta (changes requested):
We often use reduce method to calculate the average age in this task, and map method to make an array with ages.
It may be better to create a separate function for these goals.

Ex.:
function countAverage(numbers) { return numbers.reduce((a, b) => a + b, 0) / numbers.length; }
function getAges(people) { return people.map(person => person.died - person.born); }

Malva37:
> nice advice, thanks

src/averageAges.js
```diff
@@ -14,12 +14,24 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const mans = people.filter(man => {
+    return man.sex === 'm'
+    && (century === undefined ? true : Math.ceil(man.died / 100) === century);
+  });
```
alexandra-protyanova, line 30:
```suggestion
  const mans = people.filter(man => {
    return man.sex === 'm'
    && (!century  ? true : Math.ceil(man.died / 100) === century);
  });
```

src/averageAges.js
```diff
@@ -37,7 +49,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(woman => {
+    return woman.sex === 'f'
+    && (withChildren === undefined
+      ? true
+      : people.some(person => person.mother === woman.name));
+  });
```
alexandra-protyanova, line 57:
```suggestion
  const women = people.filter(woman => {
    return woman.sex === 'f'
    && (!withChildren
      ? true
      : people.some(person => person.mother === woman.name));
  });
```

src/averageAges.js
```diff
@@ -54,8 +75,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const mothers = people.filter(woman => {
+    return woman.sex === 'f'
+    && people.some(person => person.mother === woman.name);
+  });
+  const children = people.filter(person => {
+    return mothers.some(mother => person.mother === mother.name)
+    && (onlyWithSon === undefined
+      ? true
+      : person.sex === 'm');
+  });
```
alexandra-protyanova, line 90:
```suggestion
    && (!onlyWithSon
      ? true
      : person.sex === 'm');
  });
```

src/averageAges.js
```diff
@@ -54,8 +75,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  children.forEach(child => {
+    const foundMother = mothers.find(mom => mom.name === child.mother);
```
alexandra-protyanova, line 93: why do you need mothers if you are anyway finding in it, code will be simpler if you remove it and just use `people`

alexandra-protyanova (changes requested):

src/averageAges.js
Malva37: That’s

Malva37: I compere length og people and length of mothers(39 vs 14). So, when I use forEach to array "mothers" i do less itterations

alexandra-protyanova: Btw, when you were looking for moms, you already made one extra circe in second degree for it)

alexandra-protyanova (approved):

PR #1593, opened 2022-12-28
Solution

olena-hapon:


src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
+
+  arrayOfMen = century
+    ? arrayOfMen.filter(men => Math.ceil(men.died / 100) === century)
+    : arrayOfMen;
+
+  const ageOfMan = arrayOfMen.map(person => person.died - person.born);
+
+  const sumOfAge = ageOfMan.reduce((sum, age) => sum + age, 0);
+
+  const averageAgeOfMen = sumOfAge / arrayOfMen.length;
```
tetiana-ivanchenko, line 28: create separate function that takes array of people, and return average age, and use it here, and in the next function

tetiana-ivanchenko (changes requested):

Oleksandr-Ivanchenko (approved):

tetiana-ivanchenko (approved):

PR #1595, opened 2022-12-28
Solution

constantweb:


src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
+    .filter(person =>
+      person.sex === 'm'
+      && (century === Math.ceil(person.died / 100) || !century)
+    );
+
+  const menAge = men
+    .map(age => age.died - age.born);
+
+  const medianAge = menAge
+    .reduce((prev, current) => prev + current, 0) / menAge.length;
```
nazarmatsevych, line 28: move it to separate function as you use similar logic few times

nazarmatsevych (changes requested):

src/averageAges.js
```diff
@@ -55,7 +80,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    .map(child => people
+      .find(mother => mother.name === child.mother));
+
+  const ageDifference = children
+    .map((child, index) =>
+      child.born - mothers[index].born);
+
+  const result = calculateMedianAge(ageDifference);
+
+  return result;
```
TarasHoliuk, line 98:
```suggestion
  return calculateMedianAge(ageDifference);
```
FYI

TarasHoliuk (approved):
GJ! Left one comment FYI

PR #1597, opened 2022-12-30
average ages completed

MykolaKaradzha:


src/averageAges.js
```diff
@@ -1,62 +1,77 @@
...
  * @return {number}
  */
+
+const calculateAge = (person) => {
+  return person.died - person.born;
+};
+
+const filterBySex = (people, sex) => {
+  return people.filter(person => person.sex === sex);
+};
```
Viktor-Kost, line 16: No needs to create a separated function for an array method filter

src/averageAges.js
```diff
@@ -1,62 +1,77 @@
...
-}
+  const women = filterBySex(people, 'f');
+  const children = people.filter(person => people
+    .find(human => person.mother === human.name));
+  const childrenFiltered = onlyWithSon
+    ? filterBySex(children, 'm')
+    : children;
+  const ageDiffs = childrenFiltered.map(person => person.born - women
+    .find(woman => person.mother === woman.name).born);
+  const ageDiffsSum = ageDiffs.reduce((acc, age) => acc + age);
```
Viktor-Kost, line 71: here you an use methods chaining
Like
```
const ageDifference = array.map(....).reduce(...)
```

Viktor-Kost (changes requested):

src/averageAges.js
MykolaKaradzha: Thanks, got it.

tetiana-ivanchenko (approved):

PR #1598, opened 2022-12-30
solution

nzrii-kysel:


ArturErmolenko (approved):
I like this readable version more than previous :) 👍

PR #1601, opened 2023-01-02
Solution

Avichez:


Malva37 (approved):
good job)

src/averageAges.js
```diff
@@ -37,7 +45,20 @@ function calculateMenAverageAge(people, century) {
...
+      || people.some(person => person.mother === name);
+
+    return sex === 'f' && hasChildren;
+  });
+
+  const sumAgesWomen = onlyWomen.reduce((sum, woman) => {
+    const womanAge = woman.died - woman.born;
+
+    return sum + womanAge;
+  }, 0);
```
Anna-Tkachenko, line 59: the functions for calculating ages sum for men and women are the same, let's create universal function that will accept the array of people and will return the sum of their ages

Anna-Tkachenko (changes requested):
Well Done! ⭐️
Clear and understandable code, good job
Just small improvements can be implemented

src/averageAges.js
Avichez: Roger that, thanks for the hint =)

Anna-Tkachenko (approved):
⭐️

PR #1602, opened 2023-01-02
add solution

Anastasiia-Khudych:


src/averageAges.js
```diff
@@ -15,11 +15,22 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredPeople = people.filter(person => {
```
sv-dubov, line 18: Better use the variable name related with word "men". Because you are using the same name for women.

src/averageAges.js
```diff
@@ -37,7 +48,23 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filteredPeople = people
```
sv-dubov, line 51: Better use the variable name related with word "women".

src/averageAges.js
```diff
@@ -55,7 +82,27 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const filteredPeople = people
```
sv-dubov, line 85: Also needs other naming

src/averageAges.js
```diff
@@ -15,11 +15,22 @@
...
+      : person.sex === 'm';
+  });
+
+  const peopleAges = filteredPeople.map((person) => {
+    return person.died - person.born;
+  });
+
+  const averageAge = peopleAges.reduce((prevPerson, nextPerson) => {
+    return prevPerson + nextPerson;
+  }, 0) / peopleAges.length;
```
sv-dubov, line 31: You are using the similar algorithm for men and women. So you can make one function for both options.

sv-dubov (approved):
Good job! But check my comments

src/averageAges.js
Anastasiia-Khudych: Good idea, thanks!

src/averageAges.js
Anastasiia-Khudych: I can't entirely agree with this because the filter function receives all values of people object, not only men's or women's objects and only then it returns men's or women's objects.

ArturErmolenko (approved):

PR #1606, opened 2023-01-03
Solution

andrieieva-yuliia:


sv-dubov (approved):
Looks good!

src/averageAges.js
```diff
@@ -20,6 +20,17 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const filteredMan = people.filter(person =>
```
Anna-Tkachenko, line 23:
```suggestion
  const filteredMen = people.filter(person =>
```

src/averageAges.js
```diff
@@ -20,6 +20,17 @@ function calculateMenAverageAge(people, century) {
...
+    person.sex === 'm'
+    && (!century || century === Math.ceil(person.died / 100)));
+
+  const avgAge = filteredMan.reduce((total, man) => {
+    const manAge = man.died - man.born;
+
+    return total + manAge;
+  }, 0) / filteredMan.length;
```
Anna-Tkachenko, line 31: the process of calculating average age for different people arrays is the same
let's create separate function that will accept people array and will return the average age value

src/averageAges.js
```diff
@@ -37,7 +48,17 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filteredWomen = people.filter(person =>
+    person.sex === 'f'
+    && (!withChildren || people.some(man => man.mother === person.name)));
+
+  const avgAge = filteredWomen.reduce((total, women) => {
```
Anna-Tkachenko, line 55: check variable name, don't make it shortened, use full naming for better understanding

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! 🔥

PR #1612, opened 2023-01-04
Solution

oleh-holovnykh:


src/averageAges.js
```diff
@@ -15,11 +17,26 @@
...
+    ))
+    : menToCount = men;
+
+  totalAge = menToCount.reduce((sum, person) => {
+    const personAge = person.died - person.born;
+
+    return sum + personAge;
+  }, 0);
+
+  avarageAge = totalAge / menToCount.length;
```
victorionov, line 37: Its not necessary to create variable in the beginning of the function only for assigning and returning.
you can just 
`return totalAge / menToCount.lengh`

victorionov (approved):
Good Job!

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1613, opened 2023-01-04
average ages were calculated

StetsenkoViktoriia:


src/averageAges.js
```diff
@@ -14,12 +14,20 @@
...
+  const men = people.filter(person => person.sex === 'm');
+
+  const filteredPeople = century ? menWithCentury : men;
+
+  const ages = filteredPeople.map(person => (person.died - person.born));
+
+  const sumOfAge = ages.reduce((sum, elem) => sum + elem);
+  const result = sumOfAge / ages.length;
+
+  return result;
```
SanyaBratashchuk, line 30: Move this logic to the separate function and reuse it

src/averageAges.js
```diff
@@ -14,12 +14,20 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menWithCentury = people.filter(person => person.sex === 'm'
+    && Math.ceil(person.died / 100) === century);
+  const men = people.filter(person => person.sex === 'm');
```
SanyaBratashchuk, line 21: There is no need to iterate through array twice

SanyaBratashchuk (changes requested):

loralevitska (approved):
good job!

PR #1614, opened 2023-01-04
Solution

tania-boiar:


src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+  const menDiedInCentury = century
+    ? men.filter(man => Math.ceil(man.died / 100) === century)
+    : men;
+
+  const menAges = menDiedInCentury.map(man => man.died - man.born);
```
Avichez, line 23: Here we can optimize code, by using only one reduce method. And as this code repeating in second function, we can create separate function outside and put this logic inside of that function.
```suggestion
  const agesSum = menDiedInCentury.reduce((sum, person) => {
    const personAge = person.died - person.born;
  
    return sum + personAge;
  }, 0);
```

Avichez (approved):
Nice solutions 😊
Just one comment.

src/averageAges.js
Anna-Tkachenko: agree, check @Avichez comment
your code is clean and understandable, but you repeat the same logic for several arrays
let's create separate function that will accept people array and will return corresponding value with sum of ages or average age value

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #1619, opened 2023-01-05
Implemented calculateMenAverageAge function

GarryRocksman:
((have a nice evening!

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -15,11 +15,23 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menFormPeople
+    = people
+      .filter(person => person.sex === 'm');
+
+  const menFormPeopleByCentury
+    = menFormPeople
+      .filter(man => Math.ceil(man.died / 100) === century);
+
+  const menAges
+    = (century === undefined ? menFormPeople : menFormPeopleByCentury)
+      .map(man => man.died - man.born);
+
+  const averageAge
+    = menAges.reduce((previousAge, currentAge) => previousAge + currentAge, 0)
+    / menAges.length;
+
+  return averageAge;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


svitjojo (changes requested):
Good job !!!
Also mentors usually ask to create function for counting average age and use it to do not write a lot of repeatable code. Better make it)

src/averageAges.js
```diff
@@ -14,12 +14,26 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menFormPeople
+    = people
+      .filter(person => person.sex === 'm');
```
Denchepornyuk, line 26: not the best formating for the code, fix it in all places
```suggestion
  const menFormPeople = people.filter(person => (
    person.sex === 'm'
  ));
```

src/averageAges.js
```diff
@@ -37,7 +51,19 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenFormPeople
+    = people.filter(person => person.sex === 'f');
+
+  const womenWithChildren
+    = womenFormPeople
+      .filter(woman => people
+        .some((person) => person.mother === woman.name));
```
Denchepornyuk, line 60: can be done in one filter, just a filter with two conditions inside

src/averageAges.js
```diff
@@ -55,7 +81,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const peopleWithMother
+    = people.filter(person => people
+      .some(findMom => person.mother === findMom.name));
+
+  const manWithMother
+    = peopleWithMother.filter(person => person.sex === 'm');
+
```
Denchepornyuk, line 90: the same, can be done in one filter

Denchepornyuk (changes requested):

src/averageAges.js
GarryRocksman: thanks, fixed

src/averageAges.js
GarryRocksman: thanks fixed

src/averageAges.js
GarryRocksman: thanks fixed

Denchepornyuk (approved):

PR #1624, opened 2023-01-06
solution

imariash86:


anna-inozemets (approved):
Nice job)

src/averageAges.js
```diff
@@ -20,6 +20,20 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const allMen = people.filter(person => person.sex === 'm');
+  const diedInCentury = allMen
+    .filter(person => Math.ceil(person.died / 100) === century);
+
+  // const ageOfEvery = allMen.map(person => person.died - person.born);
```
alexandra-protyanova, line 27: remove commented code

src/averageAges.js
```diff
@@ -20,6 +20,20 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const allMen = people.filter(person => person.sex === 'm');
```
alexandra-protyanova, line 23: U can use ternary here and if `century` was passed just do the correct filter condition, to prevent repetitive logic in return

src/averageAges.js
```diff
@@ -56,6 +81,21 @@ function calculateWomenAverageAge(people, withChildren) {
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  const allChildren = people
+    .filter(child => people
+      .some(person => person.name === child.mother));
+  const hasSon = allChildren
+    .filter(person => person.sex === 'm');
+
+  return onlyWithSon
```
alexandra-protyanova, line 90: You can uses first comment to each function

alexandra-protyanova (changes requested):

src/averageAges.js
```diff
@@ -38,6 +46,15 @@ function calculateMenAverageAge(people, century) {
...
   // write code here
+  const ifWithChildren = withChildren
+    ? people.filter(person => person.sex === 'f'
+      && people.map(child => child.mother)
+        .includes(person.name))
+    : people.filter(person => person.sex === 'f');
+
+  return ifWithChildren
+    .map(person => person.died - person.born)
+    .reduce((ages, age) => ages + age) / ifWithChildren.length;
```
alexandra-protyanova, line 57: this logic repeated two times good idea to move it to separate function

alexandra-protyanova (approved):

PR #1626, opened 2023-01-06
solution

victorionov:


GarryRocksman (approved):
GJ 🔥 Just move an average calculation on a separate function to prevent code duplication

src/averageAges.js
```diff
@@ -14,12 +14,22 @@
...
   // without nesting
+  const filteredMen = people.filter(person => {
+    return century
+      ? person.sex === 'm'
+      && Math.ceil(person.died / 100) === century
+      : person.sex === 'm';
+  });
+
+  return (filteredMen.reduce(
+    (sum, x) => sum + (x.died - x.born), 0) / filteredMen.length);
```
GarryRocksman, line 32: i wrote about this part of the code

GarryRocksman (approved):

src/averageAges.js
Denchepornyuk, line 32: you can create a function for repeating code like this 
```
filteredMen.reduce(
    (sum, x) => sum + (x.died - x.born), 0) / filteredMen.length);
```

Denchepornyuk (approved):

PR #1635, opened 2023-01-07
v-01

AndreyBerezyanskiy:


nzrii-kysel (approved):

src/averageAges.js
```diff
@@ -15,11 +15,23 @@
...
+
+  const totalAge = men.reduce((sum, item) => {
+    const age = item.died - item.born;
+
+    return sum + age;
+  }, 0);
+
+  const totalMen = men.length;
+
+  const averageAge = totalAge / totalMen;
```
Denchepornyuk, line 32: you can remove these variable, code is understandable without these variables

src/averageAges.js
```diff
@@ -37,7 +49,24 @@ function calculateMenAverageAge(people, century) {
...
+    const age = item.died - item.born;
+
+    return sum + age;
+  }, 0);
+
+  const totalWomen = women.length;
+
+  const averageAge = totalAge / totalWomen;
+
+  return averageAge;
```
Denchepornyuk, line 69: the same here about variables, also, you used this code twice, better create a function for it

Denchepornyuk (changes requested):

src/averageAges.js
AndreyBerezyanskiy: > 

I've deleted these variables. But I don't understand where I should use the function. There are two different functions.

AndreyBerezyanskiy: check it out, please

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
+    ? menArray.filter(object => Math.ceil(object.died / 100) === century)
+    : menArray;
+
+  const totalAge = men.reduce((sum, item) => {
+    const age = item.died - item.born;
+
+    return sum + age;
+  }, 0);
+
+  return totalAge / men.length;
```
Denchepornyuk, line 30: you should create function for reduce method, the same logic with reduce you have here and in `calculateWomenAverageAge` function

Denchepornyuk (approved):

PR #1638, opened 2023-01-08
Solution

metalikaqq:


src/averageAges.js
```diff
@@ -15,11 +15,30 @@
...
+    });
+
+    const ages = ageFilter.map(person => {
+      return person.died - person.born;
+    });
+
+    const everageAge = ages.reduce((sum, x) => sum + x);
+
+    return everageAge / ages.length;
+  }
```
maasiimka, line 41: Good, but you can take repeated logic out of if else statement

src/averageAges.js
```diff
@@ -55,7 +79,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  });
+
+  const avarageAgeDifference = children.reduce((ageDiffSum, child) => {
+    const childMother = people.find(mother => mother.name === child.mother);
+    const ageDiff = child.born - childMother.born;
+
+    return ageDiffSum + ageDiff;
+  }, 0) / children.length;
+
+  return avarageAgeDifference;
```
maasiimka, line 103: You don't need this const here.

```suggestion
  return children.reduce((ageDiffSum, child) => {
    const childMother = people.find(mother => mother.name === child.mother);
    const ageDiff = child.born - childMother.born;

    return ageDiffSum + ageDiff;
  }, 0) / children.length;
```

maasiimka (approved):
Good job! Maybe you need to fix some cases, but I think it's not critical

src/averageAges.js
```diff
@@ -15,11 +15,30 @@
...
+  const peopleSexFilter = people.filter(x => x.sex === 'm');
+
+  if (century === undefined || typeof century !== 'number') {
+    const ages = peopleSexFilter.map(person => {
+      return person.died - person.born;
+    });
+    const everageAge = ages.reduce((sum, x) => sum + x);
+
+    return everageAge / ages.length;
+  } else {
```
ArturErmolenko, line 27: avoid `if else `construction use ternary instead

src/averageAges.js
ArturErmolenko: agree make function for repeating logic

src/averageAges.js
ArturErmolenko: agree

ArturErmolenko (changes requested):

Denchepornyuk (approved):

PR #1643, opened 2023-01-08
Solution

maasiimka:


src/averageAges.js
```diff
@@ -38,6 +45,16 @@ function calculateMenAverageAge(people, century) {
...
+  const motherNames = people.reduce((acc, { mother }) => [...acc, mother], []);
+  let women = people.filter(({ sex }) => sex === 'f');
+
+  withChildren && (
+    women = women.filter(({ name }) => motherNames.includes(name))
+  );
+
+  return women.reduce((acc, { born, died }) => (
+    acc + died - born
+  ), 0) / women.length;
```
nzrii-kysel, line 57: you do this logic twice, here and inside `calculateMenAverageAge`, a separate function for it would be useful

nzrii-kysel (approved):

src/averageAges.js
```diff
@@ -16,6 +16,13 @@
  */
 function calculateMenAverageAge(people, century) {
   // write code here
+  let men = people.filter(({ sex }) => sex === 'm');
+
+  century !== undefined && (
+    men = men.filter(({ died }) => Math.ceil(died / 100) === century)
+  );
```
Anna-Tkachenko, line 23: will be better to use ternary operator here, and assign the result of the correct filter call to the men variable

Anna-Tkachenko (changes requested):

src/averageAges.js
```diff
@@ -55,7 +75,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // write code here
+  const names = people.reduce((acc, { name }) => (
+    [...acc, name]
+  ), []);
+
+  let children = people.filter(({ mother }) => (
+    mother && names.includes(mother)
+  ));
+
+  onlyWithSon && (
```
Anna-Tkachenko, line 86: you didn't fix the same issue here (as it was mentioned before), check please 🐝

Anna-Tkachenko (approved):

PR #1652, opened 2023-01-09
solution_1

Borys-Andrew:


0xNazarii (approved):

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(({ sex }) => sex === 'm');
+
+  const filteredByCentury = men.filter(
+    ({ died }) => Math.ceil(died / 100) === century
+  );
```
Denchepornyuk, line 22: what if century is undefined?
you don't have to make filter by century if you don't have century

think about how to combine this logic in the first filter

src/averageAges.js
```diff
@@ -37,7 +46,19 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const women = people.filter(({ sex }) => sex === 'f');
+
+  const womenWithChildren = women.filter(woman => (
+    people.some(child => child.mother === woman.name)));
+
+  const createArrayOfAges = (data) =>
+    data.map((person) => person.died - person.born);
+
+  const avarageAge = (data) => data.reduce((a, b) => a + b, 0) / data.length;
```
Denchepornyuk, line 57: create a separate function for this logic, and don't repeat it in two places

denys-cheporniuk (changes requested):

PR #1655, opened 2023-01-11
Finish task

U-Ra-All:


src/averageAges.js
```diff
@@ -15,11 +15,25 @@
...
+  // debugger;
   // write code here
   // learn how to use array methods like .filter .map .some .every .find .reduce
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  return people
+    .filter((person) => {
+      return century
```
witflash, line 27: You can use implicit return for arrow functions. And here is a such case.

src/averageAges.js
```diff
@@ -15,11 +15,25 @@
...
+        ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+        : person.sex === 'm';
+    })
+    .reduce((ageAccum, currentPerson, index, filteredPeople) => {
+      return (
```
witflash, line 32: The same here

witflash (changes requested):
Could you use implicit return for all arrow functions?
It will be much more readable.

witflash (approved):
Much better!
Well done 👍

U-Ra-All:
> Much better! Well done 👍

Thanks!

PR #1656, opened 2023-01-12
Solution

VadymTsyndra:


src/averageAges.js
```diff
@@ -37,7 +59,24 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  if (withChildren === undefined) {
```
Bohdan-Maksymiuk, line 62:
```suggestion
  if (!withChildren) {
```
it could be like this - no need to compare with undefined
what if it will be `false` ?

src/averageAges.js
```diff
@@ -37,7 +59,24 @@ function calculateMenAverageAge(people, century) {
...
+    const arrOfWomens = people.filter(item => item['sex'] === 'f');
+    const sumOfWomensAge = arrOfWomens.reduce((startValue, person) => {
+      return startValue + (person['died'] - person['born']);
```
Bohdan-Maksymiuk, line 65:
```suggestion
      return startValue + (person.died - person.born);
```
why do you complicate this?)

src/averageAges.js
```diff
@@ -15,13 +15,35 @@
...
+    (person) => {
+      return Math.ceil(person['died'] / 100) === century;
+    },
+  );
+  const sumOfCentury = diedInCentury.reduce(
+    (startValue, person) => startValue + (person['died'] - person['born']), 0
+  );
+  const averageOfCentury = sumOfCentury / diedInCentury.length;
+
+  return averageOfCentury;
```
Bohdan-Maksymiuk, line 39: there are a lot of repeated logic
you can prepare initial array (filtered by century or not) and remain the same for both at the end
DTY (do not repeat yourself)

Bohdan-Maksymiuk (changes requested):

SanyaBratashchuk (changes requested):
image

SanyaBratashchuk (approved):

PR #1658, opened 2023-01-13
Solution

ruslan-yarosh:


src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // without nesting
+  const men = people.filter(person => century
+    ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+    : person.sex === 'm');
+
+  const ages = men.map(person => person.died - person.born);
+
+  const average = ages.reduce((sum, years) => sum + years) / men.length;
+
+  return average;
```
SanyaBratashchuk, line 26: Move this logic to the separate function and reuse it

SanyaBratashchuk (changes requested):

PR #1673, opened 2023-01-23
Solution

kirill-vypirovskyi:


stanislavcodes (approved):

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const rightCentury = century
+    ? people.filter(person => Math.ceil(person.died / 100) === century)
+    : people;
+  const menAges = rightCentury
+    .filter(person => person.sex === 'm')
+    .map(person => person.died - person.born);
```
OleksiiBlezniuk, line 23: Let's improve function performance, why to filter array twice if we can do it only once? Just select needed condition in filter function using ternary operator

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
+
+  return menAges.reduce((age, sum) => sum + age, 0) / menAges.length;
```
OleksiiBlezniuk, line 25: Same here, instead of using map and reduce - you can just calculate age in reduce

OleksiiBlezniuk: And same logic will be used in next function, so will be better to create helper function for it

src/averageAges.js
```diff
@@ -37,7 +40,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const newWomanArray = withChildren
+    ? people.filter(person => people
+      .find(human => human.mother === person.name))
```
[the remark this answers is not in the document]
OleksiiBlezniuk: @stanislavcodes This line wrapping is good 🙂

src/averageAges.js
```diff
@@ -37,7 +40,13 @@ function calculateMenAverageAge(people, century) {
...
+    : people.filter(person => person.sex === 'f');
```
OleksiiBlezniuk, line 46: But you duplicated usage of filter method. Let's follow DRY and move ternary operator inside filter method to just select needed condition to filter

src/averageAges.js
```diff
@@ -37,7 +40,13 @@ function calculateMenAverageAge(people, century) {
...
+  const womanAges = newWomanArray.map(person => person.died - person.born);
+
+  return womanAges.reduce((age, sum) => sum + age, 0) / womanAges.length;
```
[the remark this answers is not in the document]
OleksiiBlezniuk: Agree with @stanislavcodes, but better to iterate array once and calculate age in `reduce` method

src/averageAges.js
```diff
@@ -55,7 +64,19 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const childrenWithMotherInList = people
+    .filter(child => people
+      .some(person => person.name === child.mother));
```
OleksiiBlezniuk, line 69: Very informative name for variable, but please try to make it shorter 🥲

OleksiiBlezniuk: And again, filter array once

src/averageAges.js
```diff
@@ -55,7 +64,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const newCildrenArray = !onlyWithSon
```
OleksiiBlezniuk, line 70: What does `new` mean? And avoid using data type in variables names, in case if it's an array - it means that it's a list, so it's enough to use plural name or `list` in name

src/averageAges.js
```diff
@@ -55,7 +64,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    ? childrenWithMotherInList
+    : childrenWithMotherInList.filter(person => person.sex === 'm');
+  const ages = newCildrenArray.map(child => {
+    const mother = people.find(person => person.name === child.mother);
+
+    return child.born - mother.born;
+  });
+
+  return ages.reduce((age, sum) => sum + age, 0) / ages.length;
```
OleksiiBlezniuk, line 79: Combine this logic, use only reduce

OleksiiBlezniuk (changes requested):

src/averageAges.js
kirill-vypirovskyi: > Very informative name for variable, but please try to make it shorter 🥲

As you wish!

src/averageAges.js
```diff
@@ -15,11 +21,12 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredManArray = people.filter(person => century
+    ? (person.sex === 'm') && (Math.ceil(person.died / 100) === century)
```
sergii-nosachenko, line 25:
```suggestion
    ? (person.sex === 'm' && Math.ceil(person.died / 100) === century)
```

src/averageAges.js
```diff
@@ -37,7 +44,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womanFilteredArray = people
+    .filter(person => withChildren
+      ? people.find(child => child.mother === person.name)
```
sergii-nosachenko, line 49:
```suggestion
      ? people.some(child => child.mother === person.name)
```

sergii-nosachenko: Good idea not to check if `person.sex === 'f'` as only a woman can be a mother 👍

src/averageAges.js
```diff
@@ -55,7 +68,17 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const notOrphans = people.filter(child => people
+    .some(person => onlyWithSon
+      ? (person.name === child.mother) && (child.sex === 'm')
```
sergii-nosachenko, line 73:
```suggestion
      ? (person.name === child.mother && child.sex === 'm')
```

sergii-nosachenko (approved):
Good work 👍
Left some comments.

src/averageAges.js
kirill-vypirovskyi: Why we need outer brackets then?

src/averageAges.js
kirill-vypirovskyi: > 

Why some better then find?

src/averageAges.js
kirill-vypirovskyi: mb then `  ? person.name === child.mother && child.sex === 'm'` ?

kirill-vypirovskyi: I did this brackets only for better readability

src/averageAges.js
kirill-vypirovskyi: The same questions as[ here](https://github.com/mate-academy/js_average-ages/pull/1673#discussion_r1114559542)

sergii-nosachenko: Can be omitted too. But we're using parenthesis for wrapping some meaningful blocks. It's easier to find the start and end.

src/averageAges.js
sergii-nosachenko: `some` returns boolean. Compare:
```js
const list = ['a', 'b', '', 'c'];
const findEmpty = list.find(el => el === '');
const hasEmpty = list.some(el => el === '');

if (findEmpty) <---- FALSE
if (hasEmpty) <---- TRUE
```

PR #1678, opened 2023-01-27
Solution averageAge

NazariiAlieksieiev:


src/averageAges.js
```diff
@@ -20,6 +20,19 @@ function calculateMenAverageAge(people, century) {
...
+
+  const callbackFilterMenIncludesCentury = (person) => person.sex === 'm'
+  && Math.ceil(person.died / 100) === century;
+  const callbackFilterMen = (person) => person.sex === 'm';
+  const callbackAge = (person) => person.died - person.born;
+  const callbackSum = (person1, person2) => person1 + person2;
+
+  const mens = century !== undefined
+    ? people.filter(callbackFilterMenIncludesCentury).map(callbackAge)
+    : people.filter(callbackFilterMen).map(callbackAge);
```
VitaliyBondarenko1982, line 32: better to use ternary operator inside callback function.
You can use destructuring to avoid repeating person.
For example:

```
const men = people.filter(({ sex, died, born }) => century
      ? sex === 'm' && Math.ceil(died / 100) === century
      : sex === 'm';
```

VitaliyBondarenko1982: Tre to replace using filter and map to just one method reduce

src/averageAges.js
```diff
@@ -38,6 +51,17 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+
+  const callbackAge = (person) => person.died - person.born;
+  const callbackSum = (person1, person2) => person1 + person2;
+
+  const women = people.filter(person =>
+    withChildren
+      ? people.find(child => person.name === child.mother)
```
VitaliyBondarenko1982, line 60: Now in each iteration filter method we iterating through all people array.
We can avoid O**2 complaxity. Suggest before filter create object mothers (use reduce method) where key will be mother's name. Value can be any (list of children, true). We just need list of mothers.
```
    acc[person.mother] = true;

    return acc;
  }, {});
  ```
And instead find check whether person is mother.
```
? mothers[person.name]
```

File: src/averageAges.js [hunk 3]
```diff
@@ -56,6 +80,20 @@ function calculateWomenAverageAge(people, withChildren) {
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  const children = people.filter(child =>
+    onlyWithSon
+      ? people.find(mother => mother.name === child.mother)
```
VitaliyBondarenko1982, line 85: Instead find on every filter iteration create before filter object with method reduce
```
const peopleMap = people.reduce((acc, person) => {
    acc[person.name] = person;

    return acc;
  }, {});
  ````
  and instead find 
  ```
  peopleMap[p.mother]
  ````

File: src/averageAges.js (hunk 1 above)
Comment (NazariiAlieksieiev) [reply src/averageAges.js:32]: > 

Could you show example?

File: src/averageAges.js [hunk 4]
```diff
@@ -20,6 +20,15 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  const mens = people.filter(({ sex, died }) => century
```
polosanya, line 24:
```suggestion
  const men = people.filter(({ sex, died }) => century
```

src/averageAges.js
```diff
@@ -20,6 +20,15 @@ function calculateMenAverageAge(people, century) {
...
+    ? sex === 'm' && Math.ceil(died / 100) === century : sex === 'm');
```
polosanya, line 25:
```suggestion
    ? sex === 'm' && Math.ceil(died / 100) === century
    : sex === 'm');
```

src/averageAges.js
```diff
@@ -20,6 +20,15 @@ function calculateMenAverageAge(people, century) {
...
+
+  const result = mens.reduce((accValue, { died, born }) => {
```
polosanya, line 27: Try to use more descriptive names than just result

src/averageAges.js
```diff
@@ -56,6 +82,26 @@ function calculateWomenAverageAge(people, withChildren) {
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  const peopleMap = people.reduce((acc, person) => {
```
polosanya, line 85: What is map?

polosanya (changes requested):

src/averageAges.js
```diff
@@ -56,6 +83,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      : mothers[child.mother]
+  );
+
+  const calculateAge = motherWithChildren.map(person =>
+    person.born - people.find(mother => mother.name === person.mother).born);
+
+  const averageAge = calculateAge.reduce((sum, age) =>
+    sum + age) / calculateAge.length;
+
+  return averageAge;
```
VitaliyBondarenko1982, line 105: Instead of three loops
```suggestion
  let sum = 0;

  const children = people.filter(p => {
    const isPerson = onlyWithSon
      ? p.sex === 'm' && mothers[p.mother]
      : mothers[p.mother];

    if (isPerson) {
      sum += p.born - mothers[p.mother].born;

      return true;
    }

    return false;
  });

  return sum / children.length;
```

VitaliyBondarenko1982: You asked about example. Instead of three loops it can be ome loop.
Also instead filter you can use reduce.
Anyway good job, I just wanted to think about performance in future)

VitaliyBondarenko1982 (approved):

VitaliyBondarenko1982 (approved):

PR #1679, opened 2023-01-27
Solution

bojkovladislav:


src/averageAges.js
```diff
@@ -14,12 +14,24 @@
...
   // learn how to use array methods like .filter .map .some .every .find .reduce
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  const males = century
+    ? people.filter(person => person.sex === 'm'
+  && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === 'm');
+
```
loralevitska, line 29: Fix all cases

```suggestion
  const males = century
    ? people.filter(person => person.sex === 'm'
      && Math.ceil(person.died / 100) === century)
    : people.filter(person => person.sex === 'm');

```

src/averageAges.js
```diff
@@ -14,12 +14,24 @@
...
+  const getAge = males.map(person => person.died - person.born);
+
+  return getAge.length < 2
+    ? getAge[0]
+    : getAge.reduce((a, b) => a + b) / males.length;
```
loralevitska, line 34: This logic repeats itself, put it in a helper function and reuse it.

src/averageAges.js
```diff
@@ -37,7 +49,18 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothersName = people.map(person => person.mother);
+
+  const females = withChildren
+    ? people.filter(person => person.sex === 'f'
+  && mothersName.includes(person.name))
+    : people.filter(person => person.sex === 'f');
+
+  const averageAge = females.map(person => person.died - person.born);
```
loralevitska, line 59: Same here. Move all duplicates to helper functions

loralevitska (changes requested):
Keep up the good work! Some comments to fix:

src/averageAges.js
```diff
@@ -14,12 +14,26 @@
  *
  * @return {number}
  */
+
+const calculateAverageAge = (averageAge, people) => {
```
SanyaBratashchuk, line 18: bad name for parameter, your function has name calculateAverageAge but it already accept average age as parameter. It is confusing.

src/averageAges.js
```diff
@@ -14,12 +14,26 @@
...
+  return averageAge.length < 2
+    ? averageAge[0]
+    : averageAge.reduce((a, b) => a + b) / people.length;
```
SanyaBratashchuk, line 21: Is there a need for this ternary?

SanyaBratashchuk (changes requested):

src/averageAges.js
```diff
@@ -14,12 +14,24 @@
  *
  * @return {number}
  */
+
+const calculateAverageAge = (age, people) => {
+  return age.reduce((a, b) => a + b) / people.length;
+};
+
+const getAge = (people) => people.map(person => person.died - person.born);
```
Volodymyr-Melnyk-Fron10der, line 22: 1. Name `getAge ` seems like function should return a number, not an numbers array as it is now. 
Use something like `getOnlyAges `, `transformToAges`, `convertToAgesArray`, ...etc
2. You can use destructurization and take only properties you need
`const getAge = (people) => people.map(({ died, born }) => died - born);`

src/averageAges.js
```diff
@@ -14,12 +14,24 @@
  *
  * @return {number}
  */
+
+const calculateAverageAge = (age, people) => {
+  return age.reduce((a, b) => a + b) / people.length;
+};
```
Volodymyr-Melnyk-Fron10der, line 20: 1. Why do you need pass people array if `age.length === people.lengt`h already?
2. Again age is a bad name
```suggestion
const calculateAverageAge = (ages) => {
  return ages.reduce((a, b) => a + b) / ages.length;
};
```

src/averageAges.js
```diff
@@ -14,12 +14,24 @@
...
+
+const getAge = (people) => people.map(person => person.died - person.born);
+
+const genderCheck = (people, gender) => {
```
Volodymyr-Melnyk-Fron10der, line 24:
```suggestion
const filterByGender= (people, gender) => {
```

src/averageAges.js
```diff
@@ -37,7 +49,14 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothersName = people.map(person => person.mother);
```
Volodymyr-Melnyk-Fron10der, line 52:
```suggestion
  const mothersWithChildren = people.map(person => person.mother);
```

Volodymyr-Melnyk-Fron10der (changes requested):
Actually good, but try to pay more attention for naming variables

DarkMistyRoom (approved):

PR #1680, opened 2023-01-28
done

ghost:


src/averageAges.js
```diff
@@ -55,7 +67,20 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(child =>
+    onlyWithSon
+      ? people.find(mother => mother.name === child.mother)
+         && child.sex === 'm'
+      : people.find(mother => mother.name === child.mother)
```
Volodymyr-Melnyk-Fron10der, line 74: You can use here the same trick as on line 26-27

Volodymyr-Melnyk-Fron10der (approved):
Looks good👍

PR #1690, opened 2023-02-02
Develop

vanyachyzh:


src/averageAges.js
```diff
@@ -10,16 +10,25 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const diedPeople = people.filter(e => {
+    const deathCentury = Math.ceil(e.died / 100);
+
+    // All men are included if second argument was not passed
+    if (century === deathCentury || century === undefined) {
+      return e.sex === 'm';
```
vlad-demchuk, line 23: no need to leave comments and try to use ternary operator here

vlad-demchuk: use better variable names instead of `e`

vlad-demchuk: check all such cases

src/averageAges.js
```diff
@@ -10,16 +10,25 @@
...
+    }
+  });
+
+  const ageSum = diedPeople.reduce((acc, item) => {
+    return acc + (item.died - item.born);
+  }, 0);
+
+  return Number((ageSum / diedPeople.length).toFixed(2));
```
vlad-demchuk, line 31: create a helper function for calculating average value

src/averageAges.js
```diff
@@ -55,7 +79,31 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    let ageDiff = 0;
+
+    // To get index of woman which is mother of someone
+    if (onlyWithSon) {
+      indexMother = people.findIndex(mother => {
+        return mother.name === child.mother && child.sex === 'm';
+      });
+    } else {
+      indexMother = people.findIndex(mother => mother.name === child.mother);
+    }
```
vlad-demchuk, line 95: the same, it would be better to use ternary operator

vlad-demchuk (changes requested):
remove comments. your code should be clear with your variable names.

Anastasiia-Svintsova (approved):

PR #1691, opened 2023-02-02
● Solution

d-hubych:


src/averageAges.js
```diff
@@ -14,12 +15,19 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let men = people.filter(({ sex }) => sex === 'm');
+
+  century && (men = men
+    .filter(({ died }) => Math.ceil(died / 100) === century));
```
VitaliyBondarenko1982, line 25: You can use just one filter and check century incide callback

src/averageAges.js
```diff
@@ -14,12 +15,19 @@
...
+
+  const lifetimeArr = men.map(({ born, died }) => died - born);
+  const sumAges = lifetimeArr.reduce(reducer);
```
VitaliyBondarenko1982, line 28: It is not required, but i just want to suggest you try to solve this function in one iteration.
In most cases when we should filter and than map same array it is possible to replace on reduce method and do everything in one iteration.
Hint: accamulator can be any structure of data.

src/averageAges.js
```diff
@@ -55,7 +73,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  onlyWithSon && (kids = kids.filter(kid => kid.sex === 'm'));
+
+  const ages = kids.map(kid => {
+    const momObj = people.find(mom => mom.name === kid.mother);
+
+    return kid.born - momObj.born;
+  });
+
+  return ages.reduce(reducer) / ages.length;
```
VitaliyBondarenko1982, line 88:
```suggestion
  const mothers = people.reduce((acc, person) => {
    acc[person.name] = person;

    return acc;
  }, {});

  const { sum, childrenCount } = people.reduce((acc, person) => {
    const { sex, mother, born } = person;

    const isPerson = onlyWithSon
      ? sex === 'm' && mothers[mother]
      : mothers[mother];

    return isPerson ? {
      sum: acc.sum + born - mothers[mother].born,
      childrenCount: acc.childrenCount + 1,
    } : acc;
  }, {
    sum: 0, childrenCount: 0,
  });

  return sum / childrenCount;
```

VitaliyBondarenko1982: It is nor required, I just wanted to show how we can improve performance and make just two itearation on array.

VitaliyBondarenko1982 (approved):
Approve. 
You do not need to re-request review again.

PR #1698, opened 2023-02-04
commit message

Peti51:


qoa11a (changes requested):
let's create a function for calculating average ages as you use it in every other function
also, you don't need to make the same calculation twice. You can just manipulate the correct array. You don't need to do the same logic inside and outside the if statement. Just filter the array of people on some condition, e.g. `people.filter((person) => century ? someFilterLogic : otherFilterLogic)`

apply these changes to every function

src/averageAges.js
```diff
@@ -37,7 +46,21 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const ArrMothers = people.filter((person) => {
+    if (withChildren === true) {
+      return person.sex === 'f' && people.find(
+        (child) => child.mother === person.name);
+    } else {
+      return person.sex === 'f';
```
polosanya, line 54: Same comments here and below. Fix all cases

src/averageAges.js
```diff
@@ -37,7 +46,21 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const ArrMothers = people.filter((person) => {
```
polosanya, line 49:
```suggestion
  const arrMothers = people.filter((person) => {
```
Don't capitalise the first letter

polosanya (changes requested):

src/averageAges.js
```diff
@@ -14,12 +14,21 @@
...
-  // without nesting
+  const manArrCentury = people.filter((person) => {
+    if (century !== undefined) {
+      return person.sex === 'm' && (Math.ceil(person.died / 100)) === century;
+    } else {
+      return person.sex === 'm';
+    };
+  });
+
+  const reducedMan = manArrCentury.reduce(function(person1, person2) {
```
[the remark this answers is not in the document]
polosanya: Not fixed

src/averageAges.js
polosanya: Not fixed

src/averageAges.js
```diff
@@ -55,7 +74,24 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const motherSon = people.map(function(child) {
+    return (onlyWithSon !== undefined
+      ? [child, people.find(
+        (mother) => mother.name === child.mother && child.sex === 'm')]
```
polosanya, line 80: You can put ternary operator inside method find

src/averageAges.js
```diff
@@ -55,7 +74,24 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const motherSon = people.map(function(child) {
```
polosanya, line 77:
```suggestion
  const motherSon = people.map((child) => {
```
It is better to use arrow functions in callbacks

polosanya (changes requested):
Make sure that you've fixed all the problems

src/averageAges.js
```diff
@@ -14,12 +14,19 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const manAges = people.filter((person) => {
+    return (century !== undefined
+      ? person.sex === 'm' && (Math.ceil(person.died / 100)) === century
+      : person.sex === 'm');
```
qoa11a, line 22:
```suggestion
    return century
      ? person.sex === 'm' && (Math.ceil(person.died / 100)) === century
      : person.sex === 'm';
```
fix comparing with undefined everywhere

qoa11a (changes requested):
you should create a function for calculating average age as you use same functionality in three functions

src/averageAges.js
```diff
@@ -14,12 +14,23 @@
...
-  // without nesting
+  const manAges = people.filter((person) => {
+    return (century
+      ? person.sex === 'm' && (Math.ceil(person.died / 100)) === century
+      : person.sex === 'm');
+  });
+
+  const manSumedAges = manAges.map((person1) => {
+    return person1.died - person1.born;
+  }, 0);
```
polosanya, line 31:
```suggestion
  const men = people.filter((person) => {
    return (century
      ? person.sex === 'm' && (Math.ceil(person.died / 100)) === century
      : person.sex === 'm');
  });

  const menAges = men.map((person1) => {
    return person1.died - person1.born;
  }, 0);
```
Okay, now you changed your logic so it's not relevant

src/averageAges.js
```diff
@@ -37,7 +48,19 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const motherAges = people.filter((person) => {
+    return (withChildren
+      ? person.sex === 'f' && people.find(
+        (child) => child.mother === person.name)
+      : person.sex === 'f');
+  });
+
+  const sumOfMotherAges
+  = motherAges.map((person1) => {
```
polosanya, line 59:
```suggestion
  const motherAges = mothers
    .map((mother) => {
```
Same here. 
Also, why is it person1 if it's the only person?

src/averageAges.js
```diff
@@ -55,7 +78,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // write code here
+  const motherAndSon = people.map((child) => {
+    return [child, people.find(
+      (mother) => onlyWithSon
+        ? mother.name === child.mother && child.sex === 'm'
+        : mother.name === child.mother)];
+  });
+
+  const filteredMotherAndSonAges = motherAndSon.filter((person) => {
+    return !person.includes(undefined);
```
polosanya, line 89: It looks a bit complicated. You can just filter to leave only children with mother (and only m if required).
And then using map method transform it into an array with age differences (you can find mother of child inside map method).

polosanya (changes requested):
A few more fixes. Feel free to ask for help in Slack chat if you have any difficulties)

Peti51:
I couldn't make the last function more simple.

polosanya (approved):

PR #1699, opened 2023-02-06
Solution

nadia-hope:


src/averageAges.js
```diff
@@ -37,7 +46,15 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenList = people.filter(person => withChildren
+    ? people.find(child => person.name === child.mother)
```
slavanosatskyi, line 50: you can use some function instead of find here. Find is usually used when you want to find an item and use it somewhere else

src/averageAges.js
```diff
@@ -55,7 +72,17 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(child => onlyWithSon
+    ? people.find(mother => mother.name === child.mother) && child.sex === 'm'
+    : people.find(mother => mother.name === child.mother));
+
+  const diffs = children.map(person =>
+    person.born - people.find(mother => mother.name === person.mother).born);
```
slavanosatskyi, line 80: in general it is quite risky to do such things people.find(mother => mother.name === person.mother).born as find may return undefined. It is better to store in a separate variable and check on validity

slavanosatskyi (changes requested):

src/averageAges.js
```diff
@@ -55,7 +80,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    : people.some(mother => mother.name === child.mother));
+
+  const diffs = children.map(child => {
+    const motherBorn = people.find(mother =>
+      mother.name === child.mother);
+
+    return child.born - motherBorn.born;
+  });
+
+  const AverageDiff = diffs.reduce((sum, currentDiff) =>
```
agniya-i, line 94:
```suggestion
  const averageDiff = diffs.reduce((sum, currentDiff) =>
```

agniya-i (approved):

PR #1704, opened 2023-02-08
Solution

vlad-ptashnichenko:


src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person =>
+    century
+      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+      : person.sex === 'm'
+  );
+  const age = men.map(item =>
+    item.died - item.born).reduce((sum, item) => sum + item);
```
etojeDenys, line 24: i think you can write it using only reduce method

src/averageAges.js
```diff
@@ -37,7 +41,16 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mother = people.map(item => item.mother);
+  const women = people.filter(person =>
+    withChildren
+      ? person.sex === 'f' && mother.includes(person.name)
+      : person.sex === 'f'
+  );
+  const age = women.map(item =>
+    item.died - item.born).reduce((sum, item) => sum + item);
```
etojeDenys, line 51: same here

src/averageAges.js
```diff
@@ -37,7 +41,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mother = people.map(item => item.mother);
+  const women = people.filter(person =>
+    withChildren
+      ? person.sex === 'f' && mother.includes(person.name)
```
etojeDenys, line 47: try to use here find method and `===` operator

src/averageAges.js
```diff
@@ -55,7 +68,19 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const ChildrensName = people.filter(child =>
```
etojeDenys, line 71:
```suggestion
  const childrenNames = people.filter(child =>
```

etojeDenys (changes requested):

IvanFesenko (approved):

PR #1705, opened 2023-02-08
solution

xnquaint:


src/averageAges.js
```diff
@@ -37,7 +47,14 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  return findAverage(
+    people.filter(person =>
+      withChildren
+        ? person.sex === 'f'
+          && people.some(individual => individual.mother === person.name)
```
tetiana-ivanchenko, line 54:
```suggestion
        ? people.some(child => child.mother === person.name)
```

tetiana-ivanchenko (changes requested):

src/averageAges.js
```diff
@@ -55,7 +71,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      ? kid.sex === 'm'
+        && people.find(person => person.name === kid.mother)
+      : people.find(person => person.name === kid.mother)
+  );
+
+  return children.reduce((sum, kid) => {
+    mother = people.find(person => person.name === kid.mother);
+
+    return sum + kid.born - mother.born;
+  }, 0) / children.length;
```
VitaliyBondarenko1982, line 86: Good job.
Approve. 
I just want to show example how we can avoid O**2 complexity and solve it just in two iterations
```
    acc[person.name] = person;

    return acc;
  }, {});

  const { sum, childrenCount } = people.reduce((acc, person) => {
    const { sex, mother, born } = person;

    const isPerson = onlyWithSon
      ? sex === 'm' && mothers[mother]
      : mothers[mother];

    return isPerson ? {
      sum: acc.sum + born - mothers[mother].born,
      childrenCount: acc.childrenCount + 1,
    } : acc;
  }, {
    sum: 0, childrenCount: 0,
  });

  return sum / childrenCount;
  ```
  of course something like that posiible to make with another functions.

VitaliyBondarenko1982 (approved):

PR #1710, opened 2023-02-10
Solution

AnastasiiaEm:


src/averageAges.js
```diff
@@ -55,7 +76,22 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(child => people.some(mother => {
+    return onlyWithSon
+      ? child.sex === 'm' && child.mother === mother.name
+      : child.mother === mother.name;
+  }));
```
etojeDenys, line 83: let`s rewrite and improve readability

```
const children = people.filter(child => {
    const hasMother = people.some(mother => mother.name === child.mother);
    const isMan = child.sex === 'm';

    return onlyWithSon
      ? hasMother && isMan
      : hasMother;
  });

```

src/averageAges.js
```diff
@@ -55,7 +76,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const ages = children.map((child) => {
+    const motherBorn = people.find(mother => mother.name === child.mother).born;
+    const difference = child.born - motherBorn;
+
+    return difference;
+  });
+
+  const average = ages.reduce((a, b) => a + b, 0);
+
+  return average / ages.length;
```
etojeDenys, line 94: we can write it much more simple

```

  const difference = children.reduce((ageSum, child) => {
    const mother = people.find(woman => woman.name === child.mother);

    return ageSum + (child.born - mother.born);
  }, 0);

  return difference / children.length;
```

etojeDenys (changes requested):

IvanFesenko (approved):

PR #1717, opened 2023-02-13
Initial commit

RomanHrytsiuk:


src/averageAges.js
```diff
@@ -15,11 +16,20 @@
...
+  const speacialMen = people.filter(person =>
+    Math.ceil(person.died / 100) === century && person.sex === 'm');
+
+  const speacialSum = speacialMen.reduce((sum, { died, born }) =>
+    (sum + (died - born)), 0);
+
+  const allMen = people.filter(person => person.sex === 'm');
+
+  const generalSum = allMen.reduce((sum, { died, born }) =>
+    (sum + (died - born)), 0);
```
DarkMistyRoom, line 28: it would be better to check if there is `century` argument in the very beginning and make all calculations according to it, to avoid unnecessary actions

src/averageAges.js
```diff
@@ -37,16 +47,31 @@ function calculateMenAverageAge(people, century) {
...
+
+  const womenWithChildren = people.filter(person =>
+    person.sex === 'f' && people.find(woman => woman.mother === person.name));
+
+  const difference = womenWithChildren.reduce((sum, { died, born }) =>
+    sum + (died - born), 0);
+
+  return withChildren === true
+    ? difference / womenWithChildren.length
+    : age / allWomen.length;
```
DarkMistyRoom, line 63: the same here. You can check it in the beginning and create the sample based on this arg

DarkMistyRoom (changes requested):

src/averageAges.js
```diff
@@ -15,11 +16,19 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filterCallBack = person =>
+    arguments.length === 2
+      ? Math.ceil(person.died / 100) === century && person.sex === 'm'
+      : person.sex === 'm';
```
ViktorHavryliuk, line 22: Lenght of arguments is not the best check here. For example, if function is called like `calculateMenAverageAge(someArray, null)`, arguments lenght will still be two and function will fail

```suggestion
  const filterCallBack = person =>
    century
      ? Math.ceil(person.died / 100) === century && person.sex === 'm'
      : person.sex === 'm';
```

ViktorHavryliuk (changes requested):

ViktorHavryliuk (approved):

PR #1719, opened 2023-02-14
Solution

RuslanLiapin:


src/averageAges.js
```diff
@@ -20,8 +20,20 @@ function calculateMenAverageAge(people, century) {
...
   // replace `if ()` statement with &&, || or ?:
   // without nesting
-}
+  let men = people.filter(obj => {
+    return obj['sex'] === 'm';
+  });
 
+  if (century) {
+    men = men.filter(obj => Math.ceil(obj.died / 100) === century);
+  };
```
Viktor-Kost, line 29: use ternary operator instead of resign value. Apply for other cases as well
Ex: 
```
const yourVariable = array.filter(val => 
    century
      ? {{some_code}}
      : {{some_code}}
)
```

src/averageAges.js
```diff
@@ -36,9 +48,25 @@ function calculateMenAverageAge(people, century) {
...
+    women = women.filter(mother => {
+      return people.some(child => mother.name === child.mother);
+    });
+  };
+
+  const totalAge = women.reduce((total, obj) => {
+    return total + (obj.died - obj.born);
+  }, 0);
+
+  return totalAge / women.length;
```
Viktor-Kost, line 68: You've already used these code twice. So, create a separated function to make these calculations

Viktor-Kost (changes requested):

src/averageAges.js
```diff
@@ -36,9 +50,17 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
-}
+  const women = people.filter(obj =>
```
TarasHoliuk, line 56: `obj`, `p` - bad names for variables
use meaningful names, e.g `person`

FYI

TarasHoliuk (approved):
GJ :+1: 
Approved, since you've fixed previous comments

But I've added one comment about naming FYI (no need to re-request review)

PR #1720, opened 2023-02-16
averageSolution

edkido:


src/averageAges.js
```diff
@@ -14,12 +14,19 @@
  *
  * @return {number}
  */
+
+function calculateAverageAge(people) {
+  return people.map(person => person.died - person.born)
+    .reduce((sum, elem) => sum + elem, 0) / people.length;
```
IvanFesenko, line 20: map is redudant here. You can do it with reduce only

IvanFesenko (approved):
Good job

PR #1731, opened 2023-02-20
solution

toljan1:


src/averageAges.js
```diff
@@ -37,7 +48,20 @@ function calculateMenAverageAge(people, century) {
...
+
+  if (withChildren !== undefined) {
+    const peopleMothers = people.map(person => person.mother);
+
+    ages = [...womens]
+      .filter(women => peopleMothers.includes(women.name))
+      .map(women => women.died - women.born);
+  } else {
+    ages = [...womens].map(women => women.died - women.born);
+  }
```
VitaliyBondarenko1982, line 62: Please try to silmlify it, just filter people whicjh you need and make withChildren checking inside a filter callback.
Avoid if else.

src/averageAges.js
```diff
@@ -15,7 +15,18 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
+  let ages;
+  const mens = people.filter(person => person.sex === 'm');
+
+  if (century !== undefined) {
```
VitaliyBondarenko1982, line 21: avoid if else. make checking on century inside filter funcion and just filter people. You don't need additional map iteration

src/averageAges.js
```diff
@@ -55,7 +79,19 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = onlyWithSon
+    ? people.filter(person => person.sex === 'm' && people.some(woman =>
+      woman.name === person.mother))
+    : people
+      .filter(person => people.some(woman => woman.name === person.mother));
```
VitaliyBondarenko1982, line 86: About comments above, something like that as here.
But as for me better to check onlyWithSon insilde callback function.
Will be less repeating code.

VitaliyBondarenko1982 (changes requested):

src/averageAges.js
```diff
@@ -15,7 +15,15 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
+  const mens = people.filter(person => person.sex === 'm');
+
+  const ages = century
+    ? [...mens]
+      .filter(men => Math.ceil(men.died / 100) === century)
+      .map(men => men.died - men.born)
+    : [...mens].map(men => men.died - men.born);
```
IvanFesenko, line 24: No need to create copy of array here
```suggestion
  const ages = century
    ? mens.filter(men => Math.ceil(men.died / 100) === century)
      .map(men => men.died - men.born)
    : mens.map(men => men.died - men.born);
```

IvanFesenko (approved):

PR #1736, opened 2023-02-21
Solution

BogdanParubchenk01:


src/averageAges.js
```diff
@@ -16,6 +16,18 @@
...
+  let ages;
+  const mens = people.filter(person => person.sex === 'm');
+
+  if (century !== undefined) {
+    ages = [...mens]
+      .filter(men => Math.ceil(men.died / 100) === century)
+      .map(men => men.died - men.born);
+  } else {
+    ages = [...mens].map(men => men.died - men.born);
+  }
```
IvanFesenko, line 28: `Array.filter()` returns new array, no need creting a copy.
Use ternary operator instead of `if else` statement

src/averageAges.js
```diff
@@ -38,6 +50,20 @@ function calculateMenAverageAge(people, century) {
...
+
+  if (withChildren !== undefined) {
+    const peopleMothers = people.map(person => person.mother);
+
+    ages = [...womens]
+      .filter(women => peopleMothers.includes(women.name))
+      .map(women => women.died - women.born);
+  } else {
+    ages = [...womens].map(women => women.died - women.born);
+  }
```
IvanFesenko, line 64: Same here as comment above

src/averageAges.js
```diff
@@ -38,6 +50,20 @@ function calculateMenAverageAge(people, century) {
...
+
+  return ages.reduce((accum, current) => accum + current, 0) / ages.length;
```
IvanFesenko, line 66: It could be a separate function. You make same in previous function

IvanFesenko (changes requested):
Good job. Check my comments

PR #1741, opened 2023-02-21
Average Ages - Solution

Yevhenii0536:


src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
-  // without nesting
+  const sortByCentury = century
+    ? people.filter((person) => Math.ceil(person.died / 100) === century)
+    : people;
+
+  const sortByGender = sortByCentury.filter((person) => person.sex === 'm');
+
+  const averageAge = sortByGender.reduce((acc, person) => acc
+    + (person.died - person.born), 0)
+      / sortByGender.length;
```
KuharchukDmytro, line 26: on my opinion, its better to separate this part of code by several variables, where you will find every person's age first, and then find average age
its improve readability of code

KuharchukDmytro (changes requested):
nice work, but there is one little mistake with checklist, fix it please)

src/averageAges.js
Yevhenii0536: as for me my var has own name and inside her name we can see about what she is

Yevhenii0536: but ok, i'll add your advice

KuharchukDmytro (approved):
gj!

src/averageAges.js
```diff
@@ -1,61 +1,70 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const sortByCentury = century
+    ? people.filter((person) => Math.ceil(person.died / 100) === century)
+    : people;
+
+  const sortByGender = sortByCentury.filter((person) => person.sex === 'm');
```
sergii-nosachenko, line 14: Do this in one `filter` call instead of 2.

src/averageAges.js
```diff
@@ -1,61 +1,70 @@
...
  *
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter((person) => person.sex === 'f');
+  const motherNames = people.map((person) => person.mother);
+  const womenWithChildren = women.filter((person) => (
+    motherNames.includes(person.name))
+  );
```
sergii-nosachenko, line 34: Do this in one `filter` call instead of these 3 methods.

src/averageAges.js
```diff
@@ -1,61 +1,70 @@
...
+
+  const averageAge = withChildren
+    ? (womenWithChildren
+      .reduce((acc, person) => acc + (person.died - person.born), 0))
+      / womenWithChildren.length
+    : (women.reduce((acc, person) => acc + (person.died - person.born), 0)
+      / women.length
+    );
```
sergii-nosachenko, line 42: Ternary is for simple data.
Do all logic inside `reduce` instead of making this monstrous ternary.

sergii-nosachenko (changes requested):

src/averageAges.js
```diff
@@ -1,61 +1,65 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const sortedPeople = century
+    ? people.filter((person) => Math.ceil(person.died / 100) === century
+      && person.sex === 'm')
+    : people.filter((person) => person.sex === 'm');
```
Andr1yK, line 13: let's use single filter (DRY)

src/averageAges.js
```diff
@@ -1,61 +1,65 @@
...
  * @param {boolean} withChildren - optional
  *
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter((person) => person.sex === 'f');
+  const womenWithChildren = women.filter((woman) =>
+    people.some((person) => person.mother === woman.name)
+  );
```
Andr1yK, line 32: same here

Andr1yK (approved):

PR #1745, opened 2023-02-22
Solution for Average Ages

ctk12:


src/averageAges.js
```diff
@@ -15,11 +15,13 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const results = people.filter((data) =>
```
polosanya, line 18:
```suggestion
  const men = people.filter((person) =>
```
Use more meaningful naming. Your code should be understandable and descriptive

src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
+    !century ? data.sex === 'm'
+      : data.sex === 'm' && Math.ceil(data.died / 100) === century);
```
polosanya, line 20:
```suggestion
    !century 
      ? data.sex === 'm'
      : data.sex === 'm' && Math.ceil(data.died / 100) === century);
```
Also, it's better to check if century is present, not its absence

src/averageAges.js
```diff
@@ -37,7 +39,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const results = people.filter((data) =>
+    !withChildren ? data.sex === 'f'
+      : data.sex === 'f' && people.some(val => val.mother === data.name));
```
polosanya, line 44: Same here

src/averageAges.js
```diff
@@ -55,7 +63,19 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const results = onlyWithSon
+    ? people.filter(data => data.sex === 'm'
+      && people.some(mother => mother.name === data.mother))
+    : people.filter(data =>
+      people.some(mother => mother.name === data.mother));
```
polosanya, line 70: In both cases you filter the people. You can just move all the check inside callback of one filter method

polosanya (changes requested):
Please use more meaningful naming

PR #1746, opened 2023-02-22
did program

vadosik15:


src/averageAges.js
```diff
@@ -20,6 +20,11 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const mens = people.filter(({ sex, died }) => sex === 'm'
+    && (century ? century === Math.ceil(died / 100) : true));
+
```
olecsiy-gavrilyuk, line 25: man - its 1 , men - its many , mens doesnt corectly, and you should write ternary operation like this 
`century `
`? century === Math.ceil(died / 100)`
`: true)`

src/averageAges.js
```diff
@@ -37,7 +42,19 @@ function calculateMenAverageAge(people, century) {
...
+    const memsWidthSon = people.filter(({ name, sex }) =>
+      sex === 'f' && people
+        .some(({ mother }) => mother === name));
+
+    return Math.round(((memsWidthSon.reduce((sum, { born, died }) =>
+      sum + (died - born), 0) / memsWidthSon.length) * 100)) / 100;
+  }
+
+  const mems = people.filter(({ sex }) => sex === 'f');
+
```
olecsiy-gavrilyuk, line 55: mems?

olecsiy-gavrilyuk (approved):

src/averageAges.js
OleksiiBlezniuk: Agree, let's format code

src/averageAges.js
```diff
@@ -20,6 +20,11 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const mens = people.filter(({ sex, died }) => sex === 'm'
+    && (century ? century === Math.ceil(died / 100) : true));
+
+  return Math.round(((mens.reduce((sum, { born, died }) =>
+    sum + (died - born), 0) / mens.length) * 100)) / 100;
```
OleksiiBlezniuk, line 27: This logic will be reused in next function too, let's make a separate helper function for it
Also better to move reduce result to constant

src/averageAges.js
```diff
@@ -37,7 +42,19 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  if (withChildren) {
+    const memsWidthSon = people.filter(({ name, sex }) =>
```
OleksiiBlezniuk, line 46:
```suggestion
    const womanWithSon = people.filter(({ name, sex }) =>
```

src/averageAges.js
```diff
@@ -37,7 +42,19 @@ function calculateMenAverageAge(people, century) {
...
+      sex === 'f' && people
+        .some(({ mother }) => mother === name));
+
+    return Math.round(((memsWidthSon.reduce((sum, { born, died }) =>
+      sum + (died - born), 0) / memsWidthSon.length) * 100)) / 100;
+  }
+
+  const mems = people.filter(({ sex }) => sex === 'f');
```
OleksiiBlezniuk, line 54: Let's move this `withChildren` check to filter method instead of duplicating logic

src/averageAges.js
```diff
@@ -37,7 +42,19 @@ function calculateMenAverageAge(people, century) {
...
+
+  return Math.round(((mems.reduce((sum, { born, died }) =>
+    sum + (died - born), 0) / mems.length) * 100)) / 100;
```
OleksiiBlezniuk, line 57: Follow DRY, create helper for it

src/averageAges.js
```diff
@@ -55,7 +72,26 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const peopleWithMother = people.filter(person => people
+    .some(mother => (
+      mother.name === person.mother && (
+        onlyWithSon
+          ? person.sex === 'm'
+          : true
```
OleksiiBlezniuk, line 80: Also would be nice to move these checks to constants, but it's ok too

src/averageAges.js
```diff
@@ -55,7 +72,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const sumOfDifferences = peopleWithMother
+    .reduce((sum, { born: personBorn, mother: motherName }) => {
+      const motherBorn = people
+        .find((person) => person.name === motherName)
+        .born;
+
+      return sum + personBorn - motherBorn;
+    }, 0);
+
+  return Math.round(((sumOfDifferences / peopleWithMother.length) * 100)) / 100;
```
OleksiiBlezniuk, line 94: Actually this task doesn't require rounding, so you can remove it (but you may leave as it is if you want)

OleksiiBlezniuk (changes requested):

illiarerun (approved):

PR #1750, opened 2023-02-22
solution for average ages task

stanislavkulakovskyi:


Lilu580 (commented):
Gj

src/averageAges.js
```diff
@@ -14,12 +14,20 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let men = people.filter((person) => person.sex === 'm');
+
+  men = century
+    ? men = men.filter((person) => Math.ceil(person.died / 100) === century)
+    : men;
```
sergii-nosachenko, line 28: Better do all logic inside one `filter` call.

src/averageAges.js
```diff
@@ -55,7 +68,20 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  let kids = people.filter(kid => people.some(person =>
+    kid.mother === person.name));
+
+  kids = onlyWithSon
+    ? kids.filter(kid => kid.sex === 'm')
+    : kids;
```
sergii-nosachenko, line 76: One filter to get needed data.

src/averageAges.js
```diff
@@ -55,7 +68,20 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const mothers = kids.map(kid =>
+    people.find(mother => mother.name === kid.mother));
+  const ageDifference = kids.map((kid, index) =>
+    kid.born - mothers[index].born);
+  const totalDifference = ageDifference.reduce((sum, age) => sum + age, 0);
```
sergii-nosachenko, line 82: Instead of having two `map` loops, move logic inside the `reduce` call. It will be one array loop instead of 3.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1759, opened 2023-02-23
Solution with loops

Zelinskyi-Serhii:


src/averageAges.js
```diff
@@ -54,8 +75,30 @@ function calculateWomenAverageAge(people, withChildren) {
  *
  * @return {number}
  */
-function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+function calculateAverageAgeDiff(people, onlyWithSon = false) {
+  const mothers = people.map(person => person.mother);
+  const children = people.map(person => person.name);
```
andriibezkorovainyi, line 80: There is not a single man in people array who has `null` as name, you don't need this operation.

andriibezkorovainyi (changes requested):
Great job, mate! I've found some changes I think are needed for you to don't be rejected by mentor. Fix them and you will definitely go through.

andriibezkorovainyi (approved):
GJ

Andr1yK (approved):

PR #1786, opened 2023-02-24
Solution

dianaavem:


src/averageAges.js
```diff
@@ -55,7 +79,16 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const motherAgeDiffs = people.reduce((acc, person) =>
+    onlyWithSon && person.sex !== 'm' ? acc
+      : !people.find(p => p.name === person.mother) ? acc
+        : (acc.push(person.born - people.find(p =>
+          p.name === person.mother).born), acc), []);
```
loralevitska, line 86: Don't use nested ternary operator. Use variables and helper functions

loralevitska (changes requested):
A little more and it will be completed

DarkMistyRoom (changes requested):
I see no commit since last review(maybe you forget to push the changes?

PR #1793, opened 2023-02-26
'Solution'

nazarkharkevych:


src/averageAges.js
```diff
@@ -15,6 +15,19 @@
...
 function calculateMenAverageAge(people, century) {
+  const arrayOfMen = people.filter(person => {
+    return arguments.length < 2
+      ? person.sex === 'm'
+      : Math.ceil(person.died / 100) === century && person.sex === 'm';
+  });
+
+  const sumOfAges = arrayOfMen.reduce((acc, man) => {
+    return acc + (man.died - man.born);
+  }, 0);
```
Volodymyr-Melnyk-Fron10der, line 26: move this reduce into separated function and reuse throughout your code. Because you have duplicated this code in a few places

src/averageAges.js
```diff
@@ -15,6 +15,19 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  const arrayOfMen = people.filter(person => {
+    return arguments.length < 2
```
Volodymyr-Melnyk-Fron10der, line 19: what if `century === false` ? then `arguments.length < 2` will return `false `and `Math.ceil(person.died / 100) === century && person.sex === 'm';` will be executed, whereas it's not expected behavior

src/averageAges.js
```diff
@@ -15,6 +15,19 @@
...
+      ? person.sex === 'm'
+      : Math.ceil(person.died / 100) === century && person.sex === 'm';
```
Volodymyr-Melnyk-Fron10der, line 21: move `person.sex === 'm'` out of condition as it's common value `person.sex === 'm' && (century ? ... : .. )`

src/averageAges.js
```diff
@@ -15,6 +15,19 @@
...
+  });
+
+  const sumOfAges = arrayOfMen.reduce((acc, man) => {
+    return acc + (man.died - man.born);
+  }, 0);
+
+  const averageAge = sumOfAges / arrayOfMen.length;
+
+  return averageAge;
```
Volodymyr-Melnyk-Fron10der, line 30: you can return value right away
```suggestion
  return  sumOfAges / arrayOfMen.length;
```

src/averageAges.js
```diff
@@ -55,7 +84,39 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+    return acc + ageDiff;
+  }, 0);
+
+  function ageDiffOfChildren(women) {
+    const children = people.filter(person => {
+      return onlyWithSon
+        ? person.mother === women.name && person.sex === 'm'
+        : person.mother === women.name;
+    });
```
Volodymyr-Melnyk-Fron10der, line 106: just move it out of `ageDiffOfChildren` and you will not be needed to use `childrenCount`  variable, you would use `children.length` instead

src/averageAges.js
```diff
@@ -55,7 +84,39 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  let childrenCount = 0;
+
+  const totalAgeDiff = womenWithChildren.reduce((acc, women) => {
+    const ageDiff = ageDiffOfChildren(women);
+
+    return acc + ageDiff;
+  }, 0);
+
+  function ageDiffOfChildren(women) {
```
Volodymyr-Melnyk-Fron10der, line 101: try to get rid of this function, you use it only once, so it doesn't worth to move the code into separated function

Volodymyr-Melnyk-Fron10der (changes requested):
Good job, let's improve your code a little bit

anstsot (approved):
Great job, approved!

PR #1796, opened 2023-02-26
Solution

mstrilec:


src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const mens = people.filter(({ sex, died }) => (
+    sex === 'm' && (century ? century === Math.ceil(died / 100) : true)
+  ));
```
Andr1yK, line 20: what happened if I pass century = 0?

src/averageAges.js
```diff
@@ -55,7 +66,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  });
+
+  const differences = children.map((child) => {
+    const mother = people.find((mom) => mom.name === child.mother);
+
+    return child.born - mother.born;
+  });
+
+  return differences.reduce((acc, diff) => acc + diff, 0) / differences.length;
+}
```
Andr1yK, line 86: we can use fewer methods.. so let's do it)

Andr1yK (changes requested):

PR #1798, opened 2023-02-27
Average ages

VadKol:


src/averageAges.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+// const people = require("./people");
```
Andr1yK, line 3: do u need it?

src/averageAges.js
```diff
@@ -15,11 +17,16 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+
+  const menInCentury = century
+    ? men.filter(person => Math.ceil(person.died / 100) === century)
+    : men;
```
Andr1yK, line 24: let's use single filter

src/averageAges.js
```diff
@@ -55,7 +71,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    if (mother) {
+      return child.born - mother.born;
+    } else {
+      return undefined;
+    }
+  }).filter(diff => diff !== undefined);
+
+  const sumOfDiffs = diffs.reduce((total, diff) => {
+    return total + diff;
+  }, 0);
```
Andr1yK, line 90: let's use less array methods))

Andr1yK (changes requested):

src/averageAges.js
VadKol: Fixed ✔

src/averageAges.js
VadKol: I got it) i'll refactor code to using less duplicates

Andr1yK (approved):

PR #1799, opened 2023-02-27
Solution

Zavalishyn-Oleksandr:


src/averageAges.js
```diff
@@ -20,6 +20,14 @@ function calculateMenAverageAge(people, century) {
...
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const arrayWithMen = people.filter(el => el.sex === 'm');
+  const diedThisCentury = century
+    ? arrayWithMen.filter(el => Math.ceil(el.died / 100) === century)
+    : arrayWithMen;
+  const sumOfAges = diedThisCentury.reduce((sum, num) =>
+    sum + (num.died - num.born), 0);
+
+  return sumOfAges / diedThisCentury.length;
```
DarkMistyRoom, line 30: would be great to move this logic to a separate function and reuse in next fuction

src/averageAges.js
```diff
@@ -37,7 +45,18 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const arrayWithWoman = people.filter(el => el.sex === 'f');
+  const womanWithKids = arrayWithWoman.filter(woman => {
+    return people.some(child => woman.name === child.mother);
+  });
+
+  const sumOfAges = withChildren
+    ? womanWithKids.reduce((sum, num) => sum + (num.died - num.born), 0)
+    : arrayWithWoman.reduce((sum, num) => sum + (num.died - num.born), 0);
```
DarkMistyRoom, line 55: use a variable to set needed sample and word only with it. you have a lot of duplicating here

src/averageAges.js
```diff
@@ -55,7 +74,37 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      if (currentMother.length > 0) {
+        counter++;
+      }
+
+      return currentMother.length > 0 ? person.born - currentMother[0].born : 0;
+    });
+    const sumOfAges = ageDifference.reduce((sum, num) => sum + num, 0);
+
+    return sumOfAges / counter;
+  }
```
DarkMistyRoom, line 107: same. avoid code repeating. `onlyWithSon` value effect only the sample you should work with

DarkMistyRoom (changes requested):

Zavalishyn-Oleksandr:
Thank you, fixed

anstsot (changes requested):
Looks good but let's make you code a little better

PR #1807, opened 2023-02-28
solution

valeriy-mikheyev:


src/averageAges.js
```diff
@@ -37,7 +42,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womens = withChildren
+    ? people.filter(woman =>
+      (people.find(child => woman.name === child.mother)))
```
slavanosatskyi, line 47: use people.some(...) instead. find is used usually when you need to find an item and perform some actions with it. In this case you just need to check that there is such person with specified name

slavanosatskyi (changes requested):
Nice, only one minor comment to be fixed

src/averageAges.js
```diff
@@ -55,7 +69,20 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(child =>
+    onlyWithSon
+      ? people.find(mother => mother.name === child.mother)
```
slavanosatskyi, line 74: you would need "some" function here as well

slavanosatskyi (approved):
Accepted, but you missed two places where you could apply some function as well. For the future try to look through your code and find the same places which could be improved according to mentors' comments

PR #1811, opened 2023-03-02
'done'

Hanna-Balabukha:


slavanosatskyi (approved):
Very nice! the only thing that I would improve is naming - arr and x are not the best variable names

PR #1813, opened 2023-03-03
solution :)

katerynashylina:


src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
+  const male = century
+    ? people.filter(person => person.sex === 'm'
+    && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === 'm');
+
+  const maleAge = male.map(person => person.died - person.born);
+
+  const averageAge = maleAge.reduce((a, b) => a + b) / maleAge.length;
+
+  return averageAge;
```
SanyaBratashchuk, line 27: Move this logic to the separated function and reuse it in your main functions to avoid duplication

src/averageAges.js
```diff
@@ -37,7 +42,18 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  // eslint-disable-next-line max-len
+  const isMother = people.filter(person => person.mother).map(person => person.mother);
+  const woman = withChildren
+    ? people.filter(person => person.sex === 'f'
+    && isMother.includes(person.name))
```
SanyaBratashchuk, line 49: instead of filter with includes you can check if woman is mother via `some` method

SanyaBratashchuk (changes requested):

src/averageAges.js
```diff
@@ -14,12 +14,17 @@
...
  */
+const age = (sum, person) => sum + (person.died - person.born);
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const personMen = (person) => century
```
Volodymyr-Melnyk-Fron10der, line 20:
```suggestion
  const personMenCallback = (person) => century
```

src/averageAges.js
```diff
@@ -14,12 +14,17 @@
...
+    ? person.sex === 'm' && century === Math.ceil(person.died / 100)
+    : person.sex === 'm';
+
+  const men = people.filter(personMen);
+  const result = men.reduce(age, 0) / men.length;
+
+  return result;
 }
```
Volodymyr-Melnyk-Fron10der, line 28:
```suggestion
  return men.reduce(age, 0) / men.length;
}
```

Volodymyr-Melnyk-Fron10der: fix everywhere

src/averageAges.js
```diff
@@ -14,12 +14,17 @@
  *
  * @return {number}
  */
+const age = (sum, person) => sum + (person.died - person.born);
```
Volodymyr-Melnyk-Fron10der, line 17:
```suggestion
const claculateAverage = people => people.reduce((sum, person) => sum + (person.died - person.born), 0) / people.length ;
```

src/averageAges.js
```diff
@@ -55,7 +67,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const child = onlyWithSon
+    ? people.filter(person => people.find(
+      mother => person.mother === mother.name
+    ) && person.sex === 'm')
+    : people.filter(person => people.find(
+      mother => person.mother === mother.name
+    ));
```
Volodymyr-Melnyk-Fron10der, line 76: move this filter out of condition to not duplicate it

src/averageAges.js
```diff
@@ -55,7 +67,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const difference = child.map(person => person.born
```
Volodymyr-Melnyk-Fron10der, line 78: use destructurizing structure whenever it's possible

Volodymyr-Melnyk-Fron10der (changes requested):

PR #1814, opened 2023-03-03
Solution

dudar-y:


src/averageAges.js
```diff
@@ -55,7 +67,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
+}
+
+// Some auxiliary function
+
+function averageAge(people) {
+  return people.reduce((sum, person) =>
+    sum + (person.died - person.born), 0) / people.length;
+}
+
+function setGender(people, sex) {
```
Volodymyr-Melnyk-Fron10der, line 95:
```suggestion
function filterByGender(people, sex) {
```

src/averageAges.js
```diff
@@ -55,7 +67,33 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const allChild = (!onlyWithSon)
+    ? people.filter(child => people
+      .some(women => women.name === child.mother))
+    : setGender(people, 'm').filter(child => people
+      .some(women => women.name === child.mother));
```
Volodymyr-Melnyk-Fron10der, line 74: move `some` method out of condition as it's applying in both cases

Volodymyr-Melnyk-Fron10der (changes requested):
I left comments + use destructurizing structure as much as possible throughout your code

polosanya (changes requested):
Almost done!

src/averageAges.js
```diff
@@ -55,7 +69,36 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const mens = filterByGender(people, 'm');
+  const allChild = !onlyWithSon
+    ? people.filter(child => findMother(people, child))
+    : mens.filter(child => findMother(people, child));
```
DarkMistyRoom, line 75: we can optimize it. there is no need to filter men if `onlyWithSon `argument isn't passed. So better use this query for creating sample(men or all people) where we will `findMother`.

DarkMistyRoom (changes requested):
Let's improve a bit

Volodymyr-Melnyk-Fron10der (approved):
Great now!

PR #1817, opened 2023-03-06
Solution

valeriaoleynik:


qoa11a (changes requested):

src/averageAges.js
```diff
@@ -14,12 +14,19 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const onlyMen = people.filter(person =>
+    century
+      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+      : person.sex === 'm'
+  );
+
+  return +(averageAge(onlyMen)).toFixed(2);
```
agniya-i, line 29: .toFixed() also repeated several times, you can move it inside `averageAge` funtion to call just this function `return averageAge(onlyMen)`

agniya-i (changes requested):

nazarmatsevych (approved):

PR #1825, opened 2023-03-11
Solution

zvir91:


src/averageAges.js
```diff
@@ -15,6 +15,23 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  // const centuryOfPerson = people.map(person => Math.ceil(person.died / 100));
```
Anastasiia-Svintsova, line 18: Remove all comments, please

src/averageAges.js
```diff
@@ -15,6 +15,23 @@
...
+
+  const centuryOrNot = century ? filtredPersonByCentury : malePerson;
+
+  personAge = centuryOrNot.map(person => person.died - person.born);
+
+  const sumAge = personAge.reduce((sum, age) => sum + age, 0);
+
+  const averageAge = sumAge / personAge.length;
+
+  return +averageAge.toFixed(2);
```
Anastasiia-Svintsova, line 34: Create a method to calculate the average age. You repeat the same logic not once

src/averageAges.js
```diff
@@ -15,6 +15,23 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  // const centuryOfPerson = people.map(person => Math.ceil(person.died / 100));
+
+  const malePerson = people.filter(person => person.sex === 'm');
+
+  let personAge = 0;
```
Anastasiia-Svintsova, line 22: Why do you need to create this variable here?

src/averageAges.js
```diff
@@ -37,7 +54,22 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let ageWomen = 0;
```
Anastasiia-Svintsova, line 57: Same
Why do you need it with a 0 value?

Anastasiia-Svintsova (changes requested):

src/averageAges.js
```diff
@@ -15,7 +15,19 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
+  const malePerson = people.filter(person => person.sex === 'm');
+
+  const filtredPersonByCentury = malePerson.filter(
+    person => Math.ceil(person.died / 100) === century);
+
+  const centuryOrNot = century ? filtredPersonByCentury : malePerson;
```
sTorba24, line 23:
```suggestion
  const men = people.filter({ sex, died } => sex === 'm' && (
    century
      ? Math.ceil(died / 100) === century
      : true
  ));
```

src/averageAges.js
```diff
@@ -15,7 +15,19 @@
...
+
+  const personAge = centuryOrNot.map(person => person.died - person.born);
+
+  const averageAge = personAge.reduce(
+    (sum, age) => sum + age) / personAge.length;
```
sTorba24, line 28:
```suggestion
  const averageAge = personAge.reduce(
    (sum, { died, born }) => sum + died - born) / personAge.length;
```

src/averageAges.js
```diff
@@ -55,7 +78,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const sons = children.filter(child => child.sex === 'm');
+
+  const allChildrenOrSons = onlyWithSon ? sons : children;
+
+  const res = allChildrenOrSons.map(child => {
+    const foundMom = mothers.find(mom => mom.name === child.mother);
+
+    return foundMom ? child.born - foundMom.born : null;
+  }).filter(val => val !== null);
```
sTorba24, line 97:
```suggestion
  }).filter(val => val);
```

src/averageAges.js
```diff
@@ -55,7 +78,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // write code here
+  const femaleWomen = people.filter(person => person.sex === 'f');
+
+  const mothers = femaleWomen.filter(
+    women => people.some(person => person.mother === women.name));
+
+  const children = people.filter(
+    person => mothers.some(mother => person.mother === mother.name));
+
+  const sons = children.filter(child => child.sex === 'm');
```
sTorba24, line 89: you can use one do it in one variable
like above
if with son check also sex in filter
then you can get average ages => use `map` for children and save child.born - people.find(here find mother of this child).born
Reduce it

sTorba24 (changes requested):
Good work! Lets make it simplier :)

src/averageAges.js
```diff
@@ -37,7 +47,17 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const female = people.filter(({ sex }) => sex === 'f');
+
+  const withChild = withChildren
+    ? female.filter(
+      ({ name }) => people.some(({ mother }) => mother === name))
+    : female;
+
+  const averageAge = withChild.reduce(
```
agniya-i, line 57: It would be great to create separate method to calculate average age as mentioned in comments above to not repeat this code

agniya-i (changes requested):

src/averageAges.js
```diff
@@ -15,7 +15,17 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
+  const men = people.filter(({ sex, died }) => sex === 'm' && (
+    century
+      ? Math.ceil(died / 100) === century
+      : true
+  )
+  );
```
polosanya, line 23:
```suggestion
  ));
```

src/averageAges.js
```diff
@@ -37,7 +47,15 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const female = withChildren
+    ? people.filter(({ sex }) => sex === 'f').filter(
+      ({ name }) => people.some(({ mother }) => mother === name))
+    : people.filter(({ sex }) => sex === 'f');
```
polosanya, line 53: Do you really need to call filter method three times? 
You can call it once like in calculateMenAverageAge function

polosanya (changes requested):
The previous comment is not fixed. Create a separate function helper for calculating the average age
image
Feel free to ask for help in fe_chat if you have any difficulties)

vlad-demchuk (approved):

PR #1827, opened 2023-03-14
solution

Dmytro-Kretov:


TarasHoliuk (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,12 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menArr = people.filter(x => x.sex === 'm'
```
DarkMistyRoom, line 18: use more informative names
```suggestion
  const menArr = people.filter(person => person.sex === 'm'
```

DarkMistyRoom: fix it everywhere

src/averageAges.js
```diff
@@ -15,11 +15,12 @@
...
+    && (!century || Math.ceil(x.died / 100) === century));
+
+  const ages = menArr.reduce((acc, x) => acc + x.died - x.born, 0);
+
+  return ages / menArr.length;
```
DarkMistyRoom, line 23: you duplicate getting average ages logic in the next function. move it to a separate service function and reuse

DarkMistyRoom (changes requested):

qoa11a (approved):

PR #1832, opened 2023-03-19
averageAge was created

Pavlo-Petrashevskyi:


src/averageAges.js
```diff
@@ -37,7 +41,15 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothers = people.filter(mother =>
+    withChildren
+      ? mother.sex === 'f'
+      && people.filter(child => child.mother === mother.name).length > 0
+      : mother.sex === 'f');
+
+  return +(mothers.map(({ born, died }) => died - born)
```
TarasHoliuk, line 50: you are doing almost the same 3 times, so would be nice to move it to a separate function

src/averageAges.js
```diff
@@ -37,7 +41,15 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothers = people.filter(mother =>
+    withChildren
+      ? mother.sex === 'f'
+      && people.filter(child => child.mother === mother.name).length > 0
```
TarasHoliuk, line 47:
```suggestion
      && people.some(child => child.mother === mother.name)
```

TarasHoliuk (changes requested):
GJ :+1: 
I've added 2 small comments

src/averageAges.js
Pavlo-Petrashevskyi: Done

src/averageAges.js
SanyaBratashchuk: Fix it

src/averageAges.js
```diff
@@ -37,7 +52,13 @@ function calculateMenAverageAge(people, century) {
...
```
SanyaBratashchuk, line 58:
```suggestion
      && people.some(child => child.mother === mother.name)
```

src/averageAges.js
```diff
@@ -14,12 +14,27 @@
  *
  * @return {number}
  */
+
+const findAverage = function(arr, special, people) {
```
SanyaBratashchuk, line 18: What doest `special` mean?

SanyaBratashchuk: It is not the best idea to try to combine this logic in this way

SanyaBratashchuk (changes requested):

src/averageAges.js
Pavlo-Petrashevskyi: I guess you are right

src/averageAges.js
Pavlo-Petrashevskyi: done

src/averageAges.js
Pavlo-Petrashevskyi: I`ve redone function findAverage to a simplier for the first two functions, will it be alright?

src/averageAges.js
```diff
@@ -14,12 +14,21 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(man =>
+    century
+      ? man.sex === 'm'
+      && Math.ceil(man.died / 100) === century
+      : man.sex === 'm');
```
DarkMistyRoom, line 29:
```suggestion
  const men = people.filter(man => century
    ? man.sex === 'm'
      && Math.ceil(man.died / 100) === century
    : man.sex === 'm'
  );
```

src/averageAges.js
```diff
@@ -55,7 +70,16 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const childs = people.filter(child =>
+    onlyWithSon
+      ? child.sex === 'm' && people.find(mother => mother.name === child.mother)
+      : people.find(mother => mother.name === child.mother));
```
DarkMistyRoom, line 76: also, we can use method `some `here

DarkMistyRoom (approved):
Good job!

PR #1837, opened 2023-03-21
Solution

Irina-Kulish:


src/averageAges.js
```diff
@@ -15,13 +15,22 @@
...
+  return total / result.length;
+}
+// write code here
+// learn how to use array methods like .filter .map .some .every .find .reduce
+// avoid using loop and forEach
+// replace `if ()` statement with &&, || or ?:
+// without nesting
 /**
  * Implement calculateWomenAverageAge function
  *
```
tetiana-ivanchenko, line 36: don't left comments

src/averageAges.js
```diff
@@ -15,13 +15,22 @@
...
+  const result = people.filter(person => !century
+    ? person.sex === 'm'
+    : Math.ceil(person.died / 100) === century && person.sex === 'm'
+  );
+
+  const total = result.reduce((sum, person) =>
+    sum + (person.died - person.born),
+  0);
 
+  return total / result.length;
```
tetiana-ivanchenko, line 27: move it to the separate function, and use it here and in the next function

tetiana-ivanchenko (changes requested):

src/averageAges.js
```diff
@@ -26,25 +12,6 @@ function calculateMenAverageAge(people, century) {
...
- * specified then function calculates average age only for women with children
- *
- * Hint: To check if a woman has children you should find someone who mention
- * her as mother.
- *
- * @param {object[]} people
- * @param {boolean} withChildren - optional
- *
- * @return {number}
- */
```
TarasHoliuk, line 38: that's not the ordinary comment - it's [JS Doc](https://jsdoc.app/)
So in the future, you may leave JS Docs

src/averageAges.js
TarasHoliuk: you are doing almost the same three times (but in the third function you should prepare data before reducing) - so better to move it to a separate function as suggested

I'll approve, but please, in the future take into account all comments (if something is unclear, etc - feel free to ask in chat)

TarasHoliuk (approved):

PR #1838, opened 2023-03-21
solution

daryna-hnidash:


src/averageAges.js
```diff
@@ -1,61 +1,43 @@
...
-  // without nesting
+  const man
+    = century
+      ? people.filter(person => person.sex === 'm' && Math.ceil(person.died / 100) === century)
+      : people.filter(person => person.sex === 'm');
+
+  const ages = man.map(person => person.died - person.born);
+  const total = ages.reduce((sum, i) => sum + i, 0);
+
+  return total / ages.length;
```
DarkMistyRoom, line 13: would be good to move this logic to a function and reuse

DarkMistyRoom (changes requested):
Great job! Just let's optimize the code a bit

nazarmatsevych (approved):

PR #1841, opened 2023-03-21
average-ages

OlegRak86:


src/averageAges.js
```diff
@@ -14,12 +14,18 @@
  *
  * @return {number}
  */
-function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+const result = arr => Math.round((arr.reduce((prev, item) => prev
```
slavanosatskyi, line 17: variable names are too abstract. This is a function, so it should be a verb, not a noun - for example calculateAverageAge, also arr can be improved - choose something more specific to the task like ages or years

Check other places and try to improve variable namings. You can read as a reference this page https://betterprogramming.pub/clean-code-naming-b90740cbae12

src/averageAges.js
```diff
@@ -14,30 +15,44 @@
  *
  * @return {number}
  */
-function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+const averageAges = arrAge => Math.round((arrAge.reduce((prev, item) => prev
```
IvanFesenko, line 18:
```suggestion
const getAverageAge = arrAge => Math.round((arrAge.reduce((prev, item) => prev
```

IvanFesenko (approved):

PR #1860, opened 2023-03-23
solution

Anastasia2403:


src/averageAges.js
MylaCrystal, line 1: good job!

src/averageAges.js
```diff
@@ -55,7 +65,20 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  return onlyWithSon
+    ? sonsList.reduce((acc, child) => (
+      acc + child.born - mothersList.find(mother => (
+        mother.name === child.mother)).born
+    ), 0) / sonsList.length
+    : childrenList.reduce((acc, child) => (
+      acc + child.born - mothersList.find(mother => (
+        mother.name === child.mother)).born
+    ), 0) / childrenList.length;
```
liliia-livandovskaya, line 81: you can add a variable where you would define which list to use (childrenList or sonsList) and then add this variable to the `reduce` to make the code simpler

liliia-livandovskaya (changes requested):
GJ! Lets improve your code

liliia-livandovskaya (approved):

PR #1876, opened 2023-03-23
Develop

iHateAntichrist:


src/averageAges.js
```diff
@@ -37,7 +40,14 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const woman = withChildren
+    ? people.filter(person => (person.sex === 'f'
+    && people.some(child => child.mother === person.name)))
+    : people.filter(person => person.sex === 'f');
+
+  const age = woman.reduce((sum, pers) => (sum + (pers.died - pers.born)), 0);
+
+  return age / woman.length;
```
VladislavaT, line 50: I think u can create function to calculate average age and use it for both function

VladislavaT (approved):

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = century
+    ? people.filter(person => (person.sex === 'm'
+    && Math.ceil(person.died / 100) === century))
+    : people.filter(person => person.sex === 'm');
+
+  const age = men.reduce((sum, pers) => (sum + (pers.died - pers.born)), 0);
+
+  return age / men.length;
```
SerhiiKirik, line 25: create a separate func for this & reuse in all cases

src/averageAges.js
SerhiiKirik: Agree

SerhiiKirik (changes requested):
Well done 🚀 
Let's improve your code 🧑‍💻

SerhiiKirik (changes requested):
Fix comments 🧑‍💻

PR #1882, opened 2023-03-23
Develop

IvanTkachenko3301:


SerhiiKirik (changes requested):
Well done 🔥 
Let's improve your code 🧑‍💻

src/averageAges.js
```diff
@@ -55,7 +69,21 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(child =>
+    child.mother && people.some(parent => parent.name === child.mother));
```
ivadimko, line 73: You can simplify the children filtering by checking if onlyWithSon is true and child.sex === 'm' in the same filter function.

src/averageAges.js
```diff
@@ -37,7 +41,17 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(woman => woman.sex === 'f');
+
```
ivadimko, line 45: Consider using a single filter for women with or without children, instead of filtering twice.

jstmpelowycz (approved):

PR #1883, opened 2023-03-24
solution

yapelenua:


src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm'
+    && (!century || Math.ceil(person.died / 100) === century));
+
+  const totalAge = men.reduce(
+    (acc, person) => acc + (person.died - person.born), 0
+  );
```
SerhiiKirik, line 23: Move to the separate function & reuse it in all cases

src/averageAges.js
```diff
@@ -55,7 +68,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const peopleWithMother = people.filter(
+    child => child.mother && people.find(
+      parent => parent.name === child.mother));
+
+  const filteredPeople = onlyWithSon ? peopleWithMother.filter(
+    child => child.sex === 'm') : peopleWithMother;
+
+  const ageDifferences = filteredPeople.map(
+    person => person.born - people.find(
+      parent => parent.name === person.mother).born);
```
SerhiiKirik, line 80: if you do this with find - you can reuse it if you move this to the separate func
```suggestion
  const peopleWithMother = people.filter(
    child => child.mother && getMother(people, child.name));

  const filteredPeople = onlyWithSon ? peopleWithMother.filter(
    child => child.sex === 'm') : peopleWithMother;

  const ageDifferences = filteredPeople.map(
    person => person.born - getMother(people, person.name).born);
```

SerhiiKirik (changes requested):
Well done 🚀 
Let's make some improvements 🧑‍💻

PR #1888, opened 2023-03-24
Solution

ArtemVelychko:


src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const man = people.filter(person => century
+    ? person.sex === 'm' && century === Math.ceil(person.died / 100)
+    : person.sex === 'm'
+  );
```
anstsot, line 21: you can use 2 filters in a row 
```suggestion
  const man = people
    .filter(person => person.sex === 'm')
    .filter(person => century ? century === Math.ceil(person.died / 100) : true);
```

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
+  const manAge = man.map(person => person.died - person.born);
```
anstsot, line 22: you're using the same logic in 3 functions. you can create separate function for it

anstsot (approved):
Good job, approving. But I left few comments fo improve your code if you want

PR #1892, opened 2023-03-25
$%^&*()

victor-buldenko:


src/averageAges.js
```diff
@@ -14,7 +14,28 @@
  *
  * @return {number}
  */
```
OleziO, line 16: Remove all comments from the code.

OleziO (approved):

PR #1895, opened 2023-03-26
average ages solution

TolikSerhiychuk:


package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```
SerhiiKirik, line 19: In the future - Do not add extra files to PR
To add the file to a commit use git add src/.

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const males = century
```
SerhiiKirik, line 18: males?)
men would me better)
```suggestion
  const men = century
```

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
+    ? people.filter(
+      (male) => male.sex === 'm' && century === Math.ceil(male.died / 100)
+    )
+    : people.filter((male) => male.sex === 'm');
+
+  const totalAge = males.reduce(
```
SerhiiKirik, line 24: move this to the separate function & reuse it in all cases

SerhiiKirik (changes requested):
Well done 🔥 
Let's make some improvements 🧑‍💻

package.json
TolikSerhiychuk: got it

src/averageAges.js
TolikSerhiychuk: ok)

jstmpelowycz (approved):

PR #1896, opened 2023-03-26
calculate average ages

ofenchuk:


src/averageAges.js
```diff
@@ -56,6 +85,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  );
+  let age = [];
+
+  kids.map(children => {
+    const mother = people.find(person => person.name === children.mother);
+    const result = mother ? children.born - mother.born : undefined;
+
+    age.push(result);
+  });
+  age = age.filter((a) => a !== undefined);
```
SerhiiKirik, line 100: Use chained methods instead of reassigning to let variable.

SerhiiKirik (approved):
Well done 🚀

PR #1900, opened 2023-03-28
Solution

NYasnaya:


src/averageAges.js
```diff
@@ -8,18 +8,25 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menArr = people.filter(person => century
+    ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+    : person.sex === 'm');
+
+  const averageMen = menArr.reduce((sum, man) =>
+    sum + (man.died - man.born), 0);
```
IvanFesenko, line 27: You can move it to separate function and reuse it in `calculateWomenAverageAge` function

IvanFesenko (approved):
Well done. Pay attention on my comment to avoid code duplication

src/averageAges.js
NYasnaya: ok, thank you!!

PR #1902, opened 2023-03-29
solution

DanielSorokowski:


src/averageAges.js
```diff
@@ -20,6 +20,14 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  let mens = people.filter(person => person.sex === 'm');
+
+  /* eslint-disable max-len */
+  mens = century ? mens.filter(men => Math.ceil(men.died / 100) === century) : mens;
```
solaryasha, line 26:
```suggestion
  const men = century 
   ? people.filter(men => Math.ceil(men.died / 100) === century)
   :  people.filter(person => person.sex === 'm');
```

src/averageAges.js
```diff
@@ -20,6 +20,14 @@ function calculateMenAverageAge(people, century) {
...
+
+  const yearsAlive = mens.map(men => men.died - men.born);
+
+  return yearsAlive.reduce((sum, year) => sum + year, 0) / yearsAlive.length;
```
solaryasha, line 30: Could you combine `map` and `reduce` ?

You could calculate the age difference inside reduce (`(sum, person) => sum + person.died - person.born `)

src/averageAges.js
```diff
@@ -37,7 +45,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => person.sex === 'f');
+  const allMothers = people.map(person => person.mother);
```
solaryasha, line 49:
```suggestion
  const motherNames = people.map(person => person.mother);
```

solaryasha: Also, you could use `Set` for it. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set 
Set accepts an array, you will get a more efficient algorithm because checking the value that exists inside the set is efficient than using `includes`  in array

solaryasha (changes requested):

src/averageAges.js
```diff
@@ -20,6 +20,13 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  /* eslint-disable max-len */
+  const mens = century
+    ? people.filter(person => person.sex === 'm' && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === 'm');
+
+  return mens.map(men => men.died - men.born).reduce((sum, year) => sum + year, 0) / mens.length;
```
solaryasha, line 29:
```suggestion
  return mens.reduce((sum, person) => sum + person.died - person.born, 0) / mens.length;
```

src/averageAges.js
```diff
@@ -37,7 +44,15 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const women = people.filter(person => person.sex === 'f');
+  const motherNames = new Set(people.map(person => person.mother));
+  const mothers = people.filter(person => motherNames.has(person.name));
+
+  const yearsAlive = withChildren
+    ? mothers.map(mother => mother.died - mother.born)
+    : women.map(woman => woman.died - woman.born);
+
+  return yearsAlive.reduce((sum, year) => sum + year, 0) / yearsAlive.length;
```
solaryasha, line 55: Let's do the same as in the previous comment

src/averageAges.js
```diff
@@ -55,7 +70,13 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = onlyWithSon
+    ? people.filter(child => child.sex === 'm' && people.some(mother => mother.name === child.mother))
+    : people.filter(child => people.some(mother => mother.name === child.mother));
+
+  const difference = children.map(child => child.born - people.find(mother => mother.name === child.mother).born);
```
solaryasha, line 77:
```suggestion
  const ageDifference = children.map(child => child.born - people.find(mother => mother.name === child.mother).born);
```

src/averageAges.js
```diff
@@ -55,7 +70,13 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = onlyWithSon
+    ? people.filter(child => child.sex === 'm' && people.some(mother => mother.name === child.mother))
+    : people.filter(child => people.some(mother => mother.name === child.mother));
```
solaryasha, line 75: You could also create a set with mother names. It will be more effecient

solaryasha (changes requested):

src/averageAges.js
```diff
@@ -37,7 +44,13 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => person.sex === 'f');
+  const motherNames = new Set(people.map(person => person.mother));
+  const mothers = people.filter(person => motherNames.has(person.name));
+
+  return withChildren
+    ? mothers.reduce((sum, person) => sum + person.died - person.born, 0) / mothers.length
+    : women.reduce((sum, person) => sum + person.died - person.born, 0) / women.length;
```
solaryasha, line 53: DRY
```suggestion
  const targetedWomen = withChildren
    ? mothers
    : women;
    
  return targetedWomen.reduce((sum, person) => sum + person.died - person.born, 0) / targetedWomen.length
```

solaryasha (approved):

PR #1908, opened 2023-03-30
solution

IamLoren:


src/averageAges.js
```diff
@@ -20,6 +20,15 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const men = century
+    ? people.filter(
+      person => person.sex === 'm' && Math.ceil(person.died / 100) === century
+    )
+    : people.filter(person => person.sex === 'm');
+
+  const sum = men.reduce((total, man) => total + (man.died - man.born), 0);
```
OlEzhA0, line 29: I see you reuse this logic several times, please add some util and reuse it

OlEzhA0 (changes requested):

src/averageAges.js
```diff
@@ -55,7 +82,32 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const arrOfDiff = [];
```
ivadimko, line 85: Replace the `arrOfDiff` array with a reduce function to calculate the sum of age differences in a more efficient way.

src/averageAges.js
```diff
@@ -55,7 +82,32 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const men = people.filter(person => person.sex === 'm');
+
+  if (onlyWithSon) {
```
ivadimko, line 88: Refactor the `if (onlyWithSon)` block and the `else` block to avoid code duplication.

src/averageAges.js
```diff
@@ -38,6 +47,24 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const womenWithChildren = people.filter(person => {
```
ivadimko, line 50: Use the `Array.prototype.some()` method to simplify the `womenWithChildren` filtering logic.

src/averageAges.js
ivadimko: consider using ternary operator instead of `if/else`

src/averageAges.js
```diff
@@ -55,7 +82,32 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const arrOfDiff = [];
+  const men = people.filter(person => person.sex === 'm');
+
+  if (onlyWithSon) {
+    people.forEach(person => {
+      for (const p of men) {
```
ivadimko, line 90: use array iteration methods instead of `for` loop. Check other places in code

mykhalenych (approved):

ivadimko (approved):
.

PR #1910, opened 2023-03-31
Solution

HannaVasylieva:


src/averageAges.js
```diff
@@ -15,11 +15,12 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm'
+    && (!century || Math.ceil(person.died / 100) === century));
+  const totalAge = men.reduce((total, person) =>
+    total + (person.died - person.born), 0);
```
Anastasiia-Svintsova, line 21: Create a separate function for calculating average age. You use the same logic several times

src/averageAges.js
```diff
@@ -37,7 +38,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = (withChildren)
+    ? people.filter(female => female.sex === 'f'
+      && people.some(child => child.mother === female.name))
+    : people.filter(female => female.sex === 'f');
```
Anastasiia-Svintsova, line 44: One of possible ways 
```suggestion
  const women = people.filter(female => withChildren 
    ? female.sex === 'f' && people.some(child => child.mother === female.name)
    : female.sex === 'f'
  );
```

Anastasiia-Svintsova (changes requested):

src/averageAges.js
HannaVasylieva: I'm sorry, don't understand, what I need to do exactly? Do I need to create a function that I can use 3 times in every task?

sTorba24: Yes, as you can see you do the same in reduce, so you can create a function where you pass the same data structure and calculate average using function

sTorba24 (approved):
Approved! I left you a comment, you can improve it at your discretion :)

src/averageAges.js
```diff
@@ -37,7 +38,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(female => withChildren
+    ? female.sex === 'f' && people.some(child => child.mother === female.name)
+    : female.sex === 'f'
+  );
+
+  const averageAge = (women
```
ivadimko, line 46: Consider renaming `averageAge` to `totalAge` for consistency with the `calculateMenAverageAge` function.

and then return `totalAge / women.length`
it will be more readable this way

src/averageAges.js
```diff
@@ -37,7 +38,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(female => withChildren
```
ivadimko, line 41: It's recommended to use a ternary operator instead of `&&` to make the code more readable. Replace with `female.sex === 'f' && (withChildren ? people.some(child => child.mother === female.name) : true)`

src/averageAges.js
```diff
@@ -55,7 +65,19 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const childs = people.filter(
```
ivadimko, line 68: children is a correct plural form for child

PR #1921, opened 2023-04-05
solution

Ilya-Osoben:


src/averageAges.js
```diff
@@ -1,61 +1,100 @@
...
- * for sons and their mothers.
- *
- * @param {object[]} people
- * @param {boolean} onlyWithSon - optional
- *
- * @return {number}
- */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const arrNamePeople = people.map(item => {
```
ivadimko, line 60: The `arrNamePeople` array is not necessary. You can use the `some` method to check if a person has a mother in the array.

src/averageAges.js
```diff
@@ -1,61 +1,100 @@
...
- */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const arrayMen = people.filter(item => item.sex === 'm');
+
+  const arrayAge = arrayMen.map(item => {
```
ivadimko, line 6: Instead of creating a separate array for ages, you can calculate the sum of ages directly in the filter. Use the `reduce` method to replace the `filter` and `map` methods.

src/averageAges.js
```diff
@@ -1,61 +1,100 @@
...
+    });
+    const age = item.born - itemMother[0].born;
+
+    return age;
+  });
+
+  const sumAgeDiff = arrAgeDiff.reduce((accum, item) => accum + item, 0);
+  const averageAgeDiff = sumAgeDiff / arrAgeDiff.length;
+
+  const arrSonWithMom = arrPersWithMom.filter(item => item.sex === 'm');
```
ivadimko, line 81: Instead of filtering the sons separately, you can include the condition for sons in the initial filter for people with mothers. This will simplify the code and reduce redundancy.

sTorba24 (changes requested):
fix comments above

PR #1927, opened 2023-04-06
solution

MikiOOOO:


src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let men = people.filter(person => person.sex === 'm');
+
+  if (century) {
+    men = men.filter(person => Math.ceil(person.died / 100) === century);
+  }
```
d-wasilewski, line 22: use ternary operator here

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
+
+  const ageSum = men.reduce((prev, curr) => prev + (curr.died - curr.born), 0);
```
d-wasilewski, line 24: change the naming from prev, curr to something more explanatory

src/averageAges.js
```diff
@@ -37,7 +41,19 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let women = people.filter(person => person.sex === 'f');
+
+  if (withChildren) {
```
d-wasilewski, line 46: again, use ternary here

src/averageAges.js
```diff
@@ -55,7 +71,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  let ageDifferenceCount = 0;
+
+  childrenWithMothers.forEach(child => {
+    people.forEach(person => {
+      if (person.name === child.mother) {
+        ageSum += child.born - person.born;
+        ageDifferenceCount++;
+      }
+    });
+  });
```
d-wasilewski, line 91: can you do it without loops nesting? maybe split it into 2 blocks? Also, try using `reduce` here too

d-wasilewski (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => (
+    century ? person.sex === 'm'
+    && Math.ceil(person.died / 100) === century : person.sex === 'm'
+  ));
+
+  const ageSum = men.reduce((prevMan, currMan) =>
+    prevMan + (currMan.died - currMan.born), 0);
```
d-wasilewski, line 24: You don't need to use curr, it's self explanatory that its a current value
```suggestion
  const ageSum = men.reduce((sum, man) =>
    prevMan + (currMan.died - currMan.born), 0);
```

src/averageAges.js
```diff
@@ -55,7 +66,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const ageDifferences = [];
+
+  children.forEach(child => {
+    const searchedMother = mothers.find(mother =>
+      mother.name === child.mother);
+
+    if (searchedMother) {
+      ageDifferences.push(child.born - searchedMother.born);
+    }
+  });
```
d-wasilewski, line 85: use map if you want to return an array
```suggestion
  const ageDifferences = children.map(child => {
```

d-wasilewski (changes requested):

src/averageAges.js
```diff
@@ -55,7 +66,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ageDifferences = children.map(child => {
+    const searchedMother = mothers.find(mother =>
+      mother.name === child.mother);
+
+    if (searchedMother) {
+      return child.born - searchedMother.born;
+    } else {
+      return 0;
+    }
```
d-wasilewski, line 84: ternary operator

src/averageAges.js
```diff
@@ -55,7 +66,28 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const mothers = people.filter(person => people.some(human =>
+    human.mother === person.name));
+
+  const children = onlyWithSon ? people.filter(person =>
+    person.mother !== null && person.sex === 'm')
```
d-wasilewski, line 73: I think this check for null is redundant, bacause you check later if `mother.name === child.mother`. You can try to do this without this check

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1928, opened 2023-04-06
Develop

Vitalii-Maistrenko:
https://github.com/Vitalii-Maistrenko/js_task-guideline/blob/master/README.md)

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -18,44 +18,43 @@ function calculateMenAverageAge(people, century) {
   // write code here
   // learn how to use array methods like .filter .map .some .every .find .reduce
   // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
+  // replace if () statement with &&, || or ?:
   // without nesting
+
+  const men = people.filter(person => century
+    ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+    : person.sex === 'm'
+  );
+
+  return men.map(person => person.died - person.born)
+    .reduce((a, b) => a + b) / men.length;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/averageAges.js
```diff
@@ -18,44 +18,43 @@ function calculateMenAverageAge(people, century) {
...
+  // replace if () statement with &&, || or ?:
   // without nesting
+
+  const men = people.filter(person => century
+    ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+    : person.sex === 'm'
+  );
+
+  return men.map(person => person.died - person.born)
+    .reduce((a, b) => a + b) / men.length;
```
IvanFesenko, line 30: Map is redudant here. Use only `reduce`.
Also you have similar code in `calculateWomenAverageAge`
So create separate function to `getAvarageAge` and use it here and `calculateWomenAverageAge`

src/averageAges.js
```diff
@@ -18,44 +18,43 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  const children = people.filter(person => onlyWithSon
+    ? people.find(woman => woman.name === person.mother) && person.sex === 'm'
+    : people.find(woman => woman.name === person.mother));
+
+  return children.map(person =>
+    person.born - people.find(mother => person.mother === mother.name).born)
+    .reduce((a, b) => a + b, 0) / children.length;
```
IvanFesenko, line 57: Same here. You can do it without `map`

IvanFesenko (changes requested):

anstsot (approved):
Great job, approved!

PR #1930, opened 2023-04-06
Solution

mateusz-michalowski:


danielpopek94 (approved):

src/averageAges.js
```diff
@@ -20,24 +20,49 @@ function calculateMenAverageAge(people, century) {
...
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  let malesAverageAge = people.filter((person) => person.sex === 'm');
+
+  malesAverageAge = century !== undefined
+    ? malesAverageAge.filter(
+      (person) => Math.ceil(person.died / 100) === century
+    )
+    : malesAverageAge;
```
d-wasilewski, line 30: use `const` and do it in 1 condition.
use ternary, but don't check if century is undefined, go with `century ? a : b`

src/averageAges.js
```diff
@@ -20,24 +20,49 @@ function calculateMenAverageAge(people, century) {
...
  * her as mother.
  *
  * @param {object[]} people
- * @param {boolean} withChildren - optional
+ * @param {boolean} withkids - optional
  *
  * @return {number}
  */
-function calculateWomenAverageAge(people, withChildren) {
+function calculateWomenAverageAge(people, withkids) {
```
d-wasilewski, line 52: what's wrong with `children`? 😛

src/averageAges.js
```diff
@@ -20,24 +20,49 @@ function calculateMenAverageAge(people, century) {
...
   // write code here
+  let femalesAverageAge = people.filter((person) => person.sex === 'f');
+  const notMotherless = people.filter((person) => person.mother !== null);
+
+  femalesAverageAge = withkids === true
```
d-wasilewski, line 57: same, you don't have to write it like `variable === true`, just `variable` will do the same **in this case**

src/averageAges.js
```diff
@@ -20,24 +20,49 @@ function calculateMenAverageAge(people, century) {
...
  * @param {object[]} people
- * @param {boolean} withChildren - optional
+ * @param {boolean} withkids - optional
  *
  * @return {number}
  */
-function calculateWomenAverageAge(people, withChildren) {
+function calculateWomenAverageAge(people, withkids) {
   // write code here
+  let femalesAverageAge = people.filter((person) => person.sex === 'f');
```
d-wasilewski, line 54: const

src/averageAges.js
```diff
@@ -56,6 +81,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const kids = onlyWithSon
+    ? people.filter(child => child.sex === 'm'
+    && mothers.has(child.mother))
+    : people.filter(child => mothers.has(child.mother));
+
+  const difference = kids.map(
+    child => child.born - people.find(
+      mother => mother.name === child.mother).born
+  );
```
d-wasilewski, line 94: If you map over kids array, the iterator variable should also be `kid` (refering to the array name), or name the array children
```suggestion
  const ageDifference = kids.map(
    kid => kid.born - people.find(
      mother => mother.name === kid.mother).born
  );
```

d-wasilewski (changes requested):

Ozwiena (approved):

src/averageAges.js
```diff
@@ -38,6 +50,16 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const femalesAverageAge = withChildren
+    ? people.filter((person) => person.sex === 'f').filter(
+      (person) => people.filter((mum) => mum.mother !== null)
```
d-wasilewski, line 55: do you need this check to be so complex? try to do it without comparing to !== null

d-wasilewski (changes requested):

bjeziorek (approved):
For me this solution looks fine :).

kasiakubisiak (approved):

d-wasilewski (approved):

PR #1931, opened 2023-04-06
solution

danielpopek94:


YBM3 (approved):

src/averageAges.js
```diff
@@ -14,12 +14,22 @@
  *
  * @return {number}
  */
+
+const calculateAverage = (arr) => arr.reduce((a, b) => a + b, 0) / arr.length;
+const calculateLiveLength = (people) => (
+  people.map(person => person.died - person.born)
+);
+
```
d-wasilewski, line 22: 🚀

src/averageAges.js
```diff
@@ -55,7 +73,17 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(child => (
+    child.mother && (!onlyWithSon || child.sex === 'm')
+  ));
+
+  const ageDif = children.reduce((current, child) => {
```
d-wasilewski, line 80:
```suggestion
  const ageDif = children.reduce((accumulator (or ages), child) => {
```

src/averageAges.js
```diff
@@ -14,12 +14,22 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const diedCentury = (person) => (
+    !century || Math.ceil(person.died / 100) === century
+  );
+
+  const men = people.filter((person) => (
+    person.sex === 'm' && diedCentury(person)
+  ));
```
d-wasilewski, line 30: it's interesting approach 🚀
However I think it would be more readable if you used a ternary operator here instead to perfom this check, in this wait it's not so obvious at the first glance what it does and verify the conditions

d-wasilewski (changes requested):
Great job! 👏

src/averageAges.js
```diff
@@ -37,7 +49,17 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const wasMother = (woman) => (
+    withChildren
+      ? people.some(child => child.mother === woman.name)
+      : true
+  );
```
d-wasilewski, line 56: if you have to return true in ternary operator you should use `||` or `&&`. I pointed this out with ternary because I though you will use it sooner (instead of woman.sex check). This solution is good, just change it to `||` or `&&`.
Ex:
`!withChildren || people.some(child => child.mother === woman.name)`

src/averageAges.js
```diff
@@ -55,7 +77,17 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(child => (
+    child.mother && (!onlyWithSon || child.sex === 'm')
+  ));
+
+  const ageDif = children.reduce((ages, child) => {
```
d-wasilewski, line 84:
```suggestion
  const ageDifference = children.reduce((ages, child) => {
```

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1935, opened 2023-04-08
Solution

Alessandranitti15:


src/averageAges.js
```diff
@@ -15,11 +15,22 @@
...
+      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+      : person.sex === 'm';
+  }
+
+  function averageAge(prev, person) {
+    const age = person.died - person.born;
+
+    return prev + age;
+  }
+
```
agniya-i, line 32: you have similar logic in several functions for calculating average age, move this function outside and reuse

agniya-i (changes requested):

agniya-i (approved):

PR #1940, opened 2023-04-10
AverageAge function were created

MariaShmakova2014:


src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
+  const men = people.filter(person => person.sex === 'm');
+
+  let ages = men.map(person => person.died - person.born);
+
+  const centuryPersons = men.filter(person =>
+    Math.ceil(person.died / 100) === century);
+
+  if (century) {
+    ages = centuryPersons.map(person => person.died - person.born);
+  }
```
VitaliyBondarenko1982, line 27: you can avoid map iteration, if statement... 
```suggestion
  const men = people
    .filter(({ sex, died }) => century
      ? Math.ceil(died / 100) === century && sex === 'm' : sex === 'm');
```

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
+
+  const count = ages.reduce((sum, age) => sum + age, 0);
+
+  return count / ages.length;
```
VitaliyBondarenko1982, line 31: and here work with men
```suggestion
  const count = men.reduce((sum, { born, died }) => sum + died - born, 0);

  return count / men.length;
```

src/averageAges.js
```diff
@@ -37,7 +46,21 @@ function calculateMenAverageAge(people, century) {
...
+
+  let ages = wemen.map(person => person.died - person.born);
+
+  if (withChildren) {
+    const womenWithChildren = wemen.filter(person =>
+      people.some(p => p.mother === person.name)
+    );
+
+    ages = womenWithChildren.map(person => person.died - person.born);
+  }
```
VitaliyBondarenko1982, line 59: try to avoid map iteration and if as in calculateMenAverageAge

VitaliyBondarenko1982 (changes requested):

PR #1951, opened 2023-04-12
Average-Ages-1

Soi4An:


src/averageAges.js
```diff
@@ -36,8 +50,30 @@ function calculateMenAverageAge(people, century) {
...
+  const womensWithChild = [];
+
+  if (withChildren) {
+    for (const women of womens) {
+      for (const { mother } of people) {
+        if (women.name === mother) {
+          womensWithChild.push(women);
+          break;
+        }
+      }
```
vlad-demchuk, line 65: use only iteration methods in this tasks

src/averageAges.js
```diff
@@ -36,8 +50,30 @@ function calculateMenAverageAge(people, century) {
...
+    }
+
+    womens = womensWithChild;
+  }
+
+  womens.map(women => womensAges.push(women.died - women.born));
+
+  const sumWomenAge = womensAges.reduce((sum, age) => sum + age, 0);
+  const averageWomenAge = sumWomenAge / womens.length;
```
vlad-demchuk, line 74: DRY, create a separate helper function to calculate average age

vlad-demchuk (changes requested):

src/averageAges.js
```diff
@@ -54,10 +74,36 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  (onlyWithSon)
+    ? peopleWithMom = people.filter((person, i, arr) =>
+      person.mother && arr.find(arrPerson => arrPerson.name === person.mother)
+        && person.sex === 'm')
+    : peopleWithMom = people.filter((person, i, arr) =>
+      person.mother && arr.find(arrPerson => arrPerson.name === person.mother)
+    );
+
+  peopleWithMom.map(person => {
+    agesYoungMoms.push(person.born
```
nazarmatsevych, line 93: map method return new array,  so we can write the result of the map method into a variable

nazarmatsevych (approved):
Good work, 
just small suggestion

PR #1958, opened 2023-04-12
AverageAge

JaneShavrukova:


src/averageAges.js
```diff
@@ -1,61 +1,43 @@
...
- *
- * @return {number}
- */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
```
tetiana-ivanchenko, line 4: combine two filters into one that will depends on the century

src/averageAges.js
```diff
@@ -1,61 +1,43 @@
...
+
+  const menAges = (century)
+    ? men
+      .filter(person => Math.ceil(person.died / 100) === century)
+      .map(person => person.died - person.born)
+    : men.map((person) => person.died - person.born);
+
+  return (menAges.reduce((sum, age) => sum + age)) / menAges.length;
```
tetiana-ivanchenko, line 12: create separate function that will take people object and will return average age, to use it here, and in the second function

tetiana-ivanchenko (changes requested):

anstsot (approved):
Great job, approved

PR #1959, opened 2023-04-13
avarage Ages

theFallcon:


src/averageAges.js
```diff
@@ -54,8 +67,26 @@ function calculateWomenAverageAge(people, withChildren) {
  *
  * @return {number}
  */
+
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  const kids = (onlyWithSon
+    ? [...people].filter(person => person.sex === 'm')
```
witflash, line 74: you can use `people.filter` without spread operator because `filter` method returns a new array.

witflash (approved):
Well done 👍

PR #1960, opened 2023-04-13
Solution

Sergey8793:


src/averageAges.js
```diff
@@ -37,7 +39,14 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => person.sex === 'f');
+  const womenChildren = withChildren
+    ? women.filter(woman => people.some(person => person.mother === woman.name))
+    : women;
+  const womenAges = womenChildren.map(person => person.died - person.born);
+  const sumAges = womenAges.reduce((sum, age) => sum + age, 0);
+
+  return sumAges / womenAges.length;
```
vlad-demchuk, line 49: create a separate helper function to calculate average ages
you do the same three times

vlad-demchuk (changes requested):

src/averageAges.js
```diff
@@ -1,5 +1,12 @@
...
+function calculateAverageAge(people) {
+  return people
+    .reduce(
+      (accumulator, person) => accumulator + (person.died - person.born), 0
+    ) / people.length;
+}
+
 /**
  * Implement calculateMenAverageAge function
  *
```
OleziO, line 12: Remove al comments.

OleziO (changes requested):

AlisaVlasova (approved):

PR #1962, opened 2023-04-14
solution

mariuszgit:


src/averageAges.js
```diff
@@ -55,7 +81,21 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const filteredTab = (onlyWithSon)
```
DorotaLeniecDev, line 84: Find better name so your code is easier to read. Now it is hard to guess what is in filterTab const

src/averageAges.js
```diff
@@ -55,7 +81,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    ? people.filter(filterMan).filter(filterChilds)
```
DorotaLeniecDev, line 85: Try to filter the array one time

src/averageAges.js
```diff
@@ -55,7 +81,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    : people.filter(filterChilds);
+
+  function filterMan(person) {
```
DorotaLeniecDev, line 88: I see this function is the same as in **calculateMenAverageAge**. Instead of creating this function 2 times you can create it one time (in global scope) and use it inside where you need it. Check if you can extract some other helper functions that appears one than one time.
```suggestion
  function isMan(person) {
```

src/averageAges.js
```diff
@@ -55,7 +81,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    return person.sex === 'm';
+  }
+
+  function filterChilds(person) {
+    return people.some(el => el.name === person.mother);
+  };
+
+  return filteredTab.reduce((prev, next) => {
+    return prev + (next.born - people.find(el => el.name === next.mother).born);
+  }, 0) / filteredTab.length;
```
DorotaLeniecDev, line 98: Try to make this code more readible by introducing some better naming and code formating

DorotaLeniecDev (changes requested):
Your code can be improved. Check my comments :)

src/averageAges.js
```diff
@@ -36,8 +50,25 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const filteredTab = withChildren
+    ? people.filter(isMother)
+    : people.filter(isWoman);
+
+  function isWoman(person) {
+    return person.sex === 'f';
+  };
+
+  function isMother(person) {
```
DorotaLeniecDev, line 63: The `isMother` function should be moved outside of `calculateWomenAverageAge` to keep the main function cleaner, and to avoid redefining it on each function call. 😇

src/averageAges.js
```diff
@@ -14,12 +14,26 @@
...
-  // without nesting
+  const filteredTab = people
+    .filter(isMan)
+    .filter(haveDiedInCentury);
+
+  function haveDiedInCentury(person) {
+    return (century) ? (Math.ceil(person.died / 100)) === century : true;
+  }
+
+  function sumAllAges(prev, next) {
```
DorotaLeniecDev, line 32: The `sumAllAges` function should be moved outside of `calculateMenAverageAge` to keep the main function cleaner, and to avoid redefining it on each function call. 😇

src/averageAges.js
```diff
@@ -14,12 +14,26 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredTab = people
+    .filter(isMan)
+    .filter(haveDiedInCentury);
+
+  function haveDiedInCentury(person) {
```
DorotaLeniecDev, line 28: The `haveDiedInCentury` function should be moved outside of `calculateMenAverageAge` to keep the main function cleaner, and to avoid redefining it on each function call. 😇

src/averageAges.js
```diff
@@ -36,8 +50,25 @@ function calculateMenAverageAge(people, century) {
...
+
+  function isWoman(person) {
+    return person.sex === 'f';
+  };
+
+  function isMother(person) {
+    return people.some(el => el.mother === person.name);
+  };
+
+  function sumAllAges(prev, next) {
```
DorotaLeniecDev, line 67: The `sumAllAges` function should be moved outside of `calculateWomenAverageAge` to keep the main function cleaner, and to avoid redefining it on each function call. 😇

src/averageAges.js
```diff
@@ -36,8 +50,25 @@ function calculateMenAverageAge(people, century) {
...
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filteredTab = withChildren
+    ? people.filter(isMother)
+    : people.filter(isWoman);
+
+  function isWoman(person) {
```
DorotaLeniecDev, line 59: The `isWoman` function should be moved outside of `calculateWomenAverageAge` to keep the main function cleaner, and to avoid redefining it on each function call. 😇

src/averageAges.js
```diff
@@ -14,12 +14,26 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredTab = people
+    .filter(isMan)
+    .filter(haveDiedInCentury);
```
DorotaLeniecDev, line 26: It's better to use a single `filter` method instead of chaining two filters. You can combine the conditions using `&&` inside the `filter` callback. 🧐

src/averageAges.js
```diff
@@ -55,7 +86,18 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const filteredTab = (onlyWithSon)
+    ? people.filter(isMan).filter(isChild)
+    : people.filter(isChild);
+
+  function isChild(person) {
```
DorotaLeniecDev, line 93: The `isChild` function should be moved outside of `calculateAverageAgeDiff` to keep the main function cleaner, and to avoid redefining it on each function call. 😇

src/averageAges.js
DorotaLeniecDev, line 26: Try to avoid chaining filter() method it can cos you performance when dealing with big data. You can use one filter and combine the conditions. Also here you dont need to check the other condition if **country** is not specified.

src/averageAges.js
```diff
@@ -14,12 +14,26 @@
...
+  return person.sex === 'm';
+}
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredTab = people
```
DorotaLeniecDev, line 24: Name this cost more descriptively
```suggestion
  const men= people
```

src/averageAges.js
```diff
@@ -37,7 +47,23 @@ function calculateMenAverageAge(people, century) {
...
+  const filteredTab = (withChildren)
+    ? people.filter(filterMothers)
+    : people.filter(filterWoman);
+
+  function filterWoman(person) {
+    return person.sex === 'f';
+  };
+
+  function filterMothers(person) {
+    return people.some(el => el.mother === person.name);
```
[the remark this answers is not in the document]
DorotaLeniecDev: Use more descriptive name for **el**

src/averageAges.js
```diff
@@ -36,8 +50,25 @@ function calculateMenAverageAge(people, century) {
...
+    return person.sex === 'f';
+  };
+
+  function isMother(person) {
+    return people.some(el => el.mother === person.name);
+  };
+
+  function sumAllAges(prev, next) {
+    return prev + (next.died - next.born);
+  };
```
DorotaLeniecDev, line 69: You could create helper function for calculating average as you are using this in multiple places

src/averageAges.js
```diff
@@ -55,7 +86,18 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const filteredTab = (onlyWithSon)
+    ? people.filter(isMan).filter(isChild)
```
DorotaLeniecDev, line 90: Also here avoid chaining filter methods. Instead use one filter with multiple conditions

src/averageAges.js
DorotaLeniecDev:
```suggestion
  const children= (onlyWithSon)
```

src/averageAges.js
```diff
@@ -55,7 +86,18 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    : people.filter(isChild);
+
+  function isChild(person) {
+    return people.some(child => child.name === person.mother);
+  };
+
+  return filteredTab.reduce((prev, next) => {
+    return prev + (next.born - people
+      .find(person => person.name === next.mother).born);
+  }, 0) / filteredTab.length;
```
DorotaLeniecDev, line 100:
```suggestion
  return children.reduce((sum, child) => {
    const mother = people.find(person => person.name === child.mother);
    
    return sum+ (child.born - mother.born);
  }, 0) / children.length;
```

DorotaLeniecDev (changes requested):

DorotaLeniecDev (changes requested):
Two more comments :)

src/averageAges.js
```diff
@@ -36,8 +61,13 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const woman = withChildren
+    ? people.filter((person, index, arr) => isMother(person, index, arr))
```
Radoslaw-Czerniawski, line 67: `isMother` function accepts same arguments as filter callback so you could just pass the `isMother` similarly to the `isWoman`
```
people.filter(isMother)
```
instead of writing callback arrow function that passes same arguments to the `isMother`

src/averageAges.js
```diff
@@ -14,12 +14,37 @@
  *
  * @return {number}
  */
+
+// helper fn
+function isMan(person) {
```
Radoslaw-Czerniawski, line 19: You could declare some of these functions as arrow functions and do implicit returns example:
```
const isMan = (person) => person.sex === 'm'
```
and also destructure the person argument
```
const isMan = ({sex}) =>sex === 'm'
```
But know we can see that `isMan` doesn't actually need the whole person object it could just be
```
const isMan = (sex) => sex === 'm'
```
You just need to pass person.sex to the function call instead of the whole person object

Radoslaw-Czerniawski: You may do something similar for other functions

src/averageAges.js
```diff
@@ -14,12 +14,37 @@
...
+};
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people
+    .filter(person => isMan(person) && haveDiedInCentury(person, century));
```
Radoslaw-Czerniawski, line 45: You could make those function accept only parts of the person object and destructure the person object in the callback params like so
```
.filter({sex, died} => isMan(sex) && haveDiedInCentury(died, century));
```

Radoslaw-Czerniawski (changes requested):
Generally left some ideas how to make this code shorter / more readable, adjust according to those ideas if you agree with them, otherwise point out why not in rereview

PR #1965, opened 2023-04-17
Develop

AnastasiiaVorobets:


src/averageAges.js
```diff
@@ -55,7 +64,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    women.some(mother => mother.name === person.mother));
+
+  const filteredChildren = onlyWithSon
+    ? children.filter(child => child.sex === 'm')
+    : children;
+
+  const ageDifference = filteredChildren.map(person =>
+    person.born - people.find(mother => mother.name === person.mother).born);
+
+  const averageAge = ageDifference.reduce((sum, age) =>
```
vlad-demchuk, line 79: create a separate helper function to calculate average age
you do the same three times

vlad-demchuk (changes requested):
Looks good, just a small improvement

src/averageAges.js
```diff
@@ -1,5 +1,12 @@
...
 
+function calculateAverageAge(people) {
+  const totalAge = people
+    .reduce((sum, person) => sum + (person.died - person.born), 0);
+
+  return totalAge / people.length;
+}
+
 /**
  * Implement calculateMenAverageAge function
```
OleziO, line 11: Remove all comments from the code.

OleziO (changes requested):

nazarmatsevych (approved):
Well done 👍

PR #1968, opened 2023-04-19
first solution

spencerspe:


qoa11a (changes requested):
Please, read the condition carefully. And apply all the changes I've suggested.

Anastasiia-Svintsova (changes requested):
Please, fix the comments and only then request the review again

src/averageAges.js
```diff
@@ -20,6 +20,21 @@ function calculateMenAverageAge(people, century) {
...
+  const men = people.filter(person => person.sex === 'm');
+  const peopleOfCentury = people.filter(person =>
+    (Math.ceil(person.died / 100))
+    === century && person.sex === 'm');
+
+  if (century) {
+    return (peopleOfCentury.map(person =>
+      person.died - person.born)
+      .reduce((a, b) => a + b, 0)
+      / peopleOfCentury.length);
```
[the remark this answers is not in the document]
anstsot: not fixed

src/averageAges.js
```diff
@@ -38,6 +47,21 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const women = people.filter(person => person.sex === 'f');
+  const mothersNames = people.map(person => person.mother);
+  const mothersInList = women.filter(mothers =>
+    mothersNames.includes(mothers.name));
```
anstsot, line 53: you can simplify and filter the mothers with `withChildren` check inside the `filter` method

src/averageAges.js
```diff
@@ -15,12 +15,21 @@
...
+    (Math.ceil(man.died / 100))
+    === century);
+
+  const age = (timePeriod) =>
+    timePeriod.map(person =>
+      person.died - person.born)
+      .reduce((a, b) => a + b, 0)
+      / timePeriod.length;
+
+  const round = (num) => Math.round(num * 100) / 100;
```
anstsot, line 29: create separate function to calculate average age for array of people

anstsot (changes requested):

src/averageAges.js
```diff
@@ -14,13 +14,26 @@
  *
  * @return {number}
  */
+
+// Shared Functions
+const age = (timePeriod) =>
+  timePeriod.map(person =>
```
witflash, line 20: No need to use `map` method here. You can do it inside your `reduce` callback function.

witflash (approved):
Good job!  👍🏻

PR #1976, opened 2023-04-22
js_average-ages

Donovanoff:


src/averageAges.js
```diff
@@ -15,7 +15,22 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
+  const men = people.filter((person) => person.sex === 'm');
+  let sorted;
+
+  if (century === undefined) {
```
TarasHoliuk, line 21: don't use `if` statement here
use ternary inside the filter's callback instead

src/averageAges.js
```diff
@@ -37,7 +52,22 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter((person) => person.sex === 'f');
+  let sorted = women;
+
+  if (withChildren === true) {
```
TarasHoliuk, line 58: the same as in the previous function - check `withChildren` argument inside the filter's callback and if it `true` add an additional condition using `&&` operator

src/averageAges.js
```diff
@@ -37,7 +52,22 @@ function calculateMenAverageAge(people, century) {
...
+    });
+  }
+
+  const sumOfYears = sorted.reduce((acc, person) => {
+    const yearsLived = person.died - person.born;
+
+    return acc + yearsLived;
+  }, 0);
+
+  return sumOfYears / sorted.length;
```
TarasHoliuk, line 70: you are doing the same at least twice (or even 3 times) - move it to a separate function

src/averageAges.js
```diff
@@ -15,7 +15,22 @@
...
   // learn how to use array methods like .filter .map .some .every .find .reduce
```
TarasHoliuk, line 34: remove this comment

src/averageAges.js
```diff
@@ -54,8 +84,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  let children;
+
+  if (onlyWithSon) {
+    children = people.filter((person) => person.sex === 'm');
+  } else {
+    children = people.filter((person) => person.mother !== undefined);
+  }
+
+  const diffs = children.map((child) => {
+    const mother = mothers.find((person) => person.name === child.mother);
```
TarasHoliuk, line 98: you may use `people` array here, i.e `mothers` variable is redundant (you are checking `mother` field, so there won't be match with people that have `sex === 'm'`)

src/averageAges.js
```diff
@@ -54,8 +84,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+    if (mother !== undefined) {
+      return child.born - mother.born;
+    }
+  }).filter((diff) => diff !== undefined);
+
+  const sumOfDiffs = diffs.reduce((acc, diff) => acc + diff, 0);
+  const avgDiff = sumOfDiffs / diffs.length;
```
TarasHoliuk, line 106:
```suggestion
  return sumOfDiffs / diffs.length;
```

TarasHoliuk (changes requested):

VitaliyBondarenko1982 (approved):

src/averageAges.js
```diff
@@ -37,7 +44,17 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter((person) => person.sex === 'f');
```
ivadimko, line 47: You can simplify the women filtering by using a single filter statement with the && operator instead of using ternary operator.

src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter((person) => person.sex === 'm');
+
+  // eslint-disable-next-line max-len
+  const sorted = men.filter((person) => century === undefined ? true : century === Math.ceil(person.died / 100));
```
ivadimko, line 21: Use a single filter statement instead of chaining multiple filter statements. You can use the && operator in the callback function to simplify the logic.

src/averageAges.js
```diff
@@ -54,8 +71,30 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    children = people.filter((person) => person.mother !== undefined);
+  }
+
+  const diffs = children.map((child) => {
+    const mother = people.find((person) => person.name === child.mother);
+
+    if (mother !== undefined) {
+      return child.born - mother.born;
+    }
+  }).filter((diff) => diff !== undefined);
```
ivadimko, line 89: Instead of using separate map and filter functions, you can use a single reduce function to calculate the sum of age differences and the number of differences at once, then calculate the average outside the reduce function.

PR #1980, opened 2023-04-27
solution

luka4ok:
Please, check my solution.

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -15,11 +15,13 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person =>
+    century
+      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+      : person.sex === 'm'
+  );
+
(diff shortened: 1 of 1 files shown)


src/averageAges.js
```diff
@@ -55,7 +62,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    const yearOfMothersBirth = people.find(person =>
+      person.name === child.mother).born;
+
+    return (total + (child.born - yearOfMothersBirth) / children.length);
+  }, 0);
+}
+
+function getAverageAge(people) {
+  const ages = people.map(person => person.died - person.born);
+  const averageAges = ages.reduce((total, age) => total + age);
```
loralevitska, line 81: The second argument for the reduce function is missing, which should be the initial value of the accumulator.

loralevitska (approved):
Good job!

src/averageAges.js
luka4ok: Thank you for your response.

On [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#:~:text=was%20called%20upon.-,initialValue,-Optional) I found, that:

> If initialValue is not specified, accumulator is initialized to the first value in the array, and callbackFn starts executing with the second value in the array as currentValue.

But, probably, in the case of readability and consistency (in another reduces I am using initial values), I should use “, 0”.

PR #1988, opened 2023-04-30
Develop

domkab:


src/averageAges.js
```diff
@@ -1,61 +1,61 @@
...
+  const men = people.filter(person =>
+    century
+      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+      : person.sex === 'm'
+  );
+  const sumAge = men.reduce(
+    (sum, person) => sum + (person.died - person.born),
+    0
+  );
+
```
Anastasiia-Svintsova, line 13: You use this logic not once. Create a separate method for calculating the average age.

src/averageAges.js
```diff
@@ -1,61 +1,61 @@
...
- * @return {number}
- */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const mothers = onlyWithSon
+    ? people.filter(woman =>
+      people.some(
+        person => person.mother === woman.name && person.sex === 'm'))
+    : people.filter(woman =>
+      people.some(person => person.mother === woman.name));
```
Anastasiia-Svintsova, line 37: Refactor it to filter people only once. Move a ternary operator inside the filter method

src/averageAges.js
```diff
@@ -1,61 +1,61 @@
...
+
+  const children = onlyWithSon
+    ? people.filter(person =>
+      person.sex === 'm' && mothers.some(mother =>
+        person.mother === mother.name))
+
+    : people.filter(person =>
+      mothers.some(mother =>
+        person.mother === mother.name));
```
Anastasiia-Svintsova, line 46: same

Anastasiia-Svintsova (changes requested):

src/averageAges.js
domkab: `  const mothers = people.filter(woman =>
    onlyWithSon
      ? people.some(person =>
        person.mother === woman.name && person.sex === 'm')
      : people.some(person => person.mother === woman.name)
  );`

I changed the logic, filter it once, and then check if condition withSon is true, if yes then we perform first version of some method, if not we just look for all children.

Further nesting some statement would add an unnecesarry level of looping for checking onlyWithSons condition, to add i find it more readable.

Anastasiia-Svintsova (approved):

PR #1991, opened 2023-05-02
solution

generatedman:


DidenkoDenis306 (approved):

src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm'
+  && (!century || Math.ceil(person.died / 100) === century));
+
+  const ages = men.map(person => person.died - person.born);
+  const totalAge = ages.reduce((sum, age) => sum + age, 0);
+
+  return totalAge / men.length;
```
Anna-Tkachenko, line 24: You've repeated the same logic for calculating average age for array of people several time
Let's create separate function with this functionality

Anna-Tkachenko (changes requested):

DidenkoDenis306 (approved):

src/averageAges.js
```diff
@@ -14,12 +14,19 @@
  *
  * @return {number}
  */
+function average(ages, persons) {
```
33Orange, line 17: What is `average`?
Use more descriptive names. 
If you do some action in function, then describe it. Use verb as first word in name.
`calculateAverageAges` or `getAverageAges` would be better

src/averageAges.js
```diff
@@ -55,7 +68,18 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = onlyWithSon
+    ? people.filter(person => person.sex === 'm'
+    && people.some(mother => mother.name === person.mother))
+    : people.filter(person => people.some(mother =>
+      mother.name === person.mother));
```
33Orange, line 75: You have repeated logic in two parts of your comparison.
Try to improve the readability of this code.

33Orange (changes requested):
Well done! 🔥 

A few moments with readability.

src/averageAges.js
dinaminahi: Review @33Orange comment. You can have a check for `onlyWithSon` inside a first `filter`, then second filter won't be needed

src/averageAges.js
```diff
@@ -14,12 +14,19 @@
  *
  * @return {number}
  */
+function calculateAverageAges(ages, persons) {
```
dinaminahi, line 17: You can use the length of `ages` and you won't need a second argument

dinaminahi (changes requested):

dinaminahi (approved):
Well done! 👍

PR #1994, opened 2023-05-03
Develop

maxonshev:


src/averageAges.js
```diff
@@ -14,12 +14,27 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+
+  const diedInCentury = men.filter(man =>
+    Math.ceil(man.died / 100) === century);
```
polosanya, line 25: What if century is not passed? You should check it. 
It would be great to call filter method only once and use the ternary operator inside its callback

src/averageAges.js
```diff
@@ -14,12 +14,27 @@
...
+
+  const menAges = century
+    ? diedInCentury.map(man =>
+      man.died - man.born)
+
+    : men.map(man => man.died - man.born);
+
+  const sumMenAge = menAges.reduce((sum, age) => sum + age, 0);
```
polosanya, line 33: Let's create a function-helper, that takes an array of people and return their average age. It will help not to duplicate yourself and make your code cleaner

src/averageAges.js
```diff
@@ -37,7 +52,23 @@ function calculateMenAverageAge(people, century) {
...
+    people.some(person => person.mother === woman.name)
+  );
+
+  const womenAges = withChildren
+    ? mother.map(woman =>
+      woman.died - woman.born)
+
+    : women.map(woman => woman.died - woman.born);
+
+  const sumWomanAges = womenAges.reduce((sum, age) => sum + age, 0);
```
polosanya, line 67: Same comments here

polosanya (changes requested):
Good job! Let's try to optimize your solution a bit)

anstsot (approved):
Great job, approved!

PR #2000, opened 2023-05-04
finished version

EvhenIvchenko:


src/averageAges.js
```diff
@@ -15,6 +15,17 @@
...
+  let men = people.filter(person => person.sex === 'm');
+
+  century && (men = men
+    .filter(person => Math.ceil(person.died / 100) === century));
+
+  const sumAge = men.reduce(
+    (sum, man) => sum + (man.died - man.born), 0
+  );
+
+  return sumAge / men.length;
```
tetiana-ivanchenko, line 27: move it to the separate function

src/averageAges.js
```diff
@@ -15,6 +15,17 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  let men = people.filter(person => person.sex === 'm');
+
+  century && (men = men
+    .filter(person => Math.ceil(person.died / 100) === century));
```
tetiana-ivanchenko, line 21:
```suggestion
  const men = people.filter(person =>  century 
    ? person.sex === 'm'
    : person.sex === 'm' && Math.ceil(person.died / 100) === century
  );
```

src/averageAges.js
```diff
@@ -37,7 +48,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let women = people.filter(person => person.sex === 'f');
+
+  withChildren && (women = women
+    .filter(woman => people.find(child => child.mother === woman.name)));
```
tetiana-ivanchenko, line 54: combine filters

tetiana-ivanchenko (changes requested):

src/averageAges.js
EvhenIvchenko: why is it bad to use it inside reduce ?

src/averageAges.js
EvhenIvchenko: the same here , is it bad ? because my code follows the instruction 
  // without nesting

EvhenIvchenko: and maybe you meant that   
const men = people.filter(person => century
    ? person.sex === 'm' && Math.ceil(person.died / 100) === century
    : person.sex === 'm'
  );

src/averageAges.js
tetiana-ivanchenko: you should move reduce, because you use logic of getting average age twice.
If you have any questions about mentor comments - you can ask in in the fe_chat, instead of requesting re-review

tetiana-ivanchenko (changes requested):

src/averageAges.js
```diff
@@ -37,7 +51,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => withChildren
+    ? person.sex === 'f' && people.find(child => child.mother === person.name)
+    : person.sex === 'f');
+
+  const sumMowanAge = ReducedAverageAge(women);
```
etojeDenys, line 58: There is a typo in the variable name 'sumMowanAge', change it to 'sumWomenAge'

etojeDenys (changes requested):

IvanFesenko (approved):

PR #2008, opened 2023-05-09
add task solution

maksym-nemera:


src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
-  // without nesting
+  const filterMen = people.filter(person =>
+    (century
+      ? Math.ceil(person.died / 100) === century
+      : true
+    )
+    && person.sex.toLowerCase() === 'm');
+
+  const mensAges = filterMen.map(men => men.died - men.born);
+  const sumAges = mensAges.reduce((sum, year) => sum + year, 0);
```
DidenkoDenis306, line 26: You calculate the average age in each function. It would be cool to put this operation in a separate function and use it instead of writing the same thing several times.

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
+
+  return sumAges / mensAges.length;
```
DidenkoDenis306, line 28: You calculate the average age in each function. It would be cool to put this operation in a separate function and use it instead of writing the same thing several times.

src/averageAges.js
```diff
@@ -37,7 +43,17 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filterWomen = people.filter(person =>
+    (withChildren
+      ? people.find(children => children.mother === person.name)
+      : true
+    )
+    && person.sex.toLowerCase() === 'f');
+
+  const womensAges = filterWomen.map(men => men.died - men.born);
```
DidenkoDenis306, line 53: I think we take the woman from the women, not the men))

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filterMen = people.filter(person =>
+    (century
+      ? Math.ceil(person.died / 100) === century
+      : true
+    )
+    && person.sex.toLowerCase() === 'm');
+
+  const mensAges = filterMen.map(men => men.died - men.born);
```
DidenkoDenis306, line 25: it would be better to replace men with man

DidenkoDenis306 (changes requested):
Good Job, almost done!👍

src/averageAges.js
```diff
@@ -55,7 +69,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    const filterMother = people.find(person => person.name === child.mother);
+
+    return child.born - filterMother.born;
+  });
+
+  return averageAges(diffAges);
+}
+
+function averageAges(arrayAges) {
+  const sumAges = arrayAges.reduce((sum, year) => sum + year, 0);
```
silchyn, line 89:
```suggestion
  const sumAges = arrayAges.reduce((sum, age) => sum + age, 0);
```

src/averageAges.js
```diff
@@ -15,11 +15,16 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filterMen = people.filter(person =>
```
silchyn, line 18:
```suggestion
  const men = people.filter(person =>
```

silchyn (changes requested):

src/averageAges.js
```diff
@@ -55,7 +69,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const diffAges = children.map(child => {
+    const filterMother = people.find(person => person.name === child.mother);
+
+    return child.born - filterMother.born;
+  });
+
+  return averageAges(diffAges);
+}
+
+function averageAges(arrayAges) {
```
silchyn, line 88:
```suggestion
function averageAges(ages) {
```

silchyn (approved):

PR #2011, opened 2023-05-10
solution

dseq0:


src/averageAges.js
```diff
@@ -1,61 +1,51 @@
...
- */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const result = people.filter(person => person.sex === 'f'
+    && (!withChildren || people.find(children =>
+      children.mother === person.name)));
+  const finalResult = result.reduce((sum, person) =>
+    sum + person.died - person.born, 0);
+
+  return finalResult / result.length;
```
DarkMistyRoom, line 19: this logic is the same as in the `calculateMenAverageAge`, so move it to a separate function

src/averageAges.js
```diff
@@ -1,61 +1,51 @@
...
+    const mother = mothersObject[motherName];
+
+    return mother.childrenBornYears.map(year => {
+      mothersTotal += 1;
+
+      return year - mother.born;
+    });
+  });
+
+  return ageDiffs.reduce((ages, age) => ages + age, 0) / mothersTotal;
```
DarkMistyRoom, line 48: interesting solution! But consider changing the approach to start with finding all children's -> mother of each child -> calculate the difference. It should reduce calculations

DarkMistyRoom (approved):

PR #2014, opened 2023-05-10
try

Volodymyr-Kustikov:


src/averageAges.js
```diff
@@ -14,12 +14,25 @@
...
+
+  function withoutCentury() {
+    const foundation = people.filter(person => person.sex === 'm');
+    const finallyFunc = foundation.reduce((sum, person) =>
+      sum + person.died - person.born, 0);
+
+    return finallyFunc / foundation.length;
+  }
+
+  return century > 0 ? withCentury() : withoutCentury();
```
tetiana-ivanchenko, line 35: let's simplify your solution
at first you should create variable where you should filter people array depending on the condition
also use more meaningful names
then create separate function outside, that will calculate average age, and use it here and in the next
function

src/averageAges.js
```diff
@@ -37,7 +50,24 @@ function calculateMenAverageAge(people, century) {
...
+  function withChildrens() {
+    const foundation = people.filter(person => person.sex === 'f'
+    && people.some(mom => mom.mother === person.name));
+    const finallyFunc = foundation.reduce((sum, person) =>
+      sum + person.died - person.born, 0);
+
+    return finallyFunc / foundation.length;
+  }
+
+  return withChildren === true ? withChildrens() : withoutChildrens();
```
tetiana-ivanchenko, line 70: the same here

tetiana-ivanchenko (changes requested):

src/averageAges.js
SanyaBratashchuk, line 35:
```suggestion
  const men = people.filter(({sex, died })=> century ? sex === 'm' && Math.ceil(died / 100) === century ) : sex === 'm';
  // move logic below to the separate function 
  return men.reduce((sum, man) => sum + man.died - man.born, 0) / men.length
```

SanyaBratashchuk: Try to rewrite functions

SanyaBratashchuk (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(({ sex, died }) => century ? sex === 'm'
+    && Math.ceil(died / 100) === century : sex === 'm');
+
+  const brosReducing = men.reduce((sum, man) => sum + man.died - man.born, 0)
+    / men.length;
+
+  return brosReducing;
```
anstsot, line 24: let's create separate function for calculating the average age for people

src/averageAges.js
```diff
@@ -37,7 +39,34 @@ function calculateMenAverageAge(people, century) {
...
+  // function withChildrens() {
+  //   const foundation = people.filter(person => person.sex === 'f'
+  //   && people.some(mom => mom.mother === person.name));
+  //   const finallyFunc = foundation.reduce((sum, person) =>
+  //     sum + person.died - person.born, 0);
+
+  //   return finallyFunc / foundation.length;
+  // }
+
+  // return withChildren === true ? withChildrens() : withoutChildrens();
```
anstsot, line 69: don't push comments

anstsot (changes requested):

src/averageAges.js
```diff
@@ -1,61 +1,42 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(({ sex, died }) => century ? sex === 'm'
+    && Math.ceil(died / 100) === century : sex === 'm');
+
+  const brosReducing = men.reduce((sum, man) => sum + man.died - man.born, 0);
+
+  const calculatingTheAverageAgeForPeople = brosReducing / men.length;
```
sTorba24, line 9: still not fixed. You can create separated function where you pass an array, then calculating sum inside and return value divided by array length

sTorba24 (approved):
Good work! I approve it, but recommend to do with separated function to not repeat the same action as on `calculateMenAverageAge` and `calculateWomenAverageAge`. It's good practice to write DRY code

PR #2018, opened 2023-05-12
avarageAges Solution

ZupaFly:


src/averageAges.js
```diff
@@ -15,11 +15,22 @@
...
+  const menLeng = (people
+    .filter((i) => arguments.length < 2
+      ? i.sex === 'm'
+      : i.sex === 'm' && Math.ceil(i.died / 100) === century))
+    .length;
+
+  const age = +(people
+    .filter((i) => arguments.length < 2
+      ? i.sex === 'm'
+      : i.sex === 'm' && Math.ceil(i.died / 100) === century)
```
agniya-i, line 27: there is similar logic here, you can combine it together to find all men first and then use this variable to get the length and calculate the sum

src/averageAges.js
```diff
@@ -37,7 +48,19 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let totalAge = 0;
+  let numWomen = 0;
+
+  for (const person of people) {
+    if (person.sex === 'f') {
```
agniya-i, line 55: It would be better to use `filter` method also here

src/averageAges.js
```diff
@@ -54,8 +77,24 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    : people
+      .filter(person => people
+        .find(mother => mother.name === person.mother));
+
+  const diffArr = filteredPeople.map(person =>
+    person.born - people
+      .find(mother => mother.name === person.mother).born);
+
+  function getAverageAge(input) {
+    return input.reduce((sum, age) => sum + age, 0) / input.length;
```
agniya-i, line 94: You can use `getAverageAge` for other functions, where you have similar calculation

src/averageAges.js
```diff
@@ -15,11 +15,22 @@
...
-  // without nesting
+  const menLeng = (people
+    .filter((i) => arguments.length < 2
+      ? i.sex === 'm'
+      : i.sex === 'm' && Math.ceil(i.died / 100) === century))
+    .length;
+
+  const age = +(people
+    .filter((i) => arguments.length < 2
+      ? i.sex === 'm'
```
agniya-i, line 26: `i` is not really good name for a variable, use more descriptive names

agniya-i (changes requested):

src/averageAges.js
```diff
@@ -37,7 +39,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenAge = people
+    .filter((woman) => arguments.length < 2
```
Anastasiia-Svintsova, line 43: Instead of checking `arguments.length < 2`, consider using `!withChildren` for better readability.

src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menAge = people
+    .filter((man) => arguments.length < 2
+      ? man.sex === 'm'
+      : man.sex === 'm' && Math.ceil(man.died / 100) === century)
+    .map((index) => index.died - index.born);
```
Anastasiia-Svintsova, line 22: Consider using a more descriptive name for the variable `index` in the `.map()` function. It represents a person, so you could rename it to `person`.

src/averageAges.js
```diff
@@ -37,7 +39,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenAge = people
+    .filter((woman) => arguments.length < 2
+      ? woman.sex === 'f'
+      : woman.sex === 'f' && people.some(child => child.mother === woman.name))
+    .map((index) => index.died - index.born);
```
Anastasiia-Svintsova, line 46: The same comment about naming goes here

src/averageAges.js
```diff
@@ -15,11 +15,13 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menAge = people
+    .filter((man) => arguments.length < 2
```
Anastasiia-Svintsova, line 19: no need to use arguments here. Think how to use the `century` variable correctly

src/averageAges.js
```diff
@@ -54,8 +62,24 @@ function calculateWomenAverageAge(people, withChildren) {
...
-function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+function calculateAverageAgeDiff(people, onlyWithSon = false) {
+  const filteredPeople = onlyWithSon
+    ? people
+      .filter(person => person.sex === 'm' && people
+        .find(mother => mother.name === person.mother))
+    : people
+      .filter(person => people
+        .find(mother => mother.name === person.mother));
```
Anastasiia-Svintsova, line 72: Seems that you repeat the same find method twice. Try to refactor it to avoid repetition

Anastasiia-Svintsova (changes requested):

PR #2019, opened 2023-05-12
Average Ages

levchenko-dmytro:


src/averageAges.js
```diff
@@ -20,6 +20,18 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
```
OleziO, line 20: Remove all comments from the code.

OleziO (approved):

PR #2020, opened 2023-05-13
Solution

Pa1eOrc:


src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const men = people.filter(person => person.sex === 'm');
+  const menLifespans = century
+    ? men.filter(person => Math.ceil(person.died / 100) === century)
```
SanyaBratashchuk, line 25: move ternary inside filter to avoid extra call of filter

src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
...
+      .map(person => person.died - person.born)
+    : men.map(person => person.died - person.born);
+  const totalMenLifespan = menLifespans.reduce(
+    (sum, lifespan) => sum + lifespan, 0);
+  const averageMenLifespan = totalMenLifespan / menLifespans.length;
+
+  return averageMenLifespan;
```
SanyaBratashchuk, line 32: move logic related to calculate average age to the separate function and reuse it

src/averageAges.js
```diff
@@ -37,7 +47,21 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const women = people.filter(person => person.sex === 'f');
+  const womenLifespans = [];
+
+  women.forEach(woman => {
+    if (withChildren ? people.some(
+      person => person.mother === woman.name) : true) {
+      womenLifespans.push(woman.died - woman.born);
+    }
+  });
```
SanyaBratashchuk, line 58: same simplify the solution by moving ternary inside filter's callback

SanyaBratashchuk (changes requested):

src/averageAges.js
```diff
@@ -37,7 +46,17 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => person.sex === 'f');
+
+  const womenLifespans = women.reduce((totalWomenLifespan, woman) =>
+    (withChildren ? people.some(person => person.mother === woman.name) : true)
+      ? (totalWomenLifespan.push(woman.died - woman.born), totalWomenLifespan)
+      : totalWomenLifespan,
+  []);
```
anstsot, line 55: please filter women with `withChildren` argument and use simple `reduce` method without `push`

src/averageAges.js
```diff
@@ -55,7 +74,23 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const childrenWithMother = [];
+
+  people.filter(person => person.mother && (!onlyWithSon
```
anstsot, line 79: filter returns the filtered array so please rewrite the logic without pushing into `childrenWithMother` array and directly return the needed data

anstsot (changes requested):

SanyaBratashchuk (approved):

PR #2023, opened 2023-05-15
solution

skinnyjazz:


src/averageAges.js
```diff
@@ -14,14 +14,18 @@
  *
  * @return {number}
  */
+const averageOfPeople = (people) =>
```
33Orange, line 17: The function name should be more descriptive. Consider renaming it to 'calculateAverageAge' instead of 'averageOfPeople'
`[NAMING] - Name of functions should start with verbs. It will be easier to understand a function's purpose.` Checklist p1.

33Orange (changes requested):
Well done!

Just a one moment with readability.

src/averageAges.js
```diff
@@ -36,8 +40,18 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothersNames = [
+    ...new Set(people.map((person) => person.mother || 'not specified')),
+  ];
+  const arrayOfWoman = people.filter((person) => {
```
dinaminahi, line 48: It is better to not specify the type of variable in it's name, so it can be just `women`

src/averageAges.js
```diff
@@ -14,14 +14,18 @@
...
+  people.reduce((acc, item) => acc + item.died - item.born, 0) / people.length;
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
-}
+  const arrayOfMen = people.filter((person) => {
```
dinaminahi, line 21: It is better to not specify the type of variable in it's name, so it can be just `men`

dinaminahi (approved):
Good job 👍

PR #2024, opened 2023-05-15
solution

HannieA:


src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = century
+    ? people.filter(person => person.sex === 'm'
+      && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === 'm');
+  const sum = men.reduce((prev, man) => (man.died - man.born) + prev, 0);
+
+  return sum / men.length;
```
anstsot, line 24: create separate function to calculate the avarage age of array of the people and reuse it here and in `calculateWomenAverageAge`

src/averageAges.js
```diff
@@ -14,12 +14,18 @@
...
+}
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm'
+  && (century ? Math.ceil(person.died / 100) === century : true));
```
Anastasiia-Svintsova, line 26: The ternary operator `century ? Math.ceil(person.died / 100) === century : true` can be simplified by removing the `: true` part. This will make the code more readable.

PR #2025, opened 2023-05-16
Solution

yuliia-nudyk:


src/averageAges.js
```diff
@@ -20,6 +20,25 @@ function calculateMenAverageAge(people, century) {
...
   // without nesting
+
+  const manCheck = person => person.sex === 'm';
+  const centuryCheck = person => {
+    return manCheck(person) && Math.ceil(person.died / 100) === century;
+  };
+
+  let men;
+
+  if (arguments.length === 2) {
```
polosanya, line 31: It is better to check if century is not undefined

src/averageAges.js
```diff
@@ -20,6 +20,25 @@ function calculateMenAverageAge(people, century) {
...
+    men = people.filter(centuryCheck);
+  } else {
+    men = people.filter(manCheck);
```
polosanya, line 34: Better use ternary operator for that:
```
const men = condition
  ? .....
  : .....
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator

src/averageAges.js
```diff
@@ -37,7 +56,26 @@ function calculateMenAverageAge(people, century) {
...
+
+  if (arguments.length === 2) {
+    women = people.filter(person => motherCheck(person.name));
+  } else {
+    women = people.filter(womanCheck);
+  }
+
+  const womenAges = women.map(woman => woman.died - woman.born);
+  const sumAges = womenAges.reduce((sum, age) => sum + age);
+  const averageAge = sumAges / womenAges.length;
```
polosanya, line 76: You have the same logic for both women and men. Let's create a separate function for calculating an average age of the array

polosanya (changes requested):
Good job! Let's improve your solution a bit)

src/averageAges.js
```diff
@@ -20,6 +20,19 @@ function calculateMenAverageAge(people, century) {
...
+    return manCheck(person) && Math.ceil(person.died / 100) === century;
+  };
+
+  const men = century !== undefined
+    ? people.filter(centuryCheck)
+    : people.filter(manCheck);
+
+  const menAvrAge = getAverageAge(men);
+
+  return menAvrAge;
```
anstsot, line 35: you don't have to create separate variable
```suggestion
  return getAverageAge(men);
```

src/averageAges.js
```diff
@@ -37,7 +50,20 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womanCheck = person => person.sex === 'f';
+  const motherCheck = motherName => {
+    const mother = people.find(person => person.mother === motherName);
+
+    return mother;
+  };
+
+  const women = withChildren === true
```
anstsot, line 60:
```suggestion
  const women = withChildren
```

anstsot (changes requested):

src/averageAges.js
```diff
@@ -55,7 +77,35 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const children = onlyWithSon === true
+    ? people.filter(person => checkMother(person.mother) && manChecking(person))
+    : people.filter(person => checkMother(person.mother));
+
+  const ageDiff = children.map(
+    child => child.born - findMother(child.mother).born
+  );
+
+  function findMother(motherName) {
```
Anastasiia-Svintsova, line 93: Remove the `findMother` function since it is a duplicate of the `checkMother` function.

src/averageAges.js
```diff
@@ -20,6 +20,17 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  const manCheck = person => person.sex === 'm';
+  const centuryCheck = person => {
+    return manCheck(person) && Math.ceil(person.died / 100) === century;
+  };
+
+  const men = century !== undefined
```
Anastasiia-Svintsova, line 29: You can simplify it
```suggestion
  const men = century
```

src/averageAges.js
```diff
@@ -37,7 +48,18 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womanCheck = person => person.sex === 'f';
+  const motherCheck = motherName => {
+    const mother = people.find(person => person.mother === motherName);
+
+    return mother;
+  };
```
Anastasiia-Svintsova, line 56: What will be in case the function won't find anything?
```suggestion
  const motherCheck = (motherName) => 
    people.find(person => person.mother === motherName) || null;
```

src/averageAges.js
```diff
@@ -55,7 +77,35 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const checkMother = motherName => people.find(
+    person => person.name === motherName
+  );
+  const manChecking = person => person.sex === 'm';
+
+  const children = onlyWithSon === true
```
Anastasiia-Svintsova, line 85:
```suggestion
  const children = onlyWithSon
```

src/averageAges.js
```diff
@@ -55,7 +77,35 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    ? people.filter(person => checkMother(person.mother) && manChecking(person))
```
Anastasiia-Svintsova, line 86: You can move a ternary operator inside the find function to not repeat it twice

Anastasiia-Svintsova (changes requested):

src/averageAges.js
yuliia-nudyk: How could i do it?

src/averageAges.js
```diff
@@ -20,6 +20,17 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
```
OleziO, line 20: Remove all comments from the code.

OleziO (approved):

PR #2027, opened 2023-05-16
Solution

Ihor-Zakharov:


src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredPeople = (century)
+    ? people.filter(person => Math.ceil(person.died / 100) === century
+      && person.sex === 'm')
+    : people.filter(person => person.sex === 'm');
```
anstsot, line 21: it can be simplified. the `centrury` check can be used inside the `filter` method.
please fix other cases and simplify them as well.
```suggestion
  const filteredPeople = people.filter(person => person.sex === 'm' 
    && century ? Math.ceil(person.died / 100) === century : true);
```

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
+
+  return filteredPeople.reduce(
+    (sum, person) => sum + person.died - person.born, 0
+  ) / filteredPeople.length;
```
anstsot, line 25: create separate function and reuse to calculate average age of people

anstsot (changes requested):

PR #2028, opened 2023-05-16
Develop

VolodymyrPst:


src/averageAges.js
```diff
@@ -55,7 +61,23 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const ages = [];
+
+  onlyWithSon ? people.filter(person => {
+    people.some(male => {
+      if (male.mother === person.name && male.sex === 'm') {
+        ages.push(male.born - person.born);
```
vlad-demchuk, line 69: you can do this using iteration methods without `push`

PR #2029, opened 2023-05-17
average ages

pos1bl:


src/averageAges.js
```diff
@@ -15,11 +15,12 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const result = people
+    .filter(person => person.sex === 'm'
+      && (Math.ceil(person.died / 100) === century || arguments.length < 2))
+    .map(person => person.died - person.born);
+
+  return result.reduce((avg, age) => avg + age) / result.length;
```
volodymyr-soltys97, line 23: You can create a helper function, take it outside the bounds and use it everywhere in the return to avoid duplicating code

volodymyr-soltys97 (approved):
Great job 👍

PR #2030, opened 2023-05-17
solution

Anastasiia-Shahoval:


src/averageAges.js
```diff
@@ -1,61 +1,53 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menDied = century
+    ? people.filter(person => person.sex === 'm'
+    && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === 'm');
+  const menAge = menDied.map(person => person.died - person.born);
+  const menAverageAge = menAge.reduce((sum, age) => {
+    return sum + age;
+  }, 0) / menAge.length;
```
anstsot, line 11: create separate function to calculate average age and reuse it

src/averageAges.js
```diff
@@ -1,61 +1,53 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menDied = century
+    ? people.filter(person => person.sex === 'm'
+    && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === 'm');
```
anstsot, line 7: add century check inside the filter
```suggestion
  const menDied = people.filter(person => person.sex === 'm'
      && century ? Math.ceil(person.died / 100) === century) : true);
```

src/averageAges.js
```diff
@@ -1,61 +1,53 @@
...
- * @return {number}
- */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = withChildren
+    ? people.filter(person => people.some(child => {
+      return person.name === child.mother;
+    })
+    && person.sex === 'f')
+    : people.filter(person => person.sex === 'f');
```
anstsot, line 22: can be simplified as well

src/averageAges.js
```diff
@@ -1,61 +1,53 @@
...
+    const mother = people.find(person => child.mother === person.name);
+    const ages = child.born - mother.born;
+
+    return ages;
+  });
+  const AverageAgeDiff = ageDiff.reduce((sum, age) => {
+    return sum + age;
+  }, 0) / ageDiff.length;
+
+  return AverageAgeDiff;
```
anstsot, line 50: there is no need to create separate variable
```suggestion
  return ageDiff.reduce((sum, age) => {
    return sum + age;
  }, 0) / ageDiff.length;
```

anstsot (changes requested):

src/averageAges.js
mykhalenych: didn't fix it

src/averageAges.js
mykhalenych: as well

mykhalenych (changes requested):

src/averageAges.js
```diff
@@ -1,61 +1,46 @@
...
- * @return {number}
- */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => person.sex === 'f'
+    && withChildren ? people.some(child => {
+      return person.name === child.mother;
+    }) : true);
+
+  const womenAge = women.map(person => person.died - person.born);
```
denys-danyliuk, line 22: you repeat this logic, it's better to move this to function

denys-danyliuk (approved):

PR #2035, opened 2023-05-17
solution

serentinos:


src/averageAges.js
```diff
@@ -20,6 +20,21 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
```
33Orange, line 22: Clean up the code from unnecessary comments.

src/averageAges.js
```diff
@@ -20,6 +20,21 @@ function calculateMenAverageAge(people, century) {
...
+      const personCenturyDeath = Math.ceil(person.died / 100);
+      const personSex = person.sex;
+
+      return century
+        ? personSex === 'm' && century === personCenturyDeath
+        : personSex === 'm';
+    });
+
+  const averageMenAge = mens
+    .reduce((prev, { born, died }) => prev + (died - born), 0) / mens.length
```
33Orange, line 34: You do the same calculations with reduce in every function. Maybe we should move this calculation to a separate helper?

src/averageAges.js
```diff
@@ -31,13 +46,25 @@ function calculateMenAverageAge(people, century) {
...
+    ? people
+      .filter(person => (
+        people.find(child => child.mother === person.name && person.sex === 'f')
+      ))
+    : people
+      .filter(person => person.sex === 'f');
+
+  const averageAge = filteredWoman
+    .reduce((sum, { born, died }) => sum + (died - born), 0)
+      / filteredWoman.length;
```
33Orange, line 65: You do the same calculations with reduce in every function. Maybe we should move this calculation to a separate helper?

src/averageAges.js
```diff
@@ -49,13 +76,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    const ageDifference = personChildren
+      .map(child => child.born - person.born);
+
+    differences
+      .push(...ageDifference);
+  });
+
+  const avgDifference = differences
+    .reduce((sum, current) => sum + current, 0)
+    / differences.length;
```
33Orange, line 103: You do the same calculations with reduce in every function. Maybe we should move this calculation to a separate helper?

src/averageAges.js
```diff
@@ -31,13 +46,25 @@ function calculateMenAverageAge(people, century) {
...
+//  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filteredWoman = withChildren
+    ? people
+      .filter(person => (
+        people.find(child => child.mother === person.name && person.sex === 'f')
+      ))
+    : people
+      .filter(person => person.sex === 'f');
```
33Orange, line 61: You use a filter in both cases. Maybe it's better to do a check `withChildren` condition inside the filter?

src/averageAges.js
```diff
@@ -49,13 +76,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const differences = [];
+
+  people.forEach(person => {
+    const personChildren = onlyWithSon
+      ? people
+        .filter(child => person.name === child.mother && child.sex === 'm')
+      : people
+        .filter(child => person.name === child.mother);
```
33Orange, line 92: You use a filter in both cases. Maybe it's better to do a check `onlyWithSon` condition inside the filter?

33Orange (changes requested):
Well done!

A few moment with readability

serentinos:
@33Orange Thank you for all your suggestions! This advice is really helpful!

33Orange (approved):
Well done ! 🔥

PR #2037, opened 2023-05-17
add task solution

AlexKherson:


src/averageAges.js
```diff
@@ -20,6 +20,13 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  const agesOfMen = people
+    .filter(({ sex, died }) => sex === 'm' && isDiedInCentury(died, century))
```
33Orange, line 25: The isDiedInCentury function should be called only when `century` is provided.

src/averageAges.js
```diff
@@ -37,7 +44,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const agesOfWomen = people
+    .filter(person => {
+      const isMother = !withChildren || isHasChildren(people, person.name);
```
33Orange, line 49: The isHasChildren function should be called only when `withChildren` is true.

src/averageAges.js
```diff
@@ -20,6 +20,13 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
```
33Orange, line 22: Delete unnecessary comments

src/averageAges.js
```diff
@@ -20,6 +20,13 @@ function calculateMenAverageAge(people, century) {
...
+
+  const agesOfMen = people
+    .filter(({ sex, died }) => sex === 'm' && isDiedInCentury(died, century))
+    .map(person => person.died - person.born);
+  const totalAges = agesOfMen.reduce((a, b) => a + b, 0);
+
+  return (totalAges / agesOfMen.length) || 0;
```
33Orange, line 29: You use `reduce` and division by length in every function. Maybe we should move this logic to a separate helper?

src/averageAges.js
```diff
@@ -37,7 +44,16 @@ function calculateMenAverageAge(people, century) {
...
+  const agesOfWomen = people
+    .filter(person => {
+      const isMother = !withChildren || isHasChildren(people, person.name);
+
+      return person.sex === 'f' && isMother;
+    })
+    .map(person => person.died - person.born);
+  const totalAges = agesOfWomen.reduce((a, b) => a + b, 0);
+
+  return (totalAges / agesOfWomen.length) || 0;
```
33Orange, line 56: You use `reduce` and division by length in every function. Maybe we should move this logic to a separate helper?

src/averageAges.js
```diff
@@ -55,7 +71,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
+        .includes(child.mother) && (!onlyWithSon || child.sex === 'm');
+    });
+  const differenceAges = children.map(child => {
+    const mother = mothers.find(mom => mom.name === child.mother);
+
+    return child.born - mother.born;
+  });
+  const totalDifferenceAges = differenceAges.reduce((a, b) => a + b, 0);
+
+  return (totalDifferenceAges / differenceAges.length) || 0;
```
33Orange, line 89: You use `reduce` and division by length in every function. Maybe we should move this logic to a separate helper?

33Orange (changes requested):
Good job !

A few moments with readability.

PR #2039, opened 2023-05-17
Solution

bohdan-mykhailenko:


src/averageAges.js
```diff
@@ -34,10 +49,32 @@ function calculateMenAverageAge(people, century) {
  * @param {object[]} people
  * @param {boolean} withChildren - optional
  *
- * @return {number}
+ * @return {Sum}
```
dinaminahi, line 52: Change the type of return value from 'Sum' to 'number'.

src/averageAges.js
```diff
@@ -10,16 +9,32 @@
  * Divide year of person's death by 100: Math.ceil(person.died / 100)
  *
  * @param {object[]} people
- * @param {number} century - optional
+ * @param {Sum} century - optional
  *
- * @return {number}
+ * @return {Sum}
```
dinaminahi, line 14: Change the type of return value from 'Sum' to 'number'.

dinaminahi (changes requested):
Good job! 👍 
Few moments to fix

naziamoff (approved):
👍

PR #2040, opened 2023-05-17
Solution

vovaAlieksieiev:


src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const man = people.filter(
+    person => person.sex === 'm'
+      && (!century || century
+        === Math.ceil(person.died / 100))
+  );
+
+  const manAge = man.map(person => person.died
```
YegorKorenkov, line 24: you can shorten these two lines of code, just count the difference in the reduce method

YegorKorenkov (approved):
Looks good

PR #2043, opened 2023-05-18
solution

Yuliya1907:


src/averageAges.js
```diff
@@ -15,6 +15,19 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  let mens = people.filter(person => person.sex === 'm');
+
+  if (century) {
+    mens = mens.filter(men => Math.ceil(men.died / 100) === century);
+  }
```
naziamoff, line 22: let's handle that inside a ternary in filter method

src/averageAges.js
```diff
@@ -15,6 +15,19 @@
...
+
+  const total = mens.reduce((sum, age) =>
+    sum + (age.died - age.born),
+  0
+  );
```
naziamoff, line 27: total what?
Also - seems you use this logic in each function - why don't we create a helper for that?

naziamoff: also, let's take a look at the checklist again, p.2

src/averageAges.js
```diff
@@ -15,6 +15,19 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  let mens = people.filter(person => person.sex === 'm');
```
naziamoff, line 18: lexical comment
```suggestion
  let men = people.filter(person => person.sex === 'm');
```

src/averageAges.js
```diff
@@ -37,7 +50,18 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womens = people.filter((person) =>
```
naziamoff, line 53: woman - one
women - multiple
```suggestion
  const women = people.filter((person) =>
```

src/averageAges.js
```diff
@@ -37,7 +50,18 @@ function calculateMenAverageAge(people, century) {
...
+    withChildren
+      ? people.find((child) => child.mother === person.name) !== undefined
```
naziamoff, line 55: what's the check for undefined?

src/averageAges.js
```diff
@@ -55,7 +79,31 @@ function calculateWomenAverageAge(people, withChildren) {
...
+        }
+      })
+      : people.some((mother) => {
+        if (person.mother === mother.name) {
+          person['motherBorn'] = mother.born;
+
+          return true;
+        }
+      })
+  );
```
naziamoff, line 99: super difficult to read. Try to separate some logic into helpers or find better ways to improve its redability

33Orange (approved):
Well done ! 🔥

PR #2045, opened 2023-05-18
add solution

growingananas:


src/averageAges.js
```diff
@@ -20,6 +20,18 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const men = people.filter(person => person.sex === 'm');
```
naziamoff, line 23: let's filter with/without century right here, inside the filter callback

src/averageAges.js
```diff
@@ -20,6 +20,18 @@ function calculateMenAverageAge(people, century) {
...
+
+  const menFilteredByCentury = century
+    ? men.filter(person => Math.ceil(person.died / 100) === century)
+    : men;
+
+  const sumOfMenAges = menFilteredByCentury.reduce((sum, man) =>
+    sum + (man.died - man.born), 0);
```
naziamoff, line 30: you use this logic a couple of times in the task, so it seems reasonable to create a helper function for that

naziamoff (changes requested):
fix comments and also take a look at the checklist and fix formatting

src/averageAges.js
```diff
@@ -20,6 +20,21 @@ function calculateMenAverageAge(people, century) {
...
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const filteredMen = people.filter(person => person.sex === 'm'
+    && (century
+      ? Math.ceil(person.died / 100) === century
+      : true));
+
+  const sumOfMenAges = calculateSumOfAges(filteredMen);
+
+  const menAgeAverage = sumOfMenAges / filteredMen.length;
```
silchyn, line 30: this step is repeated in every function

silchyn, line 30: what if `filteredMen.length` is 0?

silchyn (changes requested):

src/averageAges.js
```diff
@@ -20,6 +20,23 @@ function calculateMenAverageAge(people, century) {
...
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const filteredMen = people.filter(person => person.sex === 'm'
+    && (century
+      ? Math.ceil(person.died / 100) === century
+      : true));
+
+  const sumOfMenAges = calculateSumOfAges(filteredMen);
+
+  return filteredMen.length ? getAverage(sumOfMenAges, filteredMen.length) : 0;
```
silchyn, line 30: this check should be performed in the `getAverage`

src/averageAges.js
```diff
@@ -20,6 +20,23 @@ function calculateMenAverageAge(people, century) {
...
+}
+
+function calculateSumOfAges(people) {
+  return people.reduce((sum, person) =>
+    sum + (person.died - person.born), 0);
+}
+
+function getAverage(sum, length) {
```
silchyn, line 38: what's the point of the `getAverage` function if we calculate the sum ourselves?

silchyn (changes requested):

src/averageAges.js
```diff
@@ -20,6 +20,20 @@ function calculateMenAverageAge(people, century) {
...
+    && (century
+      ? Math.ceil(person.died / 100) === century
+      : true));
+
+  const menAges = filteredMen.map(man => man.died - man.born);
+
+  return getAverage(menAges, filteredMen.length);
+}
+
+function getAverage(ages, length) {
```
silchyn, line 33: do we need `length` here?

silchyn (approved):

src/averageAges.js
growingananas: not length, but quantity? function getAverage(ages, quantity) ?

PR #2047, opened 2023-05-18
Solution

Shevtsooov:


src/averageAges.js
```diff
@@ -14,12 +14,17 @@
  *
  * @return {number}
  */
+
+const averageAge = (people) => people.reduce(
+  (sum, person) => sum + person.died - person.born, 0)
+  / people.length;
+
```
naziamoff, line 21: does `averageAge` name sounds good for a function?

naziamoff (approved):
great, super clear code. Just one thing:

naziamoff (approved):
great, super clear code. Just one thing:

src/averageAges.js
Shevtsooov: Got it, will correct, thanks!

PR #2048, opened 2023-05-18
add averageAges task solution

booobryyyk:


src/averageAges.js
```diff
@@ -37,7 +44,17 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const women = people.filter(person =>
+    withChildren
+      ? people.some(child => child.mother === person.name)
+      : person.sex === 'f'
+  );
+
+  const womenAges = women.map(person => person.died - person.born);
+
+  const totalAge = womenAges.reduce((age, average) => age + average);
```
33Orange, line 55: You use reduce in every function. It should be added to the `calculateAverageAge` function.

src/averageAges.js
```diff
@@ -55,7 +72,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const differenceAges = children.map(child => {
+    const mother = people.find(person => person.name === child.mother);
+
+    const differenceAge = child.born - mother.born;
+
+    return differenceAge;
+  });
+
+  const totalAge = differenceAges.reduce((age, average) => age + average);
```
33Orange, line 90: You use reduce in every function. It should be added to the `calculateAverageAge` function.

src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
+  const men = people.filter(person => (
+    person.sex === 'm'
+    && (century
+      ? century === Math.ceil(person.died / 100)
+      : true)
+  ));
+
+  const menAges = men.map(person => person.died - person.born);
+
+  const totalAge = menAges.reduce((age, average) => age + average, 0);
```
33Orange, line 27: You use reduce in every function. It should be added to the `calculateAverageAge` function.

33Orange (changes requested):
Well done ! 🔥 

Just a one moment with readability.

src/averageAges.js
```diff
@@ -55,7 +68,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      ? person.sex === 'm'
+      : true)
+  );
+
+  const differenceAges = children.map(child => {
+    const mother = people.find(person => person.name === child.mother);
+
+    const differenceAge = child.born - mother.born;
+
+    return differenceAge;
```
33Orange, line 83: You can return `child.born - mother.born` here without creating variable `differenceAge` which you instantly return

33Orange (approved):
Good job

PR #2049, opened 2023-05-18
Solution

annholovko:


src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+
+  const filteredMen = century
+    ? men.filter(man => Math.ceil(man.died / 100) === century)
+    : men;
+
+  const averageAge = filteredMen.reduce((sum, man) =>
+    sum + (man.died - man.born), 0) / filteredMen.length || 0;
```
33Orange, line 25: Create separate function `calculateAverageAge` and move this logic to this function.
Now you have the logic with reduce repeated in all 3 functions.
Make it so that the function takes an array of numbers and the length by which it needs to be divided

src/averageAges.js
```diff
@@ -37,7 +42,16 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => person.sex === 'f');
+
+  const womenWithChildren = withChildren
+    ? women.filter(woman => people.some(person => woman.name === person.mother))
+    : women;
+
+  const averageAge = womenWithChildren.reduce((sum, woman) =>
+    sum + (woman.died - woman.born), 0) / womenWithChildren.length || 0;
```
33Orange, line 52: Create separate function `calculateAverageAge` and move this logic to this function.
Now you have the logic with reduce repeated in all 3 functions.
Make it so that the function takes an array of numbers and the length by which it needs to be divided

src/averageAges.js
```diff
@@ -55,7 +69,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      : mother;
+  });
+
+  const totalAgeDiff = mothersWithChild.reduce((sum, person) => {
+    const mother = people.find(m => m.name === person.mother);
+
+    return sum + (person.born - mother.born);
+  }, 0);
+
+  const averageAgeDiff = totalAgeDiff / mothersWithChild.length || 0;
```
33Orange, line 86: Create separate function `calculateAverageAge` and move this logic to this function.
Now you have the logic with reduce repeated in all 3 functions.
Make it so that the function takes an array of numbers and the length by which it needs to be divided

33Orange (changes requested):
Good job.

A few moments:
- Create separate function for `calculateAverageAge`.
- Try filtering arrays in the `calculateMenAverageAge` and `calculateWomenAverageAge` functions using the 1st `filter` method.

For example: 
```
const men = people.filter(person => (
     person.sex === 'm' 
     && (century ?  Math.ceil(man.died / 100) === century : true)
));

```

33Orange (approved):
Well done ! 🔥

PR #2050, opened 2023-05-18
solution

kirilshorin:


src/averageAges.js
```diff
@@ -15,11 +15,24 @@
...
+
+function genderIs(person, sex) {
+  return person.sex === sex;
+}
+
+function fromCentury(person, century) {
+  return Math.ceil(person.died / 100) === century;
+}
+
+function avarageAge(people) {
```
naziamoff, line 33: - misspelling
- functions should usually start with a verb, so it was obvious that we're dealing with a function
```suggestion
function getAverageAge(people) {
```

src/averageAges.js
```diff
@@ -15,11 +15,24 @@
...
+  return avarageAge(men);
+}
+
+function genderIs(person, sex) {
+  return person.sex === sex;
+}
+
+function fromCentury(person, century) {
+  return Math.ceil(person.died / 100) === century;
+}
```
naziamoff, line 31: what does that function name mean?

src/averageAges.js
```diff
@@ -15,11 +15,24 @@
...
+  const men = people.filter(person => century
+    ? (genderIs(person, 'm') && fromCentury(person, century))
+    : (genderIs(person, 'm')));
+
+  return avarageAge(men);
+}
+
+function genderIs(person, sex) {
+  return person.sex === sex;
+}
```
naziamoff, line 27: I would say this one is redundant, it looks pretty good without this custom helper

naziamoff (changes requested):
not bad, I like this approach. However, there are major issues with the naming of your functions. Fix that and we're good

src/averageAges.js
```diff
@@ -15,11 +15,24 @@
...
+
+function getPersonCentury(person, century) {
+  return Math.ceil(person.died / 100) === century;
+}
+
+function getAvarageAge(people) {
```
dinaminahi, line 33:
```suggestion
function getAverageAge(people) {
```

dinaminahi (approved):

naziamoff (approved):

PR #2051, opened 2023-05-18
Solution

amashutikov:


src/averageAges.js
```diff
@@ -15,11 +15,23 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+  let sumOfMenAges;
+  let averageManAge;
+  const menDiedInGivenCentury = men.filter(person => (
+    Math.ceil(person.died / 100) === century
+  ));
```
naziamoff, line 23: what if a century is not passed? You still create this variable. But let's try to get the needed men right inside the filter callback

src/averageAges.js
```diff
@@ -15,11 +15,23 @@
...
+  function getAverageAge(mans) {
+    sumOfMenAges = mans.reduce((acc, person) => {
+      return acc + (person.died - person.born);
+    }, 0);
+    averageManAge = sumOfMenAges / mans.length;
+
+    return averageManAge;
+  }
+
+  return !century ? getAverageAge(men) : getAverageAge(menDiedInGivenCentury);
```
naziamoff, line 34: 1) ternaries should be multilined. It is much easier to read
```suggestion
  return !century 
    ? getAverageAge(men) 
    : getAverageAge(menDiedInGivenCentury);
```
 
 
2) you should start checking boolean values from the positive. Starting with negative is more difficult to read for the human brain 
```suggestion
  return century 
    ? getAverageAge(menDiedInGivenCentury)
    :  getAverageAge(men) ;
```

3) let's try to make this return statement work

```suggestion
  return getAverageAge(men)
```

src/averageAges.js
```diff
@@ -37,7 +49,23 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => person.sex === 'f');
+  const womanWithChild = women.filter(person => (
+    people.some(child => child.mother === person.name)
+  ));
+  let sumOfWomanAges;
+  let averageWomanAge;
```
naziamoff, line 57: same here, wrong approach. Check the comments for the section above

naziamoff (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,26 @@
...
+
+  century
+    ? averageManAge = getMenDiedInGivenCentury(men)
+      .reduce((acc, person) => {
+        return acc + (person.died - person.born);
+      }, 0) / getMenDiedInGivenCentury(men).length
+
+    : averageManAge = men.reduce((acc, person) => {
+      return acc + (person.died - person.born);
+    }, 0) / men.length;
```
silchyn, line 35: think about how to simplify this code

silchyn (changes requested):

silchyn (changes requested):

naziamoff (approved):
👍

PR #2054, opened 2023-05-18
add solution

ulositska:


src/averageAges.js
```diff
@@ -14,12 +14,19 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+
+  const menOfCentury = century
+    ? men.filter(man => Math.ceil(man.died / 100) === century)
+    : men;
```
naziamoff, line 23: let's do that inside a single filter callback

src/averageAges.js
```diff
@@ -36,8 +43,23 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const allWomen = people.filter(person => person.sex === 'f');
+
+  const mothers = people.map(person => person.mother);
+
+  const womenWithChildren = allWomen.filter(
+    woman => mothers.includes(woman.name)
+  );
+
+  const women = withChildren ? womenWithChildren : allWomen;
```
naziamoff, line 56: ternaries should be multilined
```suggestion
  const women = withChildren 
    ? womenWithChildren 
    : allWomen;
```

src/averageAges.js
```diff
@@ -36,8 +43,23 @@ function calculateMenAverageAge(people, century) {
...
+
+  const agesOfWomen = women.map(woman => woman.died - woman.born);
+
+  const sumOfAges = agesOfWomen.reduce((a, b) => (a + b), 0);
```
naziamoff, line 60: seems you use this logic a couple of times - it seems reasonable to create a helper function for that

naziamoff (changes requested):
Not bad 👍 
Still there are a couple of issues to fix

src/averageAges.js
ulositska: Totally! It looks much better with a separate function. Thx:fire:

src/averageAges.js
```diff
@@ -14,12 +14,14 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menOfCentury = century
+    ? people.filter(person => person.sex === 'm'
+      && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === 'm');
```
silchyn, line 22: > let's do that inside a single filter callback

src/averageAges.js
```diff
@@ -14,12 +14,14 @@
...
+
+  return getSumOfAges(menOfCentury) / menOfCentury.length;
```
silchyn, line 24: what if `menOfCentury.length` is 0?

src/averageAges.js
```diff
@@ -54,8 +65,31 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const allChildren = people
+    .filter(person => people
+      .some(woman => woman.name === person.mother));
+
+  const sons = allChildren.filter(child => child.sex === 'm');
+
+  const children = onlyWithSon ? sons : allChildren;
```
silchyn, line 76: can we do that inside a single `filter`?

src/averageAges.js
```diff
@@ -54,8 +65,31 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const sumOfageDifferences = ageDifferences.reduce((a, b) => (a + b), 0);
+
+  return sumOfageDifferences / ageDifferences.length;
+}
+
+function getSumOfAges(people) {
+  const sumOfAges = people
+    .map(person => person.died - person.born)
+    .reduce((a, b) => (a + b), 0);
```
silchyn, line 90: again, can we do that with a single iteration?

silchyn (changes requested):

src/averageAges.js
```diff
@@ -54,8 +66,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      : true));
+
+  const ageDifferences = children
+    .map(child => child.born - people
+      .find(mother => mother.name === child.mother).born);
+
+  return getAverageAge(ageDifferences, children.length);
+}
+
+function getAverageAge(ages, quantityOfAges) {
```
dinaminahi, line 84: We can use `ages.length` here, instead of passing additional argument

src/averageAges.js
```diff
@@ -36,8 +41,15 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => person.sex === 'f' && (withChildren
+    ? people.map(child => child.mother).includes(person.name)
```
dinaminahi, line 47: `find` method can be used instead of map and includes

dinaminahi (changes requested):
Good job 👍 
Almost there, just one moment to fix

naziamoff (approved):

silchyn (approved):

PR #2055, opened 2023-05-18
Solution

ivankhrsh:


src/averageAges.js
```diff
@@ -20,6 +20,28 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  people.forEach(function(person) {
+    person.age = person.died - person.born;
```
dinaminahi, line 24: Avoid using forEach
You can use `person.died - person.born` in your reduce method

src/averageAges.js
```diff
@@ -20,6 +20,28 @@ function calculateMenAverageAge(people, century) {
...
+  });
+
+  let men = [];
+
+  people.forEach(function(person) {
+    if (person.sex === 'm') {
+      men.push(person);
+    };
+  });
```
dinaminahi, line 33: Avoid using forEach, use `filter` instead

src/averageAges.js
```diff
@@ -20,6 +20,28 @@ function calculateMenAverageAge(people, century) {
...
+
+  if (arguments.length === 2) {
+    men = men.filter(person => (
+      Math.ceil(person.died / 100) === century));
+  }
```
dinaminahi, line 38: Let's use &&, || or ? instead of `if ()` statement

src/averageAges.js
```diff
@@ -37,7 +59,40 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  people.forEach(function(person) {
+    person.age = person.died - person.born;
+  });
```
dinaminahi, line 64: Same as above

dinaminahi (changes requested):

src/averageAges.js
ivankhrsh: I don't get how to change it. Because there I have only one action for condition true, I tried to change it and that's what I got

```
  (arguments.length === 2)
    ? men = men.filter(person => (
      Math.ceil(person.died / 100) === century))
    : **?????**;
```

src/averageAges.js
```diff
@@ -37,7 +50,22 @@ function calculateMenAverageAge(people, century) {
...
+
+    women = mothersInArr;
+  }
+
+  const ageSum = women.reduce((sum, current) => {
+    return sum + (current.died - current.born);
+  }, 0);
+
+  return ageSum / women.length;
 }
```
dinaminahi, line 69: We have same lines in all three functions, think of creating a separate function for it

src/averageAges.js
dinaminahi: If that is confusing you can move conditions to variables like: 
```
const men = people.filter(person => {
	const isMan = person.sex === 'm';
	const diedInCentury = Math.ceil(person.died / 100) === century;
	return isMan && (!century || diedInCentury);
});
```

src/averageAges.js
```diff
@@ -55,7 +83,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    if (onlyWithSon) {
+      children = children.filter(child => child.sex !== 'f');
+    }
+
+    const parentBornYear = person.born;
+
+    children.map(function(child) {
+      ageDifference.push(child.born - parentBornYear);
+    });
+  });
```
dinaminahi, line 106: we can search for all children: 
```
const children = people.filter(child => people.some(person =>
    child.mother === person.name
    && (!onlyWithSon || child.sex === 'm'))
  );
```

And then we can find for each one a mother (when mapping for ages )

dinaminahi (changes requested):

src/averageAges.js
```diff
@@ -55,7 +67,38 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    const parentBornYear = people.find(person =>
+      person.name === child.mother).born;
+
+    ageDifference.push(child.born - parentBornYear);
+  });
+
+  return calcAverageOfArrElements(ageDifference);
+}
+
+function calcAverageOfArrElements(arr, age = false) {
```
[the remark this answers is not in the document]
dinaminahi: And think of better name, getAverageAge or calcAverage

dinaminahi (changes requested):
Great job 👍 But there are still some fixes we need to do

src/averageAges.js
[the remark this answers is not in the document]
ivankhrsh: got it, thanks a lot!

src/averageAges.js
```diff
@@ -55,7 +67,30 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // write code here
+  const ageDifference = [];
+
+  const children = people.filter(child => people.some(person =>
+    child.mother === person.name
+    && (!onlyWithSon || child.sex === 'm')
+  )
+  );
+
+  children.map(function(child) {
```
dinaminahi, line 78: map is returning you a new array, so no need of pushing to a new array, just assign map to a new variable

dinaminahi: and it's better to use arrow function as callback

src/averageAges.js
```diff
@@ -55,7 +67,30 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    const parentBornYear = people.find(person =>
+      person.name === child.mother).born;
+
+    ageDifference.push(child.born - parentBornYear);
+  });
+
+  return getAverageAge(ageDifference);
+}
+
+function getAverageAge(arr) {
```
dinaminahi, line 88: avoid using types in variable's names
Instead of `arr` let it be `ages`

dinaminahi (changes requested):
Few small fixes and we are good to go

src/averageAges.js
```diff
@@ -55,7 +67,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    return (child.born - parentBornYear);
+  });
+
+  return getAverageAge(ageDifference);
+}
+
+function getAverageAge(ages) {
+  const arrSum = ages.reduce((sum, current) => {
+    return sum + current;
+  }, 0);
```
naziamoff, line 88:
```suggestion
  const arrSum = ages.reduce((sum, current) => (
    sum + current;
  ), 0);
```

src/averageAges.js
```diff
@@ -55,7 +67,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(child => people.some(person =>
+    child.mother === person.name
+    && (!onlyWithSon || child.sex === 'm')
+  ));
+
+  const ageDifference = children.map((child) => {
+    const parentBornYear = people.find(person =>
```
naziamoff, line 76:
```suggestion
    const parentBirthYear = people.find(person =>
```

src/averageAges.js
```diff
@@ -55,7 +67,27 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(child => people.some(person =>
+    child.mother === person.name
+    && (!onlyWithSon || child.sex === 'm')
+  ));
```
naziamoff, line 73: use round braces if you're moving the return statement to the new line

src/averageAges.js
```diff
@@ -37,7 +42,14 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => {
+    const someoneMothers = people.map(child => child.mother);
```
naziamoff, line 46:
```suggestion
    const mothers = people.map(child => child.mother);
```

src/averageAges.js
```diff
@@ -37,7 +42,14 @@ function calculateMenAverageAge(people, century) {
...
+
+    return person.sex === 'f' && (!withChildren
+      || someoneMothers.includes(person.name));
+  });
+
+  return getAverageAge(women.map(person => person.died - person.born));
```
naziamoff, line 52: better to move the map logic to the constant - it improves readability of your code

naziamoff (approved):
looks not bad, but there are some suggestions

naziamoff (approved):

PR #2058, opened 2023-05-19
Solution added

olehmahera:


src/averageAges.js
```diff
@@ -1,61 +1,46 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
-}
+  const men = people.filter(man => man.sex === 'm');
+  const diedInCentury = person => Math.ceil(person.died / 100) === century;
+
+  return !century
+    ? calculateAverageAge(men)
+    : calculateAverageAge(men.filter(diedInCentury));
```
naziamoff, line 14: interesting approach, seems you now really understand how callbacks work 😄 

But let's make this return statement work here:

```
return calculateAverageAge(men);
```

A hint: You can select men conditionally inside the filter callback

src/averageAges.js
```diff
@@ -1,61 +1,46 @@
...
-  // write code here
+  const women = people.filter(woman => woman.sex === 'f');
+  const mothers = people.filter(mother => (
+    people.some(child => child.mother === mother.name)
+  ));
+
+  return !withChildren
+    ? calculateAverageAge(women)
+    : calculateAverageAge(mothers);
 }
```
naziamoff, line 26: same here

naziamoff (approved):
interesting approach you got there 👍 
Check comments

PR #2062, opened 2023-05-19
Solution

k1samehoshigaki:


src/averageAges.js
```diff
@@ -55,7 +67,20 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    mother => mother.name === person.mother)
+    && (onlyWithSon ? person.sex === 'm' : true));
+
+  const ageDiff = children.map(child => {
+    const mother = people.find(person => child.mother === person.name);
+
+    return child.born - mother.born;
+  });
+
+  const avgAgeDiff = ageDiff.reduce(
```
dinaminahi, line 80: It's better to not shortcut the word in variable name, so it can be easy to read them

src/averageAges.js
```diff
@@ -37,7 +41,15 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const women = withChildren
+    ? people.filter(person => person.sex === 'f'
+      && people.some(kid => kid.mother === person.name))
+    : people.filter(person => person.sex === 'f');
+
+  const avgAge = women.reduce(
+    (prev, item) => prev + (item.died - item.born), 0) / women.length;
+
+  return avgAge;
```
dinaminahi, line 52: You could create a separate function for this, as you have similar in all functions

dinaminahi (approved):
Good job 🚀 
Check few comments

PR #2063, opened 2023-05-19
average age

diana-tuz:


src/averageAges.js
```diff
@@ -54,8 +78,35 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  women.forEach(woman => {
+    const findChildren = children.filter(child =>
+      child.mother === woman.name);
+
+    findChildren.forEach(child => {
+      const ageDiff = child.born - woman.born;
+
+      arr.push(ageDiff);
+    });
+  });
```
VitaliyBondarenko1982, line 104: it is bettter to make with map method.
but better to avoid this iteration and counit ageDiff in reduce below.

VitaliyBondarenko1982 (changes requested):

nazarmatsevych (approved):

PR #2064, opened 2023-05-20
add task solution

EugeneSherepa:


src/averageAges.js
```diff
@@ -37,7 +41,11 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => withChildren
```
33Orange, line 44: Consider breaking down the ternary operator into separate lines for better readability.

src/averageAges.js
```diff
@@ -55,7 +63,17 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(child => onlyWithSon
```
33Orange, line 66: Consider breaking down the ternary operator into separate lines for better readability.

src/averageAges.js
```diff
@@ -14,12 +14,16 @@
  *
  * @return {number}
  */
+
+const calculateAverageAge = (people) => people.reduce(
```
33Orange, line 18: It is recommended to use more meaningful function names, consider renaming 'calculateAverageAge' to 'calculateAverageAgeOfPeople'.

src/averageAges.js
```diff
@@ -14,12 +14,16 @@
...
+  (sum, person) => sum + person.died - person.born, 0) / people.length;
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => century
```
33Orange, line 22: Consider breaking down the ternary operator into separate lines for better readability.

33Orange (changes requested):
Well done ! 🔥 

A few moments:
- You can extend the `calculateAverageAge` function so that it also takes the length of the array as a parameter, so that this logic can be reused in the 3rd function.

For self-improvement:
In every `filter` you have repeated logic. If your "filtering parameter" matches in both cases, then you can use it before the ternary operator. In general, your version is also readable and understandable.

For example:

```
 const men = people.filter(person => person.sex === 'm' && (century
    ? Math.ceil(person.died / 100) === century
    : true));
```

src/averageAges.js
```diff
@@ -14,12 +14,16 @@
  *
  * @return {number}
  */
+
+const calculateAverageAge = (people) => people.reduce(
+  (sum, person) => sum + person.died - person.born, 0) / people.length;
```
silchyn, line 19: what if `people.length` is 0?

src/averageAges.js
```diff
@@ -14,12 +14,16 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => century
+    ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+    : person.sex === 'm');
```
silchyn, line 24: `person.sex === 'm'` is duplicated

silchyn (changes requested):

src/averageAges.js
```diff
@@ -55,7 +69,17 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const children = people.filter(child => onlyWithSon
+    ? child.sex === 'm' && people.some(mother => mother.name === child.mother)
+    : people.some(mother => mother.name === child.mother));
```
dinaminahi, line 74:
```suggestion
  const children = people.filter(child => people.some(mother => mother.name === child.mother) && (!onlyWithSon || child.sex === 'm' ) );
```

dinaminahi (approved):
Well done 👍

silchyn (approved):

PR #2066, opened 2023-05-20
made task js average-ages

vladyslav-lapshyn:


src/averageAges.js
```diff
@@ -15,11 +15,28 @@
...
+  // * I know, I should avoid if (), but that condition do my code more flexible
+  if (filteredMen.length === 0) {
+    return null;
+  }
+
+  const totalAgeOfMen = filteredMen.reduce((sumOfAge, person) => {
+    const { born, died } = person;
+
+    return sumOfAge + (died - born);
+  }, 0);
```
dinaminahi, line 37: Think of moving it to separate function as we have same logic in all functions

dinaminahi (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,24 @@
...
+  // * I'm checking if array have some data about people
+  // * (if we input in parameter `century` value which wasn't found)
+  // * I know, I should avoid if (), but that condition do my code more flexible
+  if (filteredMen.length === 0) {
+    return null;
+  }
+
+  const totalAgeOfMen = getTotalAge(filteredMen);
+
+  return totalAgeOfMen / filteredMen.length;
```
33Orange, line 35: You repeat this logic in every function.
Move this `totalAgeOfMen / filteredMen.length` division to a function `getTotalAge`.

Rename it to `getAverageAge` or  `calculateAverageAge`.

src/averageAges.js
```diff
@@ -15,11 +15,24 @@
...
+      ? sex === 'm' && Math.ceil(died / 100) === century
+      : sex === 'm';
+  });
+
+  // * I'm checking if array have some data about people
+  // * (if we input in parameter `century` value which wasn't found)
+  // * I know, I should avoid if (), but that condition do my code more flexible
+  if (filteredMen.length === 0) {
+    return null;
+  }
```
33Orange, line 31: You can handle it in `getAverageAge` function which will return 0 if `filteredMen.length === 0`.

src/averageAges.js
```diff
@@ -15,11 +15,24 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredMen = people.filter(person => {
+    const { sex, died } = person;
+
+    return century
+      ? sex === 'm' && Math.ceil(died / 100) === century
+      : sex === 'm';
```
33Orange, line 23:
```suggestion
    return sex === 'm' && 
    (century
      ? Math.ceil(died / 100) === century
      : true);
```

src/averageAges.js
```diff
@@ -37,7 +50,24 @@ function calculateMenAverageAge(people, century) {
...
+    // * if input `false` in `withChildren` parameter.
+    // * we have more flexible code, who have children or don't, and all women
+    return withChildren
+      ? sex === 'f' && people.some(personChild => personChild.mother === name)
+      : sex === 'f' && people.every(personChild => personChild.mother !== name);
+  });
+
+  const totalAgeOfWomen = getTotalAge(filteredWomen);
+
+  return totalAgeOfWomen / filteredWomen.length;
```
33Orange, line 70: Same as in `calculateMenAverageAge`.

Transfer the division to `getAverageAge` and return `0` from `getAverageAge` if `filteredWomen.length === 0`.

src/averageAges.js
```diff
@@ -37,7 +50,24 @@ function calculateMenAverageAge(people, century) {
...
+    if (withChildren === undefined) {
+      return sex === 'f';
+    }
+
+    // * I add it, if will need to get info about women who don't have children,
+    // * if input `false` in `withChildren` parameter.
+    // * we have more flexible code, who have children or don't, and all women
+    return withChildren
+      ? sex === 'f' && people.some(personChild => personChild.mother === name)
+      : sex === 'f' && people.every(personChild => personChild.mother !== name);
```
33Orange, line 65:
```suggestion
    return sex === 'f'  && 
      (withChildren
        ? people.some(personChild => personChild.mother === name)
        : true);
```
`people.every(personChild => personChild.mother !== name)` You don't need this logic. 
This function accept array of people, sort them by gender and return women average age. If `withChildren` not provided - we return average age of all women.

src/averageAges.js
```diff
@@ -55,7 +85,35 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const childrenList = people.filter(child => {
+    return onlyWithSon
+      ? people.some(mother => mother.name === child.mother && child.sex === 'm')
+      : people.some(mother => mother.name === child.mother);
+  });
```
33Orange, line 92:
```suggestion
  const childrenList = people.filter(child => {
    return people.some(mother => mother.name === child.mother && 
    (onlyWithSon 
      ? child.sex === 'm'
      : true)
    ) 
  });
```

src/averageAges.js
```diff
@@ -55,7 +85,35 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    if (foundMother) {
+      const ageDiff = child.born - foundMother.born;
+
+      return acc + ageDiff;
+    }
+
+    return acc;
+  }, 0);
+
+  return ageDiffSum / childrenList.length;
```
33Orange, line 108:
```suggestion
  const ageDifferences = childrenList.map(child => {
    const foundMother = people.find(mother => mother.name === child.mother);
    
    return child.born - foundMother.born;
  }, 0);

  return getAverageAge(ageDifferences, childrenList.length);
```

You iterating over the array of children that was filtered before. 
You don't need this condition.
```
    if (foundMother) {
      const ageDiff = child.born - foundMother.born;

      return acc + ageDiff;
    }
```

src/averageAges.js
```diff
@@ -55,7 +85,35 @@ function calculateWomenAverageAge(people, withChildren) {
...
+}
+
+function getTotalAge(arrayOfPeople) {
+  return arrayOfPeople.reduce((sumOfAge, person) => {
+    const { born, died } = person;
+
+    return sumOfAge + (died - born);
+  }, 0);
 }
```
33Orange, line 117: Extend you function to calculate average ages, not only total age.

```suggestion
function getAverageAge(ages, divider) {
  return ages.reduce((sum, age) => (sum + age, 0) / divider || 0;
}
```

33Orange (changes requested):
Some comments

src/averageAges.js
vladyslav-lapshyn: @33Orange to be honest it's beautiful

src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
+  const filteredMen = people
+    .filter(person => {
+      const { sex, died } = person;
+
+      return sex === 'm'
+      && (century
+        ? Math.ceil(died / 100) === century
+        : true);
+    })
+    .map(person => person.died - person.born);
```
dinaminahi, line 27: let's split this to two variables and assign this map to a different variable (`menAges` for example) as it will return just ages and it won't be suitable for `filteredMen` name anymore.

src/averageAges.js
```diff
@@ -37,7 +44,22 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filteredWomen = people
+    .filter(person => {
+      const { sex, name } = person;
+
+      if (withChildren === undefined) {
+        return sex === 'f';
+      }
```
dinaminahi, line 53: This check is redundant as the one below can handle it when `withChildren` is undefined

src/averageAges.js
```diff
@@ -37,7 +44,22 @@ function calculateMenAverageAge(people, century) {
...
+
+      return sex === 'f'
+      && (withChildren
+        ? people.some(personChild => personChild.mother === name)
+        : true);
+    })
+    .map(person => person.died - person.born);
```
dinaminahi, line 60: same here, let's split it into two variables for better readability

src/averageAges.js
```diff
@@ -55,7 +77,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const ageDifferences = childrenList.map(child => {
+    const foundMother = people.find(mother => mother.name === child.mother);
+
+    return child.born - foundMother.born;
+  });
+
+  return getAverageAge(ageDifferences, childrenList.length);
+}
+
+function getAverageAge(ages, divider) {
```
dinaminahi, line 97: You can use `ages.length` here I think, so you can omit second argument

dinaminahi (changes requested):
Good job! 👍 
Almost there, but need to do some fixes

src/averageAges.js
vladyslav-lapshyn: @dinaminahi I missed that moment, thanks

src/averageAges.js
vladyslav-lapshyn: @dinaminahi thanks, you're right, here a lot of details, easy to miss them

33Orange (approved):
Well done ! 🔥

PR #2068, opened 2023-05-21
Solution

MaksimKad:


src/averageAges.js
```diff
@@ -54,9 +66,29 @@ function calculateWomenAverageAge(people, withChildren) {
  *
  * @return {number}
  */
+
+function calculateAvg(arr) {
```
dinaminahi, line 70: It is better to not shortcut the words for naming
And think for more descriptive name for arguments (maybe `numbers`, `values`)

src/averageAges.js
```diff
@@ -37,8 +41,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
-}
+  const womanArr = people.filter(person =>
+    withChildren
```
dinaminahi, line 45: It is better to avoid using type of variable in it's name
womanArr => women

src/averageAges.js
```diff
@@ -54,9 +66,29 @@ function calculateWomenAverageAge(people, withChildren) {
  *
  * @return {number}
  */
+
+function calculateAvg(arr) {
+  const total = arr.reduce((a, b) => a + b, 0);
```
dinaminahi, line 71: And also here, total => totalSum, total => sumOfNumbers
and something more descriptive instead of a and b

dinaminahi (changes requested):
Well done 👍 
Just few moments with readability

src/averageAges.js
```diff
@@ -54,9 +66,29 @@ function calculateWomenAverageAge(people, withChildren) {
  *
  * @return {number}
  */
+
+function calculateAvg(peopleArray) {
```
33Orange, line 70: Check @dinaminahi comment.
Use more descriptive names. 
Try to describe action inside this function.
`calculateAverageAges` or `getAverageAges` would be better

src/averageAges.js
```diff
@@ -37,8 +56,15 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
-}
+  const woman = people.filter(person =>
```
33Orange, line 59: Grammar mistake
Check @dinaminahi comment
```suggestion
  const women = people.filter(person =>
```

src/averageAges.js
```diff
@@ -8,20 +8,39 @@
...
+  return (sumOfAges / peopleArray.length) || 0;
+};
+
+function calculateAgeDifference(differences) {
+  const sumOfAges = differences.reduce((acc, curVal) =>
+    acc + curVal, 0
+  );
+
+  return (sumOfAges / differences.length) || 0;
 }
```
33Orange, line 31: Do you have two almost identical helpers?

src/averageAges.js
```diff
@@ -54,9 +66,29 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
-}
+  const child = people.filter(person =>
+    onlyWithSon
+      ? people.find(mother => mother.name === person.mother)
+        && person.sex === 'm'
+      : people.find(mother => mother.name === person.mother)
+  );
```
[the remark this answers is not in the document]
33Orange: Why you ignore this comment ? @MaksimKad

src/averageAges.js
```diff
@@ -54,9 +80,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const ageDiff = child.map(children => {
+    const childMother = people.find(mother => mother.name === children.mother);
+
+    return children.born - childMother.born;
+  });
```
33Orange, line 95: Why can't you use the same `map()` method in `calculateWomenAverageAge` and `calculateMenAverageAge ` ?
This will remove the extra helper. The only one left is the one that gets an array of numbers and returns the average value

33Orange (changes requested):
Almost done.

Please check all previous comments and refactor your code

PR #2072, opened 2023-05-22
Develop

thienlaoo:


src/averageAges.js
```diff
@@ -14,12 +14,24 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredMen = century
+    ? people.filter((person) =>
+      person.sex === 'm' && century === Math.ceil(person.died / 100))
+    : people.filter((person) => person.sex === 'm');
+
+  const sumOfAgeMen = filteredMen.reduce((prev, item) =>
+    prev + (item.died - item.born), 0);
+  const averageAgeMen = sumOfAgeMen / filteredMen.length;
```
33Orange, line 32: You have helper `getAverage`, which do same logic. Why you didn't use it?

src/averageAges.js
```diff
@@ -37,7 +49,17 @@ function calculateMenAverageAge(people, century) {
...
+    person.sex === 'f' && (
+      withChildren
+        ? people.find(({ mother }) => mother === person.name)
+        : true
+    ));
+
+  const sumOfAgeWomen = filteredWomen.reduce((prev, item) =>
+    prev + (item.died - item.born), 0);
+
+  return sumOfAgeWomen / filteredWomen.length;
```
33Orange, line 62: Same as in `calculateMenAverageAge`. You have helper for this logic

33Orange (changes requested):
Almost done !

A few moments to fix

PR #2074, opened 2023-05-22
Solution

RoksolanaTylipska:


src/averageAges.js
```diff
@@ -1,5 +1,12 @@
 'use strict';
 
+const getAvarage = (data) => (
```
silchyn, line 3:
```suggestion
const getAverage = (data) => (
```

silchyn (approved):
Do not include files like `package.json` in commits

PR #2081, opened 2023-05-23
average-ages

DenisGurskiy:


src/averageAges.js
```diff
@@ -15,6 +15,15 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  const men = (arguments.length > 1)
```
vlad-demchuk, line 18: you can just check `century` instead of arguments length

src/averageAges.js
```diff
@@ -38,6 +47,18 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  let women = people.filter(obj => obj.sex === 'f');
+
+  if (arguments.length > 1) {
```
vlad-demchuk, line 52: the same

you can use ternary here as well

src/averageAges.js
```diff
@@ -38,6 +47,18 @@ function calculateMenAverageAge(people, century) {
...
+    women = women.filter((obj) => {
+      return people.some((elem) => elem.mother === obj.name);
+    });
+  };
+
+  const age = women.map(obj => obj.died - obj.born);
+  const sum = age.reduce((a, b) => a + b, 0);
+
+  return sum / women.length;
```
vlad-demchuk, line 61: create separate helper function to calculate average age

you repeat it several time

src/averageAges.js
```diff
@@ -56,6 +77,38 @@ function calculateWomenAverageAge(people, withChildren) {
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
   // write code here
+  const result = [];
+  let sum = 0;
+
+  if (arguments.length > 1) {
```
vlad-demchuk, line 83: the same

src/averageAges.js
```diff
@@ -56,6 +77,38 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    people.forEach((elem) => {
+      for (const elem2 of people) {
```
vlad-demchuk, line 85: Instead of using a combination of forEach and for loops, use array methods like .filter and .map

vlad-demchuk (changes requested):

src/averageAges.js
```diff
@@ -56,6 +78,37 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      people.filter(obj => obj.mother === elem.name).map(kid => {
+        result.push({
+          mother: elem,
+          kid: kid,
+        });
+      });
+    });
+  }
+
+  for (let i = 0; i < result.length; i++) {
```
volodymyr-soltys97, line 107: Replace the for loop with a more functional approach, like using .reduce() to calculate the sum.

volodymyr-soltys97 (approved):
Good job 👍

PR #2084, opened 2023-05-26
add solution

dimarogkov:


src/averageAges.js
```diff
@@ -15,11 +18,16 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const manArr = century
+    ? people.filter(
+        (item) => item.sex === "m" && Math.ceil(item.died / 100) === century
+      )
+    : people.filter((item) => item.sex === "m");
```
Viktor-Kost, line 25: you can apply ternary operator inside filter method
```suggestion
  const manArr = people.filter(
    (item) => item.sex === "m" && (!century || Math.ceil(item.died / 100) === century)
  );
```

src/averageAges.js
```diff
@@ -37,7 +45,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenArr = withChildren
```
Viktor-Kost, line 48: the same with ternary

Viktor-Kost (changes requested):
GJ!
Minor improvements can be applied

OleziO (approved):

PR #2086, opened 2023-05-26
solution

Vepr984:


src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+
+  if (century) {
```
mykhalenych: missed requirement  'replace `if ()` statement with &&, || or ?'

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
+    const filteredMen
+      = men.filter(person => Math.ceil(person.died / 100) === century);
+
+    return filteredMen
+      .reduce((sum, person) =>
+        sum + (person.died - person.born), 0) / filteredMen.length;
```
mykhalenych: create separate function with calculate average age and pass argument array that you need to calculate

src/averageAges.js
```diff
@@ -36,8 +44,18 @@ function calculateMenAverageAge(people, century) {
...
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = withChildren
+    ? people.filter(person => person.sex === 'f'
+      && people.some(child => child.mother === person.name))
+    : people.filter(person => person.sex === 'f');
+
+  const womenAgeArr = women.map(person => person.died - person.born);
+  const sum = womenAgeArr.reduce((sums, age) => sums + age, 0);
```
mykhalenych: and reuse this function for calculate average age here

mykhalenych (changes requested):

mykhalenych (changes requested):
I don't see any changes from previous review, did you push code?

Vepr984:
sorry I forgot to push the code (
I done

src/averageAges.js
```diff
@@ -36,8 +44,16 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = withChildren
+    ? people.filter(person => person.sex === 'f'
```
etojeDenys: You can simplify the filtering logic by combining it with the initial 'women' filtering. Use a single filter method with a condition that checks both if the person is female and if they have children.

src/averageAges.js
```diff
@@ -55,7 +71,17 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const child = people.filter(person => onlyWithSon
+    ? person.sex === 'm' && people.some(male => male.name === person.mother)
+    : people.some(male => male.name === person.mother));
+
+  const ages = child.map(person => {
+    const mom = people.filter(male => male.name === person.mother);
```
etojeDenys: You can use the 'find' method instead of 'filter' for finding the mother, as it returns the first element that matches the condition.

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+
+  const filteredMen
+    = century && men.filter(person => Math.ceil(person.died / 100) === century);
```
etojeDenys: You can simplify the filtering logic by combining it with the initial 'men' filtering. Use a single filter method with a condition that checks both if the person is male and if their century of death matches the given century.

etojeDenys (changes requested):

IvanFesenko (approved):

PR #2087, opened 2023-05-26
averageAges complite

hyrych-max:


src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menArray = century
+    ? people.filter(person => Math.ceil(person.died / 100) === century
+      && person.sex === 'm')
+    : people.filter(person => person.sex === 'm');
```
polosanya, line 21: You filter people in both cases. So you can move ternary operator inside filter callback to make the code a bit cleaner

src/averageAges.js
```diff
@@ -37,7 +42,16 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenArray = withChildren
+    ? people.filter(person => person.sex === 'f'
+      && people.some(child => child.mother === person.name))
+    : people.filter(person => person.sex === 'f');
+
+  const womenAges = womenArray.map(person => person.died - person.born);
+  const sumOfMenAges = womenAges.reduce((sum, age) => sum + age, 0);
```
polosanya, line 51: Are you sure that it's sumOf**Men**Ages?
By the way, you have the same logic for men and women. It would be great to create a function helper for calculating an average age of the array

polosanya (changes requested):
Almost done!

sTorba24 (approved):
GJ 🚀

PR #2088, opened 2023-05-27
task solved

Nazarht:


src/averageAges.js
```diff
@@ -14,12 +14,26 @@
...
+
+  let yearsArr;
+
+  if (century) {
+    yearsArr = manArr
+      .filter(person => Math.ceil(person.died / 100) === century)
+      .map(person => +person.died - +person.born);
+  } else {
+    yearsArr = manArr.map(person => +person.died - +person.born);
+  };
```
vovakpro13, line 34: move map(person => +person.died - +person.born) into a function to not repeat yourself.
Use destructure assignment

src/averageAges.js
```diff
@@ -14,12 +14,26 @@
...
+
+  return +(yearsArr.reduce((a, b) => a + b, 0) / yearsArr.length).toFixed(2);
```
vovakpro13, line 36: move into a func and reuse throughout your code

src/averageAges.js
```diff
@@ -56,6 +85,38 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  let ageDiffArr;
+
+  if (onlyWithSon) {
+    const manArr = people.filter(person => person.sex === 'm');
+
+    ageDiffArr = manArr.map(person => {
+      const motherArr = people.filter(el => el.name === person.mother);
+      const mother = motherArr[0];
+
+      if (mother) {
```
vovakpro13, line 97: according to the task rules, you should use only ternary operators instead of if else

src/averageAges.js
```diff
@@ -56,6 +85,38 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    });
+  } else {
+    ageDiffArr = people.map(person => {
+      const motherArr = people.filter(el => el.name === person.mother);
+      const mother = motherArr[0];
+
+      if (mother) {
+        return (person.born - mother.born);
+      }
+    });
```
vovakpro13, line 109: almost the same piece of code as in the 'if' block, try to minimize it

vovakpro13 (changes requested):

Volodymyr-Melnyk-Fron10der (changes requested):
--

src/averageAges.js
```diff
@@ -14,12 +14,39 @@
...
+const getYearsArr = (arr) => {
+  return arr.map(person => {
+    const { died, born } = person;
+
+    return +died - +born;
+  });
+};
+
 function calculateMenAverageAge(people, century) {
   // write code here
```
OleziO, line 31: Remove all comments from the code.

src/averageAges.js
OleziO: Not resolved.

OleziO (changes requested):

src/averageAges.js
```diff
@@ -14,12 +14,39 @@
...
+  let yearsArr;
+
+  if (century) {
+    const filteredArr = manArr
+      .filter(person => Math.ceil(person.died / 100) === century);
+
+    yearsArr = getYearsArr(filteredArr);
+  } else {
+    yearsArr = getYearsArr(manArr);
+  };
```
nazarmatsevych, line 47:
```suggestion
  const manArr = people.filter(person =>
    century
      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
      : person.sex === 'm'
  );
  
 let yearsArr = getYearsArr(manArr);
```

Use ternary instead of if/else statements.

Fix all such cases

nazarmatsevych (changes requested):

volodymyr-soltys97 (approved):
Good work 👍

PR #2089, opened 2023-05-27
solution

Gosyav:


src/averageAges.js
```diff
@@ -15,11 +15,28 @@
...
+    }
+
+    default: {
+      menArray = people.filter(person => {
+        const personLiveCentury = Math.ceil(person.died / 100);
+
+        return person.sex === 'm' && personLiveCentury === century;
+      });
+    }
+  }
```
VitaliyBondarenko1982, line 33: redundant to make switch/case for two cases
it is can be ternary in filter callback

src/averageAges.js
```diff
@@ -37,7 +54,27 @@ function calculateMenAverageAge(people, century) {
...
+      break;
+    }
+
+    default: {
+      womenArray = people.filter(person => {
+        return person.sex === 'f'
+        && people.some(child => child.mother === person.name);
+      });
+    }
+  }
```
VitaliyBondarenko1982, line 71: same about switch/case as comment above

VitaliyBondarenko1982 (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,21 @@
...
-  // without nesting
+  let menArray = people.filter(person => person.sex === 'm');
+
+  if (century) {
+    menArray = people.filter(person => {
+      const personLiveCentury = Math.ceil(person.died / 100);
+
+      return person.sex === 'm' && personLiveCentury === century;
+    });
+  }
```
VitaliyBondarenko1982, line 26:
```suggestion
  const menArray = people.filter(person => century ? person.sex === 'm' : Math.ceil(person.died / 100) && person.sex === 'm' );
```
it can be shorter.
same for calculateWomenAverageAge

VitaliyBondarenko1982 (approved):

IvanFesenko (approved):

PR #2094, opened 2023-06-01
add task solution

sasdaiv:


src/averageAges.js
```diff
@@ -15,29 +15,33 @@
...
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = withChildren
+    ? people.filter(person => person.sex === 'f'
+      && people.some(child => child.mother === person.name))
+    : people.filter(person => person.sex === 'f');
+  const averageAge
+    = women.reduce((sum, person) => sum + (person.died - person.born), 0);
```
polosanya, line 42: It would be great to create a function-helper for calculating an average age in order not to duplicate the same logic

src/averageAges.js
```diff
@@ -15,29 +15,33 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = century
+    ? people.filter(person => Math.ceil(person.died / 100) === century
+      && person.sex === 'm')
+    : people.filter(person => person.sex === 'm');
```
polosanya, line 21: In both cases, you call filter method. You can move the ternary operator inside filter callback to make it a bit cleaner

polosanya (changes requested):
Good job! Let's improve your code a bit)

src/averageAges.js
```diff
@@ -55,7 +56,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const averageAge = children.reduce((sum, child) => {
+    const mother = people.find(woman => woman.name === child.mother);
+
+    return sum + ((child.born - mother.born));
+  }, 0) / children.length;
+
+  return averageAge;
+}
+
+function calcAverAge(objects) {
```
polosanya, line 71:
```suggestion
function calculateAverageAge(people) {
```
It is better to use full names for functions and meaningful names for parameters

src/averageAges.js
```diff
@@ -55,7 +56,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const averageAge
+    = objects.reduce((sum, person) => sum + (person.died - person.born), 0);
```
polosanya, line 73:
```suggestion
  const averageAge = objects
    .reduce((sum, person) => sum + (person.died - person.born), 0);
```

polosanya (approved):
Nice!

PR #2101, opened 2023-06-07
SolutionAverageAges

olena-bielozorova:


src/averageAges.js
```diff
@@ -15,11 +15,18 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let countOfMen = 0;
```
marina-bidenko, line 18: Avoid using 'let', consider using 'const' for variables that will not be reassigned.

marina-bidenko (approved):
LGTM

PR #2102, opened 2023-06-07
Solution

denlubn:


src/averageAges.js
```diff
@@ -20,6 +20,16 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const menArray = people.filter(
+    person => person.sex === 'm'
+      && (!century || Math.ceil(person.died / 100) === century)
```
vlad-demchuk, line 25: you could just use ternary operator to check in filter callback if century exists

src/averageAges.js
```diff
@@ -37,7 +47,17 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenArray = people.filter(
+    (person, index, arr) => {
+      return person.sex === 'f'
+        && (!withChildren || arr.some(item => item.mother === person.name));
```
vlad-demchuk, line 53: the same

src/averageAges.js
vlad-demchuk: it will be better for performance and clarity

src/averageAges.js
```diff
@@ -37,7 +47,17 @@ function calculateMenAverageAge(people, century) {
...
+    });
+
+  const years = womenArray.map(person => person.died - person.born);
+
+  const sumOfAge = years.reduce((sum, age) => sum + age, 0);
+
+  return sumOfAge / years.length;
```
vlad-demchuk, line 60: create a separate helper function to calculate average age
you do the same several times

vlad-demchuk (changes requested):

src/averageAges.js
```diff
@@ -20,6 +20,14 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const menArray = people.filter(
+    person => person.sex === 'm'
```
anastasiiavorobiova, line 24: 'm' is a magic string. It's a common antipattern. Consider using variables for this with meaningful names.

src/averageAges.js
```diff
@@ -20,6 +20,14 @@ function calculateMenAverageAge(people, century) {
...
+      && (century ? Math.ceil(person.died / 100) === century : true)
```
anastasiiavorobiova, line 25: 100 is a magic number. See my comment above

src/averageAges.js
```diff
@@ -37,7 +45,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenArray = people.filter(
+    (person, index, arr) => {
+      return person.sex === 'f'
```
anastasiiavorobiova, line 50: There is a repeating logic for sex checks. Consider using the function isMan or something similar, which would check the sex of the person and would return true/false values.

src/averageAges.js
```diff
@@ -55,7 +72,21 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  // const differences = people.map(person => people.find(person.mother));
+  const children = people.filter(person =>
+    (!onlyWithSon || person.sex === 'm')
```
anastasiiavorobiova, line 77: See my previous comments

anastasiiavorobiova (changes requested):
Great job!
Added some comments here

src/averageAges.js
```diff
@@ -55,7 +73,21 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  // const differences = people.map(person => people.find(person.mother));
```
volodymyr-soltys97, line 76: Remove the commented out line of code.

src/averageAges.js
```diff
@@ -37,7 +45,17 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenArray = people.filter(
+    (person, index, arr) => {
```
volodymyr-soltys97, line 49: Consider using implicit return and removing unnecessary parameters 'index' and 'arr' from the filter function.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

PR #2109, opened 2023-06-13
Solution

MykolaLopushenko:


src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
+  const men = people.filter(
+    person => person.sex === 'm'
+      && (!century || century === Math.ceil(person.died / 100))
+  );
+
+  const menAge = men.map(person => person.died - person.born);
+
+  const sumMenAge = menAge.reduce((prev, next) => prev + next, 0);
+
+  return sumMenAge / menAge.length;
```
IvanFesenko, line 27: Create separate function-helper with this functionality and use it here and in `calculateWomenAverageAge`. Becase you have same code - do not repeat your self

src/averageAges.js
```diff
@@ -55,7 +71,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    return children.find(child => {
+      if (child.mother === mother.name) {
+        difference.push(child.born - mother.born);
+      }
+    });
+  });
+
+  return difference.reduce((prev, next) => (
+    prev + next
+  ), 0) / difference.length;
```
IvanFesenko, line 95: use reduce instead forEach and separate variable + reduce. Also you do it in one reduce call

IvanFesenko (changes requested):
Check my commnets

etojeDenys (approved):
lgtm

PR #2117, opened 2023-06-15
task solution

Bonum-taurus:


src/averageAges.js
```diff
@@ -15,11 +15,21 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const mens = people.filter(person => person.sex === 'm');
```
anastasiiavorobiova, line 18: * Avoid using magic strings here, consider using variables with meaningful names
* Consider writing the check of the sex to separate function isMan() and reusing it further in your code

src/averageAges.js
```diff
@@ -15,11 +15,21 @@
...
+
+  if (century) {
+    const mensInCentury = mens.filter(men =>
+      Math.ceil(men.died / 100) === century);
```
anastasiiavorobiova, line 22: Avoid using magic numbers. It's an antipattern

src/averageAges.js
```diff
@@ -55,7 +83,45 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const womens = people.filter(person => person.sex === 'f');
```
anastasiiavorobiova, line 86: * Avoid magic strings
* Avoid repeating code for sex check

anastasiiavorobiova (changes requested):
Great start! Let's make your code perfect

src/averageAges.js
```diff
@@ -55,7 +97,45 @@ function calculateWomenAverageAge(people, withChildren) {
...
+        const diffAges = person.born - mother.born;
+
+        differences.push(diffAges);
+      }
+    }
+  }
+
+  const sumDifferences = differences.reduce((prev, diff) => prev + diff, 0);
+
+  return sumDifferences / differences.length;
```
volodymyr-soltys97, line 138: Also you can create a helper function for return and use it everywhere

PR #2118, opened 2023-06-15
solution

Krivets:


src/averageAges.js
```diff
@@ -1,61 +1,69 @@
...
+  const men = people.filter(person => century
+    ? Math.ceil(person.died / 100) === century
+      && person.sex === 'm'
+    : person.sex === 'm');
+
+  const menSum = men.reduce(
+    (sum, person) => sum + (person.died - person.born), 0);
+  const average = menSum / men.length;
+
+  return average;
```
agniya-i, line 19: This logic  for calculating average age is common for several functions, it would be nice to create a separate function to calculate this value to not repeat

agniya-i (changes requested):

src/averageAges.js
```diff
@@ -1,61 +1,69 @@
...
-  // without nesting
+  const men = people.filter(person => century
+    ? Math.ceil(person.died / 100) === century
+      && person.sex === 'm'
+    : person.sex === 'm');
+
+  return getAverage(men);
+}
+
+function getAverage(personArr) {
```
agniya-i, line 18:
```suggestion
function getAverageAge(people) {
```

agniya-i (approved):

PR #2124, opened 2023-06-16
finished task

artkas03:


src/averageAges.js
```diff
@@ -15,11 +15,32 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let totalAge = 0;
+  let fitManCounter = 0;
+
+  people.forEach(person => {
```
anstsot, line 21: try to use the array methods like `map`, `reduce`, `filter` etc.

anstsot (changes requested):
Let's rewrite the code by using just learned array methods

src/averageAges.js
artkas03: > try to use the array methods like `map`, `reduce`, `filter` etc.

remade but am i wrong with fact that this makes perfomance worse?

anya-shults (approved):
Good job :)

PR #2125, opened 2023-06-16
Solution

dmytro-popovych:


src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people
+    .filter(person => person.sex === 'm' && (
+      !century || Math.ceil(person.died / 100) === century));
+
+  const totalAge = men
+    .reduce((prev, man) => prev + (man.died - man.born), 0);
+
+  const averageAge = totalAge / men.length;
```
sTorba24, line 25: create separated function for it as it used twice

src/averageAges.js
```diff
@@ -55,7 +69,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+    return child.born - mother.born;
+  });
+
+  const totalAge = ageDifference
+    .reduce((prev, difference) => prev + difference, 0);
+
+  const averageAge = totalAge / ageDifference.length;
+
+  return averageAge;
```
sTorba24, line 88:
```suggestion
  return totalAge / ageDifference.length;;
```
not needed to create additional variable here :)

src/averageAges.js
```diff
@@ -37,7 +42,16 @@ function calculateMenAverageAge(people, century) {
...
+  const women = people
+    .filter(woman => woman.sex === 'f' && (!withChildren || people
+      .find(child => child.mother === woman.name)));
+
+  const totalAge = women
+    .reduce((prev, woman) => prev + (woman.died - woman.born), 0);
+
+  const averageAge = totalAge / women.length;
+
+  return averageAge;
```
sTorba24, line 54: same

src/averageAges.js
sTorba24: not needed to create variable

sTorba24 (changes requested):
Well done! Check my comments and fix it

SanyaBratashchuk (approved):

PR #2128, opened 2023-06-17
Solution

SerhiiVoitiuk:


src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menAges = people
+    .filter(man => {
+      if (century) {
```
TarasHoliuk, line 20: don't use `if`:
image

src/averageAges.js
```diff
@@ -55,7 +69,21 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const filteredPeople = people.filter(person => (
```
TarasHoliuk, line 72: rename this variable to explain which kind of people it stores (children, etc.)

src/averageAges.js
```diff
@@ -55,7 +69,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    people.find(mother => mother.name === person.mother) && (onlyWithSon
```
TarasHoliuk, line 73: you use 2 different ways to write code to do the same things here and in `calculateWomenAverageAge` function (I mean handle optional flag in the filter). Both are good IMO (:+1: ) but choose one and use it everywhere to be consistent

src/averageAges.js
```diff
@@ -55,7 +69,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      ? person.sex === 'm'
+      : true)
+  ));
+
+  const diffArr = filteredPeople.map(person =>
```
TarasHoliuk, line 78: in this case, it's clear what is it, but it's better not to use short forms like `diffArr`

src/averageAges.js
TarasHoliuk: Also, `some` method fits here better than `find` - returns boolean

TarasHoliuk (changes requested):
GJ :+1: 
I left a few comments - no one critical but is good to know it and fix

src/averageAges.js
```diff
@@ -15,11 +15,13 @@
  * @return {number}
  */
```
OleziO, line 16: Remove all comments from the code.

OleziO (changes requested):

demetrxx (approved):
Well done 👍 Go ahead!

PR #2129, opened 2023-06-17
Solution

anya-shults:


src/averageAges.js
```diff
@@ -14,12 +14,18 @@
...
-function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+function calculateMenAverageAge(people, century = null) {
+  let menOnly = filterPersonSex(people, 'm');
+
+  if (century !== null) {
```
GGsmou, line 20: I think it could be better to rewrite `if` to `ternary operator` to avoid nesting, but there is nothing wrong with this solution

src/averageAges.js
```diff
@@ -36,8 +42,16 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
-function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+function calculateWomenAverageAge(people, withChildren = false) {
+  let womenOnly = filterPersonSex(people, 'f');
+
+  if (withChildren) {
```
GGsmou, line 48: Same with `ternary operator` here

src/averageAges.js
```diff
@@ -14,12 +14,18 @@
...
  * @return {number}
  */
-function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+function calculateMenAverageAge(people, century = null) {
+  let menOnly = filterPersonSex(people, 'm');
```
GGsmou, line 18: There is no need to use word `Only`

src/averageAges.js
```diff
@@ -36,8 +42,16 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
-function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+function calculateWomenAverageAge(people, withChildren = false) {
+  let womenOnly = filterPersonSex(people, 'f');
```
GGsmou, line 46: There is no need to use word `Only`

src/averageAges.js
```diff
@@ -54,8 +68,53 @@ function calculateWomenAverageAge(people, withChildren) {
...
-function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+function calculateAverageAgeDiff(people, onlyWithSon = false) {
+  let children = people.filter(person => person.mother !== null);
+
+  if (onlyWithSon) {
+    children = filterPersonSex(children, 'm');
+  }
+
+  const womenOnly = filterPersonSex(people, 'f');
```
GGsmou, line 78: There is no need to use word `Only`
Also, there is no need to filtrate by sex and after make a second filtration by name

src/averageAges.js
```diff
@@ -54,8 +68,53 @@ function calculateWomenAverageAge(people, withChildren) {
  *
  * @return {number}
  */
-function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+function calculateAverageAgeDiff(people, onlyWithSon = false) {
+  let children = people.filter(person => person.mother !== null);
+
+  if (onlyWithSon) {
```
GGsmou, line 74: Same with `ternary operator` here

src/averageAges.js
```diff
@@ -54,8 +68,53 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    children = filterPersonSex(children, 'm');
+  }
+
+  const womenOnly = filterPersonSex(people, 'f');
+  const mothers = filterMothersNames(womenOnly, getMothersNames(children));
+
+  children = children.filter(child => (
```
GGsmou, line 81: Same double filtration problem here

src/averageAges.js
```diff
@@ -54,8 +68,53 @@ function calculateWomenAverageAge(people, withChildren) {
...
+}
+
+function sumAge(people) {
+  return people.reduce((sum, person) => (
+    sum + (person.died - person.born)
+  ), 0);
+}
+
+function calculateAverageAge(calculatedAge, people) {
+  return Math.round(calculatedAge / people.length * 100) / 100;
```
GGsmou, line 101: There was no task to `round` average

src/averageAges.js
```diff
@@ -54,8 +68,53 @@ function calculateWomenAverageAge(people, withChildren) {
...
+}
+
+function getMothersNames(people) {
```
GGsmou, line 104: I like how you broke some functionality into different functions, but I affraid it would be better to simply reuse some code, especially with `sumAgeDiff` because it used only once

src/averageAges.js
```diff
@@ -54,8 +68,53 @@ function calculateWomenAverageAge(people, withChildren) {
...
+function filterMothersNames(mothers, mothersNames) {
+  return mothers.filter(mother => mothersNames.includes(mother.name));
+}
+
+function sumAgeDiff(children, mothers) {
+  return children
+    .map(child => (
+      child.born - mothers.find(mother => mother.name === child.mother).born)
+    )
+    .reduce((sum, age) => sum + age, 0);
```
GGsmou, line 117: It wasn`t necessary to map and reduce, it could be nested into reduce

GGsmou (approved):
Overall, I like how you broke your code into different functions
I found some code quality problems, that are not critical(so `approving` for now), but I strongly advise considering them in future to avoid refactoring

*(I could be wrong, so would be glad to hear mentor's thoughts)*

tetiana-ivanchenko (approved):

PR #2132, opened 2023-06-17
Solution

s3rg4sh:


src/averageAges.js
```diff
@@ -1,61 +1,52 @@
...
- * her as mother.
- *
- * @param {object[]} people
- * @param {boolean} withChildren - optional
- *
- * @return {number}
- */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filtered = people.filter(withChildren !== undefined
```
etojeDenys, line 19: what if someone will pass a false here 
image
```suggestion
  const filtered = people.filter(!!withChildren
```

src/averageAges.js
```diff
@@ -1,61 +1,52 @@
...
- * @param {boolean} onlyWithSon - optional
- *
- * @return {number}
- */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  let count = 0;
+
+  const childs = people.filter(
+    onlyWithSon !== undefined
```
etojeDenys, line 37: same here

src/averageAges.js
```diff
@@ -1,61 +1,52 @@
...
+      ? person => person.mother !== null && person.sex === 'm'
+      : person => person.mother !== null);
+
+  const difference = childs.reduce((prev, child) => {
+    const mother = people.find(person => person.name === child.mother);
+
+    mother !== undefined ? count++ : count += 0;
+
+    return mother === undefined ? prev : prev + (child.born - mother.born);
```
etojeDenys, line 46: better to rewrite it with if
```
if(!mother) {
 return prev
} 

...

```

etojeDenys (changes requested):

src/averageAges.js
```diff
@@ -1,61 +1,56 @@
...
+  const difference = childs.reduce((prev, child) => {
+    const mother = people.find(person => person.name === child.mother);
+
+    mother !== undefined ? count++ : count += 0;
+
+    if (!mother) {
+      return prev;
+    }
+
+    return prev + (child.born - mother.born);
```
etojeDenys, line 50:
```suggestion
    if (!mother) {
      return prev;
    }
    
    count++
    return prev + (child.born - mother.born);
```

etojeDenys (changes requested):

etojeDenys (approved):
LGTM

PR #2135, opened 2023-06-19
Added task solution

ruslan-ihnatenko:


src/averageAges.js
```diff
@@ -1,61 +1,56 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter((
+    {
+      died,
+      sex,
+    }) => century ? sex === 'm'
+  && Math.ceil(died / 100) === century : sex === 'm');
```
lerastarynets, line 9: The callback in `.filter` is reusable. I think it's better to move it to separate `const` like this
`const isMan = (sex) => sex === 'm';` and reuse it like this `people.filter(({sex, died} => isMan(sex))`
This makes your code easier to read and understand for other developers

src/averageAges.js
```diff
@@ -1,61 +1,56 @@
...
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter((
+    {
+      name,
+      sex,
+    }) => withChildren ? sex === 'f'
+  && people.some(kid => kid.mother === name) : sex === 'f');
```
lerastarynets, line 26: Same for callback in `.filter`.
`const isWoman = (sex) => sex === 'f';`....

src/averageAges.js
```diff
@@ -1,61 +1,56 @@
...
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const childWithMother = people.filter((
+    {
+      sex,
+      mother,
+    }) => onlyWithSon ? sex === 'm' && mother !== null
+    && people.find(parent => parent.name === mother) !== undefined
```
lerastarynets, line 44: try using `array.some` method here intead of `array.filter !== undefined`

src/averageAges.js
```diff
@@ -1,61 +1,56 @@
...
  *
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const childWithMother = people.filter((
+    {
+      sex,
+      mother,
+    }) => onlyWithSon ? sex === 'm' && mother !== null
```
lerastarynets, line 43: no need to write `&& mother !== null`, could be just `&& mother`

In JavaScript, the `!==` operator checks for strict inequality, meaning it checks both the value and the type of the operands. When comparing a variable to null, the `!==` operator ensures that the variable is not only not equal to `null` in value but also not of the type `null`.

However, since `null` is a falsy value in JavaScript, you can simply use something in a conditional statement to check if it evaluates to a truthy value. If something is not `null`, it will be considered truthy and pass the conditional check.

So, if you only need to check if a value is truthy (i.e., not null, undefined, false, 0, an empty string, or NaN), you can use `if (something)` instead of `if (something !== null)`. It simplifies the code and conveys the same meaning.

src/averageAges.js
```diff
@@ -1,61 +1,56 @@
...
+    && people.find(parent => parent.name === mother) !== undefined
+    : mother !== null
+    && people.find(parent => parent.name === mother) !== undefined);
```
lerastarynets, line 46: Also pls try moving `parent => parent.name === mother` to a reusable func like I suggested for previous functions

lerastarynets (approved):
All requirements are met) Pls take my comments into account to improve your code

PR #2137, opened 2023-06-19
Solution

valet2102:


src/averageAges.js
```diff
@@ -20,6 +20,22 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  // const MEN = people.filter(person => century
+  //   ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+  //   : person.sex === 'm'
+  // );
+
+  const MEN = century
```
FrontKid, line 28: it's better to rename these variables to lowercase, basically we use upper + snake keys for static variables that won't be changed

FrontKid (changes requested):

PR #2138, opened 2023-06-19
Solution

dianadiv:


src/averageAges.js
```diff
@@ -54,8 +68,32 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // write code here
+  const children = people.filter(person => (
+    person.mother !== null
+    && (!onlyWithSon || person.sex === 'm')
+  ));
+
+  let amountOfMothers = 0;
+
+  const sumOfAges = children.reduce((sum, person) => {
+    const mother = people.find(personOfPeople => (
```
ivanbilotserkivskiy: Good solution. I suppose you can avoid `if` statement here. Just make mother search logic before counting total age and you will be able to simplify reduce method.🙂

ivanbilotserkivskiy (approved):
GJ! Excellent work with callback functions  👍

src/averageAges.js
YegorKorenkov: Good tip! Please firstly find all people with mother and after use reduce method for calculate average age

YegorKorenkov (changes requested):

src/averageAges.js
```diff
@@ -54,8 +68,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    return people.some((child) => child.mother === person.name);
+  });
+
+  const children = people.filter(({ mother, sex }) => {
+    return people.some(({ name }) => {
+      if (onlyWithSon) {
+        return mother === name && sex === 'm';
+      }
+
+      return mother === name;
```
anstsot, line 83: can be simplified
```suggestion
    return people.some(({ name }) => {
      return mother === name && onlyWithSon ?  sex === 'm' : true;
```

anstsot (approved):
Good work, approving:sparkles:

PR #2139, opened 2023-06-19
averageAges solution

ivanbilotserkivskiy:


src/averageAges.js
```diff
@@ -55,7 +77,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    ? people.filter(person => person.sex === 'm')
+    : people;
+
+  const ageOdds = filteredPeople.map(human => {
+    const mother = people.find(person => person.name === human.mother);
+
+    return mother && human.born - mother.born;
+  }).filter(person => person);
+
+  const totalAgeOdds = ageOdds.reduce((accumulator, currentDiference) => {
```
tetianakostiuk1: there's no need for the return
```js
  const totalAgeOdds = ageDifferences.reduce((accumulator, currentDifference) =>
    accumulator + currentDifference, 0);

File: src/averageAges.js [hunk 2]
```diff
@@ -55,7 +77,21 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const filteredPeople = onlyWithSon
+    ? people.filter(person => person.sex === 'm')
+    : people;
+
+  const ageOdds = filteredPeople.map(human => {
```
tetianakostiuk1: maybe ageDifferences will be more descriptive

tetianakostiuk1 (approved):
Nice job!😊

src/averageAges.js
```diff
@@ -20,6 +20,17 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
```
YegorKorenkov: Please remove all comments in each function

src/averageAges.js
```diff
@@ -20,6 +20,17 @@ function calculateMenAverageAge(people, century) {
...
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  let men = people.filter(person => person.sex === 'm');
+
+  men = century
```
YegorKorenkov: You can move this condition to filter method. 
`people.filter(person => person.sex === 'm' && (century ? Math.ceil(person.died / 100) === century : true));`

src/averageAges.js
```diff
@@ -20,6 +20,17 @@ function calculateMenAverageAge(people, century) {
...
+    ? men.filter(man => Math.ceil(man.died / 100) === century)
+    : men;
+
+  const totalAge = men.reduce((accumulator, man) => {
```
YegorKorenkov: Please move to separate function, because this statement used miltiply times
image

src/averageAges.js
YegorKorenkov: Good tip, please fix!

src/averageAges.js
YegorKorenkov: Why u use map method, if you can immediately use filter?

YegorKorenkov (changes requested):

YegorKorenkov (approved):
Good job!

PR #2141, opened 2023-06-19
added task solution

tetianakostiuk1:


NazarBaraban (approved):
Good job! Interesting solution

src/averageAges.js
```diff
@@ -15,13 +15,13 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
-}
+  const male = people
+    .filter((person) =>
+      person.sex === 'm'
```
juljaVysotska, line 20: Use `person =>` instead of `(person) =>` for consistency with other arrow functions in the code

src/averageAges.js
```diff
@@ -54,8 +61,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  return ageDifferences
+    .reduce((sum, age) => sum + age, 0) / ageDifferences.length;
+}
+
+function calculateAverageAge(people) {
+  const averageAge = people.reduce(
+    (sum, person) => sum + (person.died - person.born), 0) / people.length;
+
+  return +averageAge.toFixed(2);
```
juljaVysotska, line 82: Consider removing the unary plus operator (+) before averageAge.toFixed(2) since the return type of the function is a number, and it's not necessary to explicitly convert it

juljaVysotska (approved):
Good job!

PR #2142, opened 2023-06-19
Solution

yehor-kor:


src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // without nesting
+  const men = people.filter((person) => (
+    century
+      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+      : person.sex === 'm'
+  ));
+
+  return men.reduce((sum, man) => (
+    sum + man.died - man.born
+  ), 0) / men.length;
```
polosanya, line 26: It would be great to create a function-helper for calculation the average age of people

polosanya (approved):
Good job!

PR #2143, opened 2023-06-19
Solution

ArtemBehter:


src/averageAges.js
```diff
@@ -15,11 +15,25 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  for (const person of people) {
```
darokrk: Avoid using for loop, try use array methods like .map

src/averageAges.js
```diff
@@ -15,11 +15,25 @@
...
+    person.age = person.died - person.born;
+    person.century = Math.ceil(person.died / 100);
+  }
+
+  const countOfMen = people.reduce((count, person) => {
```
darokrk: Replace the if statement with ternary operator

```suggestion
  const countOfMen = people.reduce((count, person) => 
  (person.sex === 'm' && person.century === century) || 
  (person.sex === 'm' && !century) ? count + 1 : count
, 0);
```

src/averageAges.js
```diff
@@ -37,7 +51,26 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => person.sex === 'f');
+
+  const countOfWomen = women.filter(isMother => {
```
darokrk: Replace the if statement with ternary operator

 ```suggestion
 const countOfWomen = women.filter(isMother =>
  withChildren
    ? people.some(person => person.mother === isMother.name)
    : people.some(person => isMother.name === person.name)
)
.reduce(count => count + 1, 0);
```

File: src/averageAges.js [hunk 4]
```diff
@@ -37,7 +51,26 @@ function calculateMenAverageAge(people, century) {
...
+    if (withChildren) {
+      return people.some(person => person.mother === isMother.name);
+    } else {
+      return people.some(person => isMother.name === person.name);
+    }
+  })
+    .reduce(count => (count + 1), 0);
+
+  return women.filter(isMother => {
+    if (withChildren) {
```
darokrk: try use ternary operator here also instead of if:

`withChildren
   ? people.some(person => person.mother === isMother.name)
   : people.some(person => isMother.name === person.name)
)`

darokrk (changes requested):
Looks good 👍🏽  some code improvements needed and we ready to move next :)

src/averageAges.js
```diff
@@ -55,10 +80,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+    if (mother) {
+      const ageDiff = (person.born - mother.born);
+
+      return ageDiff;
+    }
+  })
+    .filter(ageDiff => ageDiff !== undefined);
+
+  const countOfChildren = children.reduce((count) => (count + 1), 0);
```
darokrk: You don't need to use a separate reduce method to count the number of children. You can use the length property of the array.

src/averageAges.js
```diff
@@ -37,7 +45,24 @@ function calculateMenAverageAge(people, century) {
...
+
+  return women.filter(isMother => {
+    if (withChildren) {
+      return people.some(person => person.mother === isMother.name);
+    } else {
+      return people.some(person => isMother.name === person.name);
+    }
+  })
+    .map(isMother => (isMother.died - isMother.born))
+    .reduce((ageA, ageB) => (ageA + ageB)) / countOfWomen;
```
darokrk: Use the countAverageAge function you created earlier to calculate the average age, you have the same logic used here

src/averageAges.js
darokrk: Bump, why no refactored this code to use the ternary operator :) ? look at my earlier suggestion

src/averageAges.js
```diff
@@ -55,10 +80,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+      return ageDiff;
+    }
+  })
+    .filter(ageDiff => ageDiff !== undefined);
+
+  const countOfChildren = children.reduce((count) => (count + 1), 0);
+  const sumAgeDiff = children.reduce((sum, ageDiff) => (sum + ageDiff), 0);
+
+  return sumAgeDiff / countOfChildren;
```
darokrk: think here is also the same logic used in countAverageAge function, try to use it here to avoid code duplication

darokrk (changes requested):
Looks better, but it still need some fixes 🛠️

PR #2145, opened 2023-06-20
add task solution

NazarBaraban:


src/averageAges.js
```diff
@@ -55,10 +64,24 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  });
+
+  const sumOfAges = children.reduce((sum, child) => {
+    const mother = mothers.find((person) => person.name === child.mother);
+    const diff = mother ? child.born - mother.born : 0;
+
+    return sum + diff;
+  }, 0) / children.length;
+
+  return sumOfAges;
```
NataliGru, line 84: The result of these actions is the average value. 
Additionally, you are concatenating differences, not just ages. 
Please rename 'sumOfAges'✨

NataliGru (approved):
Well done!✨

src/averageAges.js
```diff
@@ -15,11 +16,13 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = century
```
juljaVysotska: This is where the code can be optimized.
In fact, with the help of the ternary, we have described two different filtering conditions, and because of this we are filtering the data twice in different ways. We can use the filter() method once, we just put the filtering condition inside.

juljaVysotska (changes requested):
Nice work! I found some issues, please check and fix them!

src/averageAges.js
```diff
@@ -15,11 +16,13 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person =>
+    (!century || Math.ceil(person.died / 100) === century) && person.sex === 'm'
+  );
+
+  const totalAge = men.reduce((sum, person) => sum + (person.died - person.born), 0);
```
YegorKorenkov, line 23: You can move this condition in separate function, because u use the same in calculateWomenAverageAge
image

YegorKorenkov (approved):
GJ! Please pay attention on comment

PR #2146, opened 2023-06-20
feat: averageAge provided

OKhalanskyi:


src/averageAges.js
```diff
@@ -15,11 +15,21 @@
...
-  // without nesting
+  let menArray = people.filter((person) => {
+    return person.sex === 'm';
+  });
+
+  if (century !== undefined) {
+    menArray = menArray.filter((man) => {
+      return century === Math.ceil(man.died / 100);
+    });
+  }
```
outofs: Good work), but you shouldn't use if statement(also in line 54, 88 and 99) in this task (read line 21). So please, replace it with &&, || or maybe ternary operator

outofs (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
+    })
+    .filter((man) => {
+      return !century || Math.ceil(man.died / 100) === century;
+    });
+
+  const totalAge = menArray.reduce((previous, current) => {
+    return previous + current.died - current.born;
+  }, 0);
+
+  return totalAge / menArray.length;
```
YegorKorenkov: Please move to separate functionthat takes people array and returns average age. You have the same logic in calculateWomenAverageAge function
image

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menArray = people
+    .filter((person) => {
+      return person.sex === 'm';
+    })
+    .filter((man) => {
+      return !century || Math.ceil(man.died / 100) === century;
+    });
```
YegorKorenkov: Please try to combine this two filters(use one instead of two)

src/averageAges.js
```diff
@@ -37,7 +45,23 @@ function calculateMenAverageAge(people, century) {
...
+    .filter((person) => {
+      return person.sex === 'f';
+    })
+    .filter((woman) => {
+      const isMother = people.some((person) => {
+        return person.mother === woman.name;
+      });
+
+      return !withChildren || isMother;
+    });
```
YegorKorenkov: the same, try to combine filters

YegorKorenkov (changes requested):

anstsot (approved):
approving:sparkles:

PR #2148, opened 2023-06-20
Solution

dipinchuk:


FlameInBrains (changes requested):
Good work, but some changes are necessary:
1. it is better to name your filtered arrays not as 'arr' but as filteredMen/filteredWomen etc.
2. Also you can use 'map' method for your filtered arrays in every function so they will return arrays of ages only. After this, you will be able to rewrite your 'calculate' function to perform actions only with numbers

src/averageAges.js
```diff
@@ -1,64 +1,58 @@
...
  * @param {boolean} withChildren - optional
  *
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filteredWomen = people.filter((person) => (withChildren
+    ? people.find((someone) => someone.mother === person.name)
+    && person.sex === 'f'
```
tetiana-ivanchenko, line 27: do we need to check the person sex if she somebody's mother?

tetiana-ivanchenko (approved):

PR #2154, opened 2023-06-20
Add task solution

vladmakohon:


kochetov-yevhen-may29 (approved):
Good Solutions!
looks concise

src/averageAges.js
```diff
@@ -20,6 +20,15 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
```
YegorKorenkov: Please remove comments from each function

src/averageAges.js
```diff
@@ -20,6 +20,15 @@ function calculateMenAverageAge(people, century) {
...
+  const menAges = people.filter((person) => {
+    const deathCentury = Math.ceil(person.died / 100);
+
+    return century !== undefined
+      ? deathCentury === century && person.sex === 'm'
+      : person.sex === 'm';
+  }).map(person => person.died - person.born);
+
+  return menAges.reduce((sum, age) => sum + age, 0) / menAges.length;
```
YegorKorenkov: Instead of iterate array twice, u can move logic from map method to reduce. Also please move this condition to separate function, because in calculateWomenAverageAge u use the same
```suggestion
  }).map(person => person.died - person.born);

  return menAges.reduce((sum, person) => sum + (person.died - person.born), 0) / menAges.length;
```
image

YegorKorenkov (changes requested):

anstsot (approved):
Good work, approving:sparkles:

PR #2156, opened 2023-06-20
Finished task

IgorOksentyuk:


vlopatko (approved):
Good job, this code works 👍 
For next time will be good practice if you create a new function for same calculations (25, 29 and 57, 61 line numbers, DRY rule) , but this is just advice

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const allMen = people.filter(person => person.sex === 'm');
+
+  const menInCentury = allMen.filter(
+    (man) => Math.ceil(man.died / 100) === century
+  );
```
tetiana-ivanchenko: combine this two filters depending on conditions

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
+
+  return century
+    ? menInCentury.reduce(
+      (acc, person) => acc + (person.died - person.born), 0
+    ) / menInCentury.length
```
tetiana-ivanchenko: create the separate function with the average age calculation

src/averageAges.js
```diff
@@ -37,7 +46,21 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(person => person.sex === 'f');
+  const mothersNames = [...new Set(people
+    .map(person => person.mother)
+    .filter((mother) => mother !== null))];
+
+  const mothers = women.filter(woman => mothersNames.includes(woman.name));
```
tetiana-ivanchenko: the same, this logic can be fit in one filter

tetiana-ivanchenko (changes requested):

juljaVysotska (approved):
Good job!

PR #2172, opened 2023-06-21
solution

Valentyna08:


src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => (!century
+    || Math.ceil(person.died / 100) === century)
+    && person.sex === 'm');
+  const averageAge = men.reduce((acc, person) =>
+    acc + (person.died - person.born), 0);
+
+  return averageAge / men.length;
```
anstsot: you can create separate function to calculate the average age based on array param and reuse it here and in `calculateWomenAverageAge`

anstsot (changes requested):
Looks good, let's add just small improvement

tetiana-ivanchenko (approved):

PR #2176, opened 2023-06-21
Task done

AntonBaranskyi:


src/averageAges.js
```diff
@@ -1,4 +1,5 @@
-'use strict';
+/*eslint-disable */
```
mgrinko: please fix code style instead of disabling linter

mgrinko (changes requested):

PR #2178, opened 2023-06-22
averageTask

zipic:


src/averageAges.js
```diff
@@ -15,31 +15,53 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let peopleCentury = [...people];
```
polosanya: There is no need to make a copy. You don't use any mutating methods

src/averageAges.js
```diff
@@ -15,31 +15,53 @@
...
+
+  if (century !== null && typeof century === 'number') {
```
polosanya: This check can be simplified
```suggestion
  if (!century) {
```

src/averageAges.js
```diff
@@ -15,31 +15,53 @@
...
+    peopleCentury = people.filter(person =>
+      Math.ceil(person.died / 100) === century);
+  }
+
+  const men = peopleCentury.filter(person => person.sex === 'm');
```
polosanya: In both cases, you have almost the same logic.
You can use a ternary operator inside the filter callback to make it a bit cleaner

src/averageAges.js
```diff
@@ -15,31 +15,53 @@
...
-  // write code here
-}
+function calculateWomenAverageAge(people, withwomen) {
+  let women = [...people];
+
+  if (withwomen === true) {
+    women = people.filter(person =>
+      person.sex === 'f' && people.some(p => p.mother === person.name));
+  } else {
+    women = women.filter(person => person.sex === 'f');
```
polosanya: Same here

src/averageAges.js
```diff
@@ -15,31 +15,53 @@
...
+  }
+
+  const sumAge = women.reduce((total, person) =>
+    total + (person.died - person.born), 0);
 
+  const averageAge = sumAge / women.length;
```
polosanya: It would be great to create a function-helper for calculating the average age as this logic duplicates both for men and women

polosanya (changes requested):
Good job! Let's improve your code a bit)

src/averageAges.js
```diff
@@ -15,31 +15,32 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const man = people.filter(person => person.sex === 'm'
+  && (century ? Math.ceil(person.died / 100) === century : true));
```
loralevitska, line 19:
```suggestion
  const man = people.filter(person => person.sex === 'm'
    && (century ? Math.ceil(person.died / 100) === century : true));
```

loralevitska (approved):
Well done)

PR #2179, opened 2023-06-23
add task solution

ArtemAndrushchenko:


src/averageAges.js
```diff
@@ -55,10 +72,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      .filter(([child, mother]) => mother)
+      .map(([child, mother]) => child.born - mother.born)
+  );
+};
+
+function getMother(people, child) {
+  return people.find(mother => mother.name === child.mother);
+}
+
+function getAverage(number) {
```
anstsot, line 88: it will be better to rename the param

src/averageAges.js
```diff
@@ -15,13 +15,24 @@
...
 
+function getCentury(person) {
+  return Math.ceil(person.died / 100);
+}
+
+function getChildren(people, person) {
+  return people
+    .filter(child =>
+      child.mother === person.name
+   || child.father === person.name);
```
anstsot, line 34:
```suggestion
  return people.filter(child => child.mother === person.name
    || child.father === person.name);
```

anstsot (approved):
approving:sparkles:

PR #2180, opened 2023-06-23
add task solution

EvgeniyChumak:


src/averageAges.js
```diff
@@ -55,10 +64,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const hasMother = (child) => {
+    return people.some(mother => mother.name === child.mother);
+  };
+
+  const children = (onlyWithSon)
+    ? people.filter(child => hasMother(child) && child.sex === 'm')
+    : people.filter(child => hasMother(child)
+    );
```
anstsot, line 74: can be simplified
```suggestion
  const children = people.filter(child => hasMother(child) &&
    onlyWithSon ? child.sex === 'm' : true);
```

anstsot (approved):
approving:sparkles:

PR #2182, opened 2023-06-24
Solution

Oleg-Lukanov:


src/averageAges.js
```diff
@@ -38,6 +53,21 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const women = people.filter(person => person.sex === 'f');
+  let womenChild = women;
```
alinatkachukkkk: Similar to the previous issue, there is no need to declare 'womenChild' as 'women' and then redefine it later based on a condition. You can declare it once inside the 'if' condition.

src/averageAges.js
```diff
@@ -38,6 +53,21 @@ function calculateMenAverageAge(people, century) {
...
+
+  if (withChildren) {
+    womenChild = women.filter(person => {
+      return people.find(wom => person.name === wom.mother);
```
alinatkachukkkk, line 55: The 'find' method returns the first matching element in the array. This means if there are multiple children from the same mother, they will not be included. Use 'filter' method to get all the matches.

src/averageAges.js
```diff
@@ -20,6 +20,21 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+
+  const men = people.filter(person => person.sex === 'm');
+  let menCentury = men;
```
alinatkachukkkk: There is no need to declare 'menCentury' as 'men' and then redefine it later based on a condition. You can declare it once inside the 'if' condition.

src/averageAges.js
```diff
@@ -59,6 +89,28 @@ function calculateAverageAgeDiff(people, onlyWithSon) {
   // 2. keep people who have mothers in the array
   // 3. calculate the difference child.born - mother.born
   // 4. return the average value
+  const women = people.filter(person => person.sex === 'f');
+  const womenChild = women
```
alinatkachukkkk: It is not necessary to filter 'women' before checking if a person's name matches a mother's name. You can make this check on the entire 'people' array.

src/averageAges.js
```diff
@@ -38,6 +53,21 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const women = people.filter(person => person.sex === 'f');
+  let womenChild = women;
+
+  if (withChildren) {
```
alinatkachukkkk: Instead of checking if 'withChildren' is true, you can simply put 'if (withChildren) {...}'.

src/averageAges.js
Oleg-Lukanov: I don't know how to apply the filter here. Here I am looking for women with at least one child, if you replace "find" with "filter" then there will be duplicate mothers (18 mothers instead of 13 unique ones)

PR #2183, opened 2023-06-24
Solution

BeeGooDee:


src/averageAges.js
```diff
@@ -55,7 +72,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    for (let c = 0; c < people.length; c++) {
+      if (children[i].mother === people[c].name) {
+        ageDiffs.push(Math.abs(children[i].born - people[c].born));
+      }
+    }
+  }
+
+  const ageDiffSum = ageDiffs.reduce((a, b) => a + b);
+
+  return ageDiffSum / ageDiffs.length;
```
TarasHoliuk: you are doing the same three times - move it to a separate function to reuse

src/averageAges.js
```diff
@@ -55,7 +72,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const ageDiffs = [];
+
+  let children = people.filter((a) => typeof a.mother === 'string');
+
+  if (onlyWithSon !== undefined) {
+    children = children.filter((b) => b.sex === 'm');
+  }
+
+  for (let i = 0; i < children.length; i++) {
+    for (let c = 0; c < people.length; c++) {
```
TarasHoliuk: >avoid using loop and forEach
use array methods like .filter .map .some .every .find .reduce

src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let men = people.filter((x) => (x.sex === 'm'));
+
+  if (century !== undefined) {
```
TarasHoliuk: > replace `if ()` statement with &&, || or ?:

use ternary and check it inside the first `filter` callback to define your condition inside
fix everywhere

TarasHoliuk (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let men = people.filter((p) => (p.sex === 'm'));
+
+  if (century) {
+    men = men.filter((man) => (Math.ceil(man.died / 100) === century));
+  }
```
loralevitska: All `if` statements should replaced with ternary. Fix it everywhere

src/averageAges.js
```diff
@@ -37,7 +42,19 @@ function calculateMenAverageAge(people, century) {
...
+    const mothers = people.map((p) => p.mother)
+      .filter((mother) => mother !== null);
+
+    women = women.filter((woman) => mothers.includes(woman.name));
+  }
+
+  const ageSum = women.map((woman) => woman.died - woman.born)
+    .reduce((a, b) => a + b);
+
+  return ageSum / women.length;
```
loralevitska: Repeatable code move to separate function and reuse it. Fix it everywhere.

loralevitska (changes requested):
Hi, I see you haven't fixed previous comments, maybe you did not push your changes? If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

src/averageAges.js
```diff
@@ -14,12 +14,24 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let men;
+
+  century
+    ? men = people
+      .filter((p) => (p.sex === 'm' && Math.ceil(p.died / 100) === century))
+    : men = people
+      .filter((p) => (p.sex === 'm'));
```
```suggestion
  const men = people.filter((p) => (p.sex === 'm' && century ? Math.ceil(p.died / 100) === century) : true)
```
the way you did it might be acceptable (except 1 thing*), but it would be more laconic to write it in this way

* not necessary to create `men` variable using let, you can do it in this way (correct usage of ternary operator):

```suggestion
    const men = century
      ? people.filter((p) => (p.sex === 'm' && Math.ceil(p.died / 100) === century))
      : people.filter((p) => (p.sex === 'm'));
```

TarasHoliuk (changes requested):
GJ :+1: 
I've added 1 comment about the correct usage of the ternary operator - check it and fix it with one of the suggested ways

src/averageAges.js
BeeGooDee: Thanks! Finally I get it about ternary :)

etojeDenys (approved):
good job

PR #2187, opened 2023-06-27
new solution

AndriiMysheniuk:


src/averageAges.js
```diff
@@ -37,7 +43,18 @@ function calculateMenAverageAge(people, century) {
...
+  const women = withChildren
+    ? people.filter(
+      person => person.sex === 'f'
+      && people.some(p => p.mother === person.name)
+    )
+    : people.filter(person => person.sex === 'f');
+
+  const ages = women.map(person => person.died - person.born);
+  const totalAge = ages.reduce((prev, age) => prev + age, 0);
+  const averageAge = totalAge / women.length;
```
DarkMistyRoom: this logic is the same as in the previous function. Make a separate function for it, to avoid duplication

DarkMistyRoom (changes requested):
Good job! Just one suggestion

src/averageAges.js
```diff
@@ -55,10 +72,22 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const motherChildPairs = onlyWithSon
+    ? people.filter(person => person.sex === 'm'
+    && people.find(p => p.name === person.mother))
```
SerhiiKirik, line 77: DRY
you can move it to the separate function & reuse it

src/averageAges.js
```diff
@@ -37,7 +39,22 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = withChildren
+    ? people.filter(
+      person => person.sex === 'f'
```
SerhiiKirik, line 44: DRY
you can move it to the separate function

SerhiiKirik (approved):
Great 🔥 
Let's check comments 🧑‍💻

PR #2189, opened 2023-06-28
task was done

anton-chornobai:


src/averageAges.js
```diff
@@ -55,7 +63,23 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // write code here
+  const hasMother = people.filter(child => {
+    const mother = people.find(person => person.name === child.mother);
```
Viktor-Kost, line 67: You are using a 'find' method inside a 'filter' method which leads to unnecessary computations and may slow down the execution of the program for large datasets. It is better to create a mapping of mothers to children before the 'filter' method to avoid this.

src/averageAges.js
```diff
@@ -55,7 +63,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+    return child.born - mother.born;
+  });
+
+  const sum = ageDiff.reduce((a, b) => a + b, 0);
+
+  const average = sum / ageDiff.length;
+
+  return average;
```
Viktor-Kost, line 82:
```suggestion
  return sum / ageDiff.length;
```

src/averageAges.js
```diff
@@ -9,17 +9,18 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filter = people.filter((person) => (century
+    ? Math.ceil(person.died / 100) === century
+    && person.sex === 'm' : person.sex === 'm'));
+  const mapped = filter.map((person) => person.died - person.born);
```
Viktor-Kost, line 21: 1) you can just use reduce here as a method chaining
2) it's better to use more meaningful variable names

Example:
```
const commonMenAge = selectedMen.map(...).reduce()
```

Viktor-Kost (changes requested):

PR #2190, opened 2023-06-30
SolutionAverageAges

anastasiia-burkovska:


src/averageAges.js
```diff
@@ -15,11 +15,21 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filterMan = people.filter(person => person.sex === 'm');
+
+  const filterManCentury = filterMan
+    .filter(person =>
+      Math.ceil(person.died / 100) === century);
```
VitaliyBondarenko1982:
```suggestion
  const filteredMen = people.filter(({ sex, died }) => century 
    ? sex === 'm' && Math.ceil(died / 100) === century
    : sex === 'm',
  );
```
consider to decrease amount of code and make code more readable.
check century inside filter callback and in reduce method will be no need ternary.
use destructuring.

src/averageAges.js
```diff
@@ -37,7 +47,21 @@ function calculateMenAverageAge(people, century) {
...
+    .filter(person =>
+      people.find(otherPerson => otherPerson.mother === person.name));
+
+  const result = withChildren
+    ? filterMothers
+      .reduce((sum, person) =>
+        sum + person.died - person.born, 0) / filterMothers.length
+
+    : filterWomen.reduce((sum, person) =>
+      sum + person.died - person.born, 0) / filterWomen.length;
```
VitaliyBondarenko1982: consider to make same refactoring as for calculateMenAverageAge

src/averageAges.js
```diff
@@ -55,10 +79,29 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      people.find(otherPerson => otherPerson.mother === person.name));
+
+  const filterChildren = people
+    .filter(person =>
+      people.find(otherPerson => otherPerson.name === person.mother));
+
+  const filterChildrenSons = people
+    .filter(person =>
+      people.find(otherPerson => otherPerson.name === person.mother)
+      && person.sex === 'm');
```
VitaliyBondarenko1982: try to unite three filter iterations in one. use ternary inside filter callback. after that will be no need make ternary for reduce method.
better to use some method instead find as we need just to know exist or not son in mother.

src/averageAges.js
```diff
@@ -37,7 +47,21 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filterWomen = people.filter(person => person.sex === 'f');
+
+  const filterMothers = people
+    .filter(person =>
+      people.find(otherPerson => otherPerson.mother === person.name));
```
VitaliyBondarenko1982: better to use in this case some methos instead find.

VitaliyBondarenko1982 (changes requested):
Good job.
Needs some refactoring.

Viktor-Kost (approved):

PR #2193, opened 2023-07-03
Develop

Lazarend:


src/averageAges.js
```diff
@@ -14,12 +14,30 @@
...
+
+  if (century !== undefined) {
+    const filteredMenCentury = filteredMen.filter(person =>
+      Math.ceil(person.died / 100) === century);
+
+    if (filteredMenCentury.length > 0) {
+      const totalAgeCentury = filteredMenCentury.reduce((sum, person) =>
+        sum + (person.died - person.born), 0);
+
+      return totalAgeCentury / filteredMenCentury.length;
```
SerhiiKirik: DRY
move it to the separate fucntion
```suggestion
```

src/averageAges.js
```diff
@@ -14,12 +14,30 @@
...
+    }
+  } else {
```
SerhiiKirik: try to avoid `if / else`

src/averageAges.js
```diff
@@ -36,8 +54,33 @@ function calculateMenAverageAge(people, century) {
...
+    });
+
+    if (womenWithChildren.length > 0) {
+      const totalAgeWithChildren = womenWithChildren.reduce((sum, woman) => {
+        return sum + (woman.died - woman.born);
+      }, 0);
+
+      return totalAgeWithChildren / womenWithChildren.length;
+    }
+  } else {
```
SerhiiKirik: the same here

src/averageAges.js
```diff
@@ -54,11 +97,43 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // 4. return the average value
+  const filteredMothers = people.filter(person =>
+    people.some(p => p.mother === person.name)
+  );
+
+  const filteredChildren = people.filter(person =>
+    people.some(p => p.name === person.mother)
+  );
+
+  const filteredSons = people.filter(person =>
```
SerhiiKirik: if you don't have a `onlyWithSon` you do this extra action
Try to avoid it

SerhiiKirik (changes requested):
Great 🔥 
Let's improve your solution 🧑‍💻

src/averageAges.js
Lazarend: Thanks. Yes, now it looks better.

src/averageAges.js
```diff
@@ -36,8 +55,33 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const filteredWomen = people.filter(person => person.sex === 'f');
+
+  if (withChildren) {
```
TarasHoliuk: don't use `if` use ternary instead
image

hint: you may use ternary inside filter's callback

TarasHoliuk: that will allow you to get rid of the next `if` as well (you will have only 1 correct array with people)

TarasHoliuk (changes requested):
Pay attention to the previous comments - feel free to ask in the chat if you need help or have any questions

src/averageAges.js
Lazarend: Oh, yes, I`m understand. Change if / else on ternary

etojeDenys (approved):
good job

PR #2195, opened 2023-07-04
Develop

F-u-n-H-o-u-s-e:


src/averageAges.js
```diff
@@ -14,12 +14,17 @@
...
-function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+function calculateMenAverageAge(people, century = 0) {
+  const age = century !== 0
+    ? people
+      .filter((person) => Math.ceil(person.died / 100) === century)
```
TarasHoliuk: instead of iterating all people twice check if `century` is defined inside the filter's callback and filter by century too if yes

src/averageAges.js
```diff
@@ -14,12 +14,17 @@
  *
  * @return {number}
  */
-function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+function calculateMenAverageAge(people, century = 0) {
```
```suggestion
function calculateMenAverageAge(people, century) {
```
no need to define it as a default value
undefined is a falsy value, so you may use it in ternary like this:
  `century ? //some code : //some code`
and it will work correctlyy

src/averageAges.js
```diff
@@ -14,12 +14,17 @@
...
+  const age = century !== 0
+    ? people
+      .filter((person) => Math.ceil(person.died / 100) === century)
+      .filter((person) => person.sex === 'm')
+      .map((person) => +`${person.died - person.born}`)
+    : people
+      .filter((person) => person.sex === 'm')
+      .map((person) => +`${person.died - person.born}`);
```
TarasHoliuk: the same about mapping
you shouldn't write it twice

also, why do you transform the value you get into a string if you want to manipulate it as a number (inside reduce) next?

src/averageAges.js
```diff
@@ -14,12 +14,17 @@
...
+
+  return +(age.reduce((all, now) => all + now) / age.length).toFixed(2);
```
TarasHoliuk: you are doing the same 3 times - move it to a separate function

src/averageAges.js
```diff
@@ -36,8 +41,16 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
-function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+function calculateWomenAverageAge(people, withChildren = false) {
+  const women = people.filter((person) => person.sex === 'f');
+  const withChidrenOrNo = withChildren
```
TarasHoliuk: the same as in the function above

src/averageAges.js
```diff
@@ -54,11 +67,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const children = people.filter((child) => people.find(mother => onlyWithSon
+    ? mother.name === child.mother && child.sex === 'm'
+    : mother.name === child.mother
+  ));
+
+  // 2. calculate the difference child.born - mother.born
+  const ageDifference = children.map(function(person) {
+    return person.born
+    - people[people.findIndex((human) => human.name === person.mother)].born;
```
```suggestion
    - people.find((human) => human.name === person.mother).born;
```
shorter way
find theoretically can return `undefined` but in this case `findIndex` will return `-1` and `people[-1] === undefined` as well

src/averageAges.js
TarasHoliuk: do it in the same way as you do inside find callback in third function

TarasHoliuk (changes requested):

src/averageAges.js
```diff
@@ -54,11 +62,30 @@ function calculateWomenAverageAge(people, withChildren) {
...
  * @return {number}
  */
-function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+
+function calculateAverageAgeDiff(people, onlyWithSon = false) {
+  // 1. keep people who have mothers in the array
```
OleziO: Remove all comments from the code

OleziO (changes requested):

src/averageAges.js
```diff
@@ -1,64 +1,45 @@
...
- *
- * @param {object[]} people
- * @param {boolean} withChildren - optional
- *
- * @return {number}
- */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter((person) => person.sex === 'f');
+  const withChidrenOrNo = women.filter((person) => withChildren
```
loralevitska: [Tips on naming boolean variables](https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig)
```suggestion
  const isWithChidren = women.filter((person) => withChildren
```

src/averageAges.js
```diff
@@ -1,64 +1,45 @@
...
- * @param {boolean} onlyWithSon - optional
- *
- * @return {number}
- */
-function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+function finish(array) {
```
loralevitska:
```suggestion
function getAvarageAge(array) {
```

loralevitska (approved):
Well done)

PR #2196, opened 2023-07-04
Solution

AnnaViolin23:


src/averageAges.js
```diff
@@ -55,12 +70,34 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+    return sum;
+  }, 0);
+
+  const averageAgeDiff = sumOfAgeDiffs / peopleWeNeed.length;
+
+  return averageAgeDiff;
 }
 
+// 1. find a mother of each person (or only for men)
```
volodymyr-soltys97: The comments appear to be instructions for the function above them. They should be moved above the function `calculateAverageAgeDiff` they're describing.

volodymyr-soltys97 (approved):
Good job 👍

PR #2198, opened 2023-07-05
Solution

Luk2asz:


src/averageAges.js
```diff
@@ -20,6 +20,25 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const menOfPeople = people.filter(x => x.sex === 'm');
+
+  const callback = (total, next) => {
```
darokrk: The 'callback' function is not descriptive enough. Consider renaming it to 'calcAge' or something more meaningful.

src/averageAges.js
```diff
@@ -38,6 +57,28 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+
+  const womenOfPeople = people.filter(x => x.sex === 'f');
+
+  const callback = (total, next) => {
```
darokrk: This 'callback' function looks the same as declared below and does the same calculation. Consider defining it globally to avoid repetition.

src/averageAges.js
```diff
@@ -20,6 +20,25 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const menOfPeople = people.filter(x => x.sex === 'm');
+
+  const callback = (total, next) => {
+    return total + (next.died - next.born);
+  };
+
+  if (century !== undefined) {
```
darokrk: As described in the task description you should try to avoid using `if` statement and rather use a ternary operator.

`const menOfPeopleCentury = century !== undefined ? menOfPeople.filter(x => Math.ceil(x.died / 100) === century) : [];`
`const ageSumCentury = menOfPeopleCentury.reduce(callback, 0);`
`const averageAgeCentury = menOfPeopleCentury.length > 0 ? ageSumCentury / menOfPeopleCentury.length : 0;`

`return averageAgeCentury;`

try using something like this also for the `calculateWomenAverageAge` and `calculateAverageAgeDiff` try refactor them to not using `if` statement

darokrk (changes requested):
Nice PR 👍🏽 need a few code improvements and we got this 😃 !

darokrk (changes requested):
What about the previous comments? I see that nothing has been changed in the code since the last PR, please take another look at the comments, add any corrections, and renew the PR 😃

Luk2asz:
> What about the previous comments? I see that nothing has been changed in the code since the last PR, please take another look at the comments, add any corrections, and renew the PR 😃

Please check commit cb15d717cb7a0ba88cd497ac0c1f3c355fece88c

darokrk (approved):
Good job 👏  approved 💪🏽 !

PR #2200, opened 2023-07-06
Solution

DaniilBarilotti:


src/averageAges.js
```diff
@@ -38,6 +50,20 @@ function calculateMenAverageAge(people, century) {
...
+  const women = people.filter(person =>
+    person.sex === 'f' && (!withChildren || hasChildren(person, people))
+  );
+
+  const womenAgesSum = women.reduce((sum, person) =>
+    sum + (person.died - person.born), 0);
+
+  const womenCount = women.length;
+
+  return womenCount > 0 ? womenAgesSum / womenCount : 0;
```
etojeDenys: we can move it to some helper function and reuse it here and in the prev func too

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):

PR #2206, opened 2023-07-08
add solution

denys-kononenko:


src/averageAges.js
```diff
@@ -59,6 +80,25 @@ function calculateAverageAgeDiff(people, onlyWithSon) {
   // 2. keep people who have mothers in the array
   // 3. calculate the difference child.born - mother.born
   // 4. return the average value
+
+  const mothers = people.filter(person => {
+    return people.some(item => item.mother === person.name);
+  });
+
+  const peopleWithMother = people.filter(person => {
+    return onlyWithSon !== undefined
```
volodymyr-soltys97, line 89: The `mothers.find()` operation is run on every iteration, which results in unnecessary computations. You can improve performance by storing the result in a variable before the `filter` operation.

src/averageAges.js
```diff
@@ -59,6 +80,25 @@ function calculateAverageAgeDiff(people, onlyWithSon) {
...
+      ? mothers.find(mother => mother.name === person.mother)
+        && person.sex === 'm'
+      : mothers.find(mother => mother.name === person.mother);
+  });
+
+  const totalAgeDiff = peopleWithMother.reduce((sum, person) => {
+    const mother = mothers.find(item => item.name === person.mother);
```
volodymyr-soltys97, line 96: Again, the `mothers.find()` operation is run on every iteration inside the `reduce` method. You can improve performance by storing the result in a variable before the `reduce` operation.

volodymyr-soltys97 (approved):
Good job 🔥

PR #2207, opened 2023-07-08
Solution

LiubomyrShyshak:


src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people
+    .filter(
+      human =>
+        century
+          ? (human.sex === 'm' && Math.ceil(human.died / 100) === century)
+          : human.sex === 'm'
+    );
```
agniya-i, line 24:
```suggestion
  const men = people.filter(human =>
        century
          ? (human.sex === 'm' && Math.ceil(human.died / 100) === century)
          : human.sex === 'm'
    );
```

Not necessary to start `.filter` from a new line, the same for `human`,  it  would be better to keep it on same line for readability

agniya-i (approved):

PR #2211, opened 2023-07-10
Average ages

dimkamg21:


src/averageAges.js
```diff
@@ -15,11 +15,15 @@
  * @return {number}
```

OleziO (changes requested):

demetrxx (approved):
Well done 👍 Go ahead!

PR #2216, opened 2023-07-13
Solution

Illia-L:


src/averageAges.js
```diff
@@ -14,12 +14,18 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const centuriedMen = people
+    .filter(person => person.sex === 'm')
+    .filter(man => !century || Math.ceil(man.died / 100) === century);
```
TarasHoliuk: instead of iterating whole array twice add this condition to the first filter's callback (use logic operator to handle 2 conditions in the correct way)

src/averageAges.js
```diff
@@ -14,12 +14,18 @@
...
+
+  return centuriedMen
+    .filter(person => person.sex === 'm')
```
TarasHoliuk: redundant filter - you've already filtered the people array in this way

src/averageAges.js
```diff
@@ -36,8 +42,23 @@ function calculateMenAverageAge(people, century) {
...
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let women = people.filter(person => person.sex === 'f');
+
+  if (withChildren) {
+    const mothersList = new Set(people.map(person => person.mother));
+
+    mothersList.delete(null);
```
TarasHoliuk: you're deleting `null` but reading a property from an object (`person.mother`) that doesn't exist return `undefined`

src/averageAges.js
```diff
@@ -36,8 +42,23 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let women = people.filter(person => person.sex === 'f');
+
+  if (withChildren) {
```
TarasHoliuk: it's forbidden to use `if` in this task
but the solution with `Set` is nice :+1: 
try to get rid of `if`

src/averageAges.js
```diff
@@ -54,7 +75,29 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const peopleWithAccountedMother
+    = peopleWithMother
+      .filter(person => accountedMothersList.includes(person.mother));
+
+  return peopleWithAccountedMother
+    .reduce(
+      (acc, person) => (
+        acc + person.born - motherBirthdays[person.mother]
+      ), 0)
+  / peopleWithAccountedMother.length;
```
TarasHoliuk: you are doing the same 3 times - move it to a separate function

src/averageAges.js
```diff
@@ -54,7 +75,29 @@ function calculateWomenAverageAge(people, withChildren) {
  *
  * @return {number}
  */
+
 function calculateAverageAgeDiff(people, onlyWithSon) {
+  const peopleWithMother = people
+    .filter(person => person.mother !== null)
+    .filter(person => !onlyWithSon || person.sex === 'm');
```
TarasHoliuk: combine into one filter

src/averageAges.js
```diff
@@ -54,7 +75,29 @@ function calculateWomenAverageAge(people, withChildren) {
  *
  * @return {number}
  */
+
 function calculateAverageAgeDiff(people, onlyWithSon) {
+  const peopleWithMother = people
+    .filter(person => person.mother !== null)
```
```suggestion
    .filter(person => person.mother)
```

TarasHoliuk (changes requested):

volodymyr-soltys97 (approved):
Good job 🔥

PR #2217, opened 2023-07-14
Add tusk solution.

MaksymBevskyi:


src/averageAges.js
```diff
@@ -36,8 +43,20 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
-function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+function calculateWomenAverageAge(people, withChildren = null) {
+  let women = people.filter(person => person.sex === 'f');
+
+  if (withChildren !== null) {
+    women = people
```
volodymyr-soltys97: You should filter the 'women' array instead of the 'people' array when checking for women with children.

src/averageAges.js
```diff
@@ -54,11 +73,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // 4. return the average value
+  const filteredPeople = onlyWithSon
+    ? people.filter(person => person.sex === 'm' && person.mother !== null)
+    : people.filter(person => person.mother !== null);
+
+  const ageDifference = filteredPeople.map(person => {
+    const mother = people.find(p => p.name === person.mother);
+
+    if (mother && mother.born) {
+      return person.born - mother.born;
```
volodymyr-soltys97, line 86: There's no need to check if 'mother.born' is truthy. If the 'mother' exists, 'born' property will always be a number.

src/averageAges.js
volodymyr-soltys97: The second argument 'withChildren' is expected to be a boolean. Using 'null' as default value may lead to incorrect results when it is not provided. Consider using 'false' as default value.

src/averageAges.js
```diff
@@ -14,12 +14,19 @@
  *
  * @return {number}
  */
-function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+function calculateMenAverageAge(people, century = null) {
```
volodymyr-soltys97: The second argument 'century' is expected to be a number. Using 'null' as default value may lead to incorrect results when it is not provided. Consider using an appropriate number or 'undefined' as default value.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

volodymyr-soltys97 (approved):
🚀

PR #2218, opened 2023-07-14
average ages

Orchibald:


src/averageAges.js
```diff
@@ -55,10 +69,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ageDiffs = filteredPeople.map(person => {
+    const mother = people.find(p => p.name === person.mother);
+
+    if (mother && mother.born) {
+      return person.born - mother.born;
+    }
+
+    return null;
+  });
```
VitaliyBondarenko1982:
```
redundant iteration. you can do it in reduce

File: src/averageAges.js [hunk 2]
```diff
@@ -55,10 +69,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const validAgeDiffs = ageDiffs.filter(ageDiff => ageDiff !== null);
```
VitaliyBondarenko1982: looks like redundant. you can do it in reduce
in reduce you can check whether was found personMother and return sum + person.born - personMother.bore (if was found mother) or return sum;

src/averageAges.js
```diff
@@ -55,10 +69,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const filteredPeople = onlyWithSon
+    ? people.filter(person => person.sex === 'm'
+      && person.mother !== null)
+    : people.filter(person => person.mother !== null);
```
VitaliyBondarenko1982:
```suggestion
  const filteredPeople =  people.filter(person => onlyWithSon ? person.sex === 'm'
      && person.mother !== null : person.mother !== null) )
```
better to make ternary inside filter

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
+
+  const filteredMen = century ? men.filter(
+    person => Math.ceil(person.died / 100) === century
+  ) : men;
+
+  const averageAge = filteredMen.length ? filteredMen.reduce(
+    (sum, person) => sum + (person.died - person.born), 0
+  ) / filteredMen.length : 0;
+
+  return isNaN(averageAge) ? 0 : averageAge;
```
VitaliyBondarenko1982: why do you need to check on isNaN?

VitaliyBondarenko1982 (changes requested):
Needs

etojeDenys (approved):

PR #2225, opened 2023-07-17
solution

NataliMax:


src/averageAges.js
```diff
@@ -20,6 +20,17 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
```

src/averageAges.js
```diff
@@ -20,6 +20,17 @@ function calculateMenAverageAge(people, century) {
...
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  const man = (!century)
+    ? people
+      .filter(person => person.sex === 'm')
```
OleziO, line 12: Move repeated checks into a variables

OleziO (changes requested):

etojeDenys (approved):

PR #2233, opened 2023-07-21
Solution

IhorVerba:


src/averageAges.js
```diff
@@ -36,8 +42,18 @@ function calculateMenAverageAge(people, century) {
...
+
+  womenToFilter = withChildren ? womenToFilter.filter(person =>
+    people.some(child => child.mother === person.name)) : womenToFilter;
+
+  const womenTotalAge = womenToFilter.reduce((sum, person) =>
+    sum + (person.died - person.born), 0);
+
+  const womenAverageAge = womenTotalAge / womenToFilter.length;
+
+  return womenAverageAge;
```
loralevitska: Repeatable code move to helper function `getAvarageAge` and reuse it everywhere where you need

loralevitska (changes requested):
Almost done) Let's improve

volodymyr-soltys97 (approved):
Nice work 🔥

PR #2238, opened 2023-07-25
Develop

Roman-Bilichenko:


src/averageAges.js
```diff
@@ -37,7 +44,18 @@ function calculateMenAverageAge(people, century) {
...
+  const women = people
+    .filter(item => withChildren === undefined
+      ? item.sex === 'f'
+      : people.some(obj => item.name === obj.mother));
+
+  const ageSum = women
+    .map(item => item.died - item.born)
+    .reduce((sum, age) => sum + age, 0);
+
+  const averageAge = ageSum / women.length;
```
vlad-demchuk: create a helper function to calculate average age, you do it several times

src/averageAges.js
```diff
@@ -37,7 +44,18 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people
+    .filter(item => withChildren === undefined
+      ? item.sex === 'f'
+      : people.some(obj => item.name === obj.mother));
```
vlad-demchuk: use better variable names instead of item and obj

vlad-demchuk (changes requested):

src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people
+    .filter(person => century === undefined
+      ? person.sex === 'm'
```
anastasiiavorobiova: Consider creating a variable with a meaningful name and reusing it in your function instead of magic strings
```suggestion
      const MAN = 'm';
      
      person.sex === MAN
```

src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
+      : person.sex === 'm' && Math.ceil(person.died / 100) === century);
```
anastasiiavorobiova: * Consider avoiding magic strings and numbers in your code. It's better to use variables here
* There is a repeating functionality, consider following DRY principle and creating a helper function to check the sex of the person; It may be the isMan or isWoman function

src/averageAges.js
```diff
@@ -37,7 +44,18 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people
+    .filter(person => withChildren === undefined
+      ? person.sex === 'f'
```
anastasiiavorobiova: See my previous comments

src/averageAges.js
```diff
@@ -59,8 +77,37 @@ function calculateAverageAgeDiff(people, onlyWithSon) {
   // 2. keep people who have mothers in the array
   // 3. calculate the difference child.born - mother.born
   // 4. return the average value
+  const mothers = people
+    .filter(person => onlyWithSon === undefined
+      ? people.some(child => person.name === child.mother)
+      : people.some(child => person.name === child.mother
+        && child.sex === 'm'));
```
anastasiiavorobiova: The same issue here

src/averageAges.js
```diff
@@ -59,8 +77,37 @@ function calculateAverageAgeDiff(people, onlyWithSon) {
...
+
+  const ages = people
+    .map(child => {
+      const mother = mothers
+        .find(person => onlyWithSon === undefined
+          ? child.mother === person.name
+          : child.sex === 'm' && child.mother === person.name);
```
anastasiiavorobiova: Also, you can use the helper function here

anastasiiavorobiova (changes requested):
Great work! Added comments to fix

anastasiiavorobiova (approved):
Awesome work!

PR #2240, opened 2023-07-28
solution

Sancz0pansa:


src/averageAges.js
```diff
@@ -55,6 +81,27 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
+  let peopleArr = [];
+
+  if (onlyWithSon === true) {
```
choeqq:
```suggestion
  if (onlyWithSon) {
```
that also works 😄

src/averageAges.js
```diff
@@ -37,7 +50,20 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let woman = [];
+
+  if (withChildren === true) {
```
choeqq:
```suggestion
  if (withChildren) {
```
that also works 😄

src/averageAges.js
```diff
@@ -15,6 +15,19 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  let man = [];
+
+  if (century === undefined) {
```
choeqq:
```suggestion
  if (!century) {
```
that also works 😄

choeqq (changes requested):
Almost there, you can make it litlte simpler 🚀

src/averageAges.js
```diff
@@ -15,6 +15,19 @@
...
+
+  if (!century) {
+    man = people.filter(person => person.sex === 'm');
+  } else {
+    man = people.filter(person =>
+      person.sex === 'm' && Math.ceil(person.died / 100) === century);
+  }
+
+  const sum = man.reduce((prev, person) =>
+    prev + (person.died - person.born), 0);
```
JoyCoffeeAddict: How about creating additional function for calculating age sum, that would work for both `men` and `women`:
`calculateAgeSum(filteredGroup)`, (you can choose your own names)

src/averageAges.js
```diff
@@ -15,6 +15,19 @@
...
  */
 function calculateMenAverageAge(people, century) {
+  let man = [];
+
+  if (!century) {
+    man = people.filter(person => person.sex === 'm');
+  } else {
+    man = people.filter(person =>
+      person.sex === 'm' && Math.ceil(person.died / 100) === century);
+  }
```
JoyCoffeeAddict: I would always assign all males at the begining: 
```
if(some condition) { 
///filter men from century
} else  {
//return calculateSum(...)
}
```

Your solution is correct, but you could extract filtering into one place.

src/averageAges.js
```diff
@@ -37,7 +50,20 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  let woman = [];
+
+  if (withChildren) {
+    woman = people.filter(person =>
+      person.sex === 'f' && people.some(otherPerson =>
+        otherPerson.mother === person.name));
+  } else {
+    woman = people.filter(person => person.sex === 'f');
+  }
```
JoyCoffeeAddict: As above, i would firstly filter `women` and then filter those with children.

src/averageAges.js
```diff
@@ -55,6 +81,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const mothers = people.filter(mother =>
+    mother.sex === 'f' && peopleArr.some(person =>
+      person.mother === mother.name));
+
+  const ageDifferences = mothers.flatMap(mother => {
+    const children = peopleArr.filter(child => child.mother === mother.name);
+
+    return children.map(child => child.born - mother.born);
+  });
```
JoyCoffeeAddict: This is correct, but i think it would be far more readable to get `childrenSum`, `mothersSum`, calculate the difference between them and divide by number of people.

``` 
x, y ->  children ages, 
a, b -> mothers ages

then: (a - x + b - y)  / 2 === (a + b - (x + y)) / 2
```

JoyCoffeeAddict: Try my solution and then decide which one do you like more.

JoyCoffeeAddict (changes requested):
Very good job!

src/averageAges.js
```diff
@@ -15,6 +15,15 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
+  let man = [];
+
+  man = people.filter((person) => person.sex === 'm');
+
+  if (century) {
+    man = man.filter((person) => Math.ceil(person.died / 100) === century);
+  }
```
asakevych: there is no sense to use filter twice

your algorithm will work faster if you iterate just once: 
  imagine you have array with n elements, when you iterate throw it you do n operations, so when you iterate throw it twice you will have 2n operations - > less operations === faster code -> n will be faster then 2n 
  
you can add our condition to filter function: 
```suggestion
    man = people.filter((person) => century 
      ?  Math.ceil(person.died / 100) === century) && person.sex === 'm' 
      :  person.sex === 'm' 
    );
```

asakevych (changes requested):
Great job!! your solution is really good but you can make it even better 😉

choeqq (approved):
Good job ✅

PR #2241, opened 2023-07-28
all test pass

aniashev:


src/averageAges.js
```diff
@@ -1,64 +1,72 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = isMan(people)
+    .filter(man => !century || Math.ceil(man.died / 100) === century);
+
+  const lifeYears = men.map(man => man.died - man.born);
+  const agesSum = lifeYears.reduce((sum, n) => sum + n, 0);
+
+  return agesSum / men.length;
```
etojeDenys: we can move this logic into some helper function

etojeDenys (changes requested):

src/averageAges.js
```diff
@@ -1,64 +1,76 @@
...
+  const lifeYears = array.map(arr => arr.died - arr.born);
+  const agesSum = lifeYears.reduce((sum, n) => sum + n, 0);
+
+  return agesSum / array.length;
+}
+
+function averageAgePartial(array) {
+  const sumOfDifferences = array.reduce((sum, n) => (sum + n), 0);
+
+  return sumOfDifferences / array.length;
```
etojeDenys:
```suggestion
function averageAgeComplete(array) {
  const lifeYears = array.map(arr => arr.died - arr.born);
  
  return averageAgePartial(lifeYears);
}

function averageAgePartial(array) {
  const sumOfDifferences = array.reduce((sum, n) => (sum + n), 0);

  return sumOfDifferences / array.length;
```

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):
Looks good

PR #2244, opened 2023-07-29
add task solution

4xmplme:


src/averageAges.js
```diff
@@ -55,10 +79,46 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  ));
+
+  const peopleWithMothers = (
+    onlyWithSon
+      ? men
+      : people
+  ).map(person => ({
+    ...person,
+    motherObj: mothers.find(mother => (
+      mother.name === person.mother || null
```
volodymyr-soltys97: The || null operation seems unnecessary because if mother.name does not equal person.mother, undefined will be returned by default.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

nazarmatsevych (approved):

PR #2245, opened 2023-07-30
averageAges solution

anpiet:


choeqq (approved):
Most of the times if statement is more readable than ternany operator, but you were prompted to use ternary, so nicely done ✅

PR #2254, opened 2023-08-01
Solution

in7264:


src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let man = people.filter(person => person.sex === 'm');
+
+  century && (
```
TarasHoliuk: let's check if `century` is defined inside `filter`s callback to avoid an additional iterating
fix in other functions too

src/averageAges.js
```diff
@@ -54,11 +76,30 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  onlyWithSon && (
+    peopleWithMothers = peopleWithMothers.filter(person => person.sex === 'm')
+  );
+
+  const { totalAgeDiff, validCount } = peopleWithMothers
+    .reduce(
+      (acc, person) => ({
+        totalAgeDiff: acc.totalAgeDiff + (people
+          .find(mother => mother.name === person.mother) ? person.born - people
```
TarasHoliuk: `.find(mother => mother.name === person.mother)`

you are doing it 3 times inside the reduce callback
save it into a variable inside the callback

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
+  century && (
+    man = man
+      .filter(person =>
+        Math.ceil(person.died / 100) === century)
+  );
+
+  const sumOfAge = man
+    .reduce((sum, person) => sum + (person.died - person.born), 0);
+
+  const averageAge = sumOfAge / man.length;
```
```suggestion
  return sumOfAge / man.length;
```

TarasHoliuk (changes requested):

TarasHoliuk (approved):
:+1:

PR #2255, opened 2023-08-01
Average age solution

oshapkun:


src/averageAges.js
```diff
@@ -1,64 +1,100 @@
...
+      : isWoman(person);
+  });
+
+  const sumOfAges = women.reduce((sum, person) => {
+    return sum + age(person);
+  }, 0);
+
+  const avarageAge = sumOfAges / women.length;
+
+  return avarageAge;
```
loralevitska: Repeatable code move to helper function and reuse it

loralevitska (changes requested):
Good job! There are things that should be fixed:

anastasiiavorobiova (approved):
Awesome!

PR #2256, opened 2023-08-01
Average Ages

Kacper-Leman:
Soultion :D

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -17,7 +17,8 @@
 function calculateMenAverageAge(people, century) {
   const men = century
     ? people.filter(person =>
-      person.sex === 'm' && Math.ceil(person.died / 100) === century)
+      person.sex === 'm'
+      && Math.ceil(person.died / 100) === century)
     : people.filter(person => person.sex === 'm');
 
   if (men.length === 0) {
@@ -77,30 +78,22 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  const mothers = people.reduce((acc, person) => {
-    if (person.mother !== null && !acc.includes(person.mother)) {
-      acc.push(person.mother);
-    }
+  const mothers = people.filter(mother =>
+    people.some(child => mother.name === child.mother)
+  );
 
-    return acc;
-  }, []);
+  const children = people.filter(child =>
+    mothers.find(mother => mother.name === child.mother)
+      && (onlyWithSon ? child.sex === 'm' : true)
+  );
 
-  const mothersWithSons = people.filter(person =>
-    mothers.includes(person.name));
+  const agesDiff = children.map(child => {
+    const mother = mothers.find(mum => mum.name === child.mother);
 
-  const children = onlyWithSon
-    ? people.filter(person =>
-      person.sex === 'm' && mothers.includes(person.mother))
-    : people;
-
-  const ageDifferences = children.map(child => {
-    const motherData = mothersWithSons.find(mother =>
-      mother.name === child.mother);
-
-    return motherData ? child.born - motherData.born : NaN;
+    return child.born - mother.born;
   });
 
-  const validAgeDifferences = ageDifferences.filter(diff => !isNaN(diff));
+  const validAgeDifferences = agesDiff.filter(diff => !isNaN(diff));
   const sumAgeDifference = validAgeDifferences.reduce((acc, diff) =>
     acc + diff, 0);
 
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -15,11 +15,20 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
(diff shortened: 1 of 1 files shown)


src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = century
+    ? people.filter(person =>
+      person.sex === 'm' && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === 'm');
```
DorotaLeniecDev: For readability:
```suggestion
  const men = century
    ? people.filter(person =>
      person.sex === 'm' 
      && Math.ceil(person.died / 100) === century)
    : people.filter(person => person.sex === 'm');
```

src/averageAges.js
```diff
@@ -55,10 +77,36 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const mothers = people.reduce((acc, person) => {
+    if (person.mother !== null && !acc.includes(person.mother)) {
+      acc.push(person.mother);
+    }
+
+    return acc;
+  }, []);
+
+  const mothersWithSons = people.filter(person =>
+    mothers.includes(person.name));
```
DorotaLeniecDev: This could be simplified, to get all mothers you can write:
```js
const mothers = people.filter(mother => 
  people.some(child => mother.name === child.mother)
)
```

src/averageAges.js
```diff
@@ -55,10 +77,36 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const children = onlyWithSon
+    ? people.filter(person =>
+      person.sex === 'm' && mothers.includes(person.mother))
+    : people;
```
DorotaLeniecDev: This should return only children not all people like so:
```js
const children = people.filter(child => 
  mothers.find(mother => mother.name === child.mother)
  && (onlyWithSon ? child.sex === 'm' : true)
);
```

src/averageAges.js
```diff
@@ -55,10 +77,36 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ageDifferences = children.map(child => {
+    const motherData = mothersWithSons.find(mother =>
+      mother.name === child.mother);
+
+    return motherData ? child.born - motherData.born : NaN;
+  });
```
DorotaLeniecDev: As we are sure all chuldrens have mothers we could write
```js
const agesDiff = children.map(child => {
  const mother = mothers.find(mum => mum.name === child.mother);

  return child.born - mother.born
})
```

DorotaLeniecDev (changes requested):
Good job, small adjustments needed

src/averageAges.js
```diff
@@ -15,11 +15,21 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = century
```
EwaKobiela, line 18: Suggestion to optimize the code: instead of filtering the same array twice for the 'century' and 'sex' conditions, you can do it in one go. This will make your code cleaner and more efficient.

src/averageAges.js
```diff
@@ -55,10 +78,28 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      && (onlyWithSon ? child.sex === 'm' : true)
+  );
+
+  const agesDiff = children.map(child => {
+    const mother = mothers.find(mum => mum.name === child.mother);
+
+    return child.born - mother.born;
+  });
+
+  const validAgeDifferences = agesDiff.filter(diff => !isNaN(diff));
```
EwaKobiela, line 96: The filter operation here seems redundant because the map operation on line 66 should always provide valid numbers. You can safely remove this line.

src/averageAges.js
```diff
@@ -37,7 +47,20 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = withChildren
```
EwaKobiela, line 50: Similar to the previous comment, it would be more efficient to perform all the conditions in one filter method instead of filtering the array twice. This will increase the performance of your code.

src/averageAges.js
```diff
@@ -55,10 +78,28 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const mothers = people.filter(mother =>
```
EwaKobiela, line 81: Instead of filtering the array twice (once for mothers and once for children), you can create a map of mothers and use it to filter children. This approach would be more efficient.

EwaKobiela (approved):
Good job!

PR #2259, opened 2023-08-02
solution

cardinal1991:


choeqq (approved):
Nicely done, you don't really have to add comments to your code - it's pretty self-explanatory, comments are necessary when we're forcing some custom behavior 🚀

PR #2261, opened 2023-08-02
init commit

domalewski:


Matylda-Lewandowska (approved):
Looks fine to me, it's really well readable, and it is working properly. :)

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
+  const menInCentury = century
+    ? men.filter(person => Math.ceil(person.died / 100) === century)
+    : men;
```
asakevych: there is no sense to use filter twice

your algorithm will work faster if you iterate just once:
imagine you have array with n elements, when you iterate throw it you do n operations, so when you iterate throw it twice you will have 2n operations - > less operations === faster code -> n will be faster then 2n

you can add our condition to filter function:
```suggestion
  const menInCentury = people.filter((person) => century 
      ?  Math.ceil(person.died / 100) === century) && person.sex === 'm' 
      :  person.sex === 'm' 
    );
```

asakevych (changes requested):
Great job!! Your solution is really good, but I think it can be event better 😉

choeqq (approved):

PR #2266, opened 2023-08-02
task solution

HubertKlamann:


src/averageAges.js
```diff
@@ -55,10 +66,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  children.forEach(person => {
+    const mothers = people.find(mother => mother.name === person.mother);
+
+    if (mothers) {
+      const difference = person.born - mothers.born;
+
+      totalMotherAge += difference;
+    }
+  });
```
choeqq, line 86: please use reduce in cases like this

choeqq (approved):
Well done, please take a look at my comment ✅

PR #2269, opened 2023-08-02
solution

PawelP2001:


src/averageAges.js
```diff
@@ -1,64 +1,81 @@
...
+        const meetsCondition = !onlyWithSon || person.sex === 'm';
+
+        if (meetsCondition) {
+          const ageDifference = Math.abs(mother.born - person.born);
+
+          ageDifferences.push(ageDifference);
+        }
+      }
+    }
+  }
```
choeqq: you can also do that with reduce

choeqq (approved):
Looks good to me, please take a look at my comment 🚀

PR #2285, opened 2023-08-06
feat: add task solution

FrontKid:


src/averageAges.js
```diff
@@ -14,12 +14,15 @@
  *
  * @return {number}
  */
+
+const getAvarageAge = people => people
+  .reduce((acc, { born, died }) => acc + (died - born), 0) / people.length;
```
33Orange: The 'getAverageAge' function could return NaN if the 'people' array is empty. You should add a check to see if 'people.length' is not zero before performing the division.

src/averageAges.js
```diff
@@ -55,10 +62,14 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const children = people.filter(({ mother, sex }) =>
+    mother && ((onlyWithSon && sex === 'm') || (!onlyWithSon))
+      && people.find(mom => mother === mom.name));
+
+  const ageDiff = children.map(({ mother: childMother, born }) =>
+    born - people.find(({ name }) => childMother === name).born);
+
+  return ageDiff.reduce((total, value) => total + value, 0) / ageDiff.length;
```
33Orange: You should use the 'getAverageAge' function here instead of repeating the logic to calculate average.

src/averageAges.js
```diff
@@ -14,12 +14,15 @@
  *
  * @return {number}
  */
+
+const getAvarageAge = people => people
```
33Orange: Grammar mistake 

```suggestion
const getAverageAge = people => people
```

33Orange (changes requested):
Almost done

Vayts (approved):
gltm!

src/averageAges.js
```diff
@@ -14,12 +14,21 @@
  *
  * @return {number}
  */
+
+const isObject = (value) => typeof value === 'object' && value !== null;
+
+const getAverageAge = people => isObject(people[0])
```
alexander-ignatow, line 20: The check is here - that's good

However, you are a bit inconsistent
if you are trying to see that the `people` array is not only empty but also has some meaningful data - it makes sense also check the `born` and `died` properties. But that looks like overkill for the current task

I suggest you simply check if `people.length` is greater than 0.

src/averageAges.js
```diff
@@ -14,12 +14,21 @@
...
+
+  ? people.reduce((acc, { born, died }) =>
+    acc + (died - born), 0) / people.length || 0
+
+  : people.reduce((total, value) => total + value, 0);
```
roman-mirzoian: Why are you checking if the first element is of type "object"?
1. The object and the array will be of this type: `typeof {} === 'object'` / `typeof [] === 'object'` -  therefore, this check makes no sense.
2. In the second `reduce`, you do not return the average value, but only calculate the sum, so the name `getAverageAge` does not fit here and it can be transferred to a separate function.

alexander-ignatow: I  don't quite get this line

If the array is empty average age should be simply `0`, right?

src/averageAges.js
```diff
@@ -14,12 +14,21 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(({ sex, died }) =>
+    sex === 'm' && (century ? Math.ceil(died / 100) === century : true));
```

src/averageAges.js
```diff
@@ -55,10 +68,14 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const children = people.filter(({ mother, sex }) =>
+    mother && ((onlyWithSon && sex === 'm') || (!onlyWithSon))
+      && people.find(mom => mother === mom.name));
```
roman-mirzoian: Now everything is written correctly, but such code is difficult to read and maintain in the future. Break it into smaller parts.

roman-mirzoian (changes requested):

alexander-ignatow (changes requested):
Good! Still some issues are still there

src/averageAges.js
FrontKid: I do that becuz my function works with type object and type number
I didn't want to create the second separate function to get average with number

src/averageAges.js
FrontKid: my func works with nums and object types so I added this check to know which type is there

src/averageAges.js
FrontKid: it's not a check for empty array I answered you from above

src/averageAges.js
```diff
@@ -14,12 +14,25 @@
  *
  * @return {number}
  */
+
+const isObject = (value) => typeof value === 'object' && value !== null;
+
+const getAverageAge = people => isObject(people[0])
```
roman-mirzoian, line 20: In the second `reduce`, you do not return the average value, but only calculate the sum, so the name `getAverageAge` does not fit here and it can be transferred to a separate function.

src/averageAges.js
```diff
@@ -14,12 +14,25 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const AGE_DIVIDER = 100;
+
+  const men = people.filter(({ sex, died }) =>
+    sex === 'm' && (century
```
roman-mirzoian, line 31: It would be cool to put all the "magic values" in separate constants.

src/averageAges.js
```diff
@@ -37,7 +50,11 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter(({ name, sex }) =>
+    sex === 'f' && (withChildren ? people.some(({ mother }) =>
```
roman-mirzoian, line 54: It would be cool to put all the "magic values" in separate constants.

src/averageAges.js
```diff
@@ -55,10 +72,17 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const children = people.filter(({ mother, sex }) => {
+    const isMotherOnlyWithSon = mother && (onlyWithSon && sex === 'm');
```
roman-mirzoian, line 76: It would be cool to put all the "magic values" in separate constants.

src/averageAges.js
```diff
@@ -55,10 +72,17 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    const personMother = people.find(mom => mother === mom.name);
+
+    return (isMotherOnlyWithSon || (!onlyWithSon)) && personMother;
+  });
+
+  const ageDiff = children.map(({ mother: childMother, born }) =>
+    born - people.find(({ name }) => childMother === name).born);
```
roman-mirzoian: For better readability, the search for the mother can be transferred to a separate variable and a ready-made value can be substituted into the calculation.

roman-mirzoian (changes requested):
Pay attention to the comments, if you disagree with something, give a reasoned answer why you left everything as it is.

src/averageAges.js
FrontKid: image
The second reduce returns average age

src/averageAges.js
FrontKid: I think it's an unnecessary constant, because when we compare sex with the first letters of "male" and "female", then it is obvious that we are comparing these two things.

BudnikOleksii: If the sex property changes to be "male" and "female" instead of "m" or "w" you have to change it in all places in your code instead of just updating a variable so it makes sense to get rid of magic values.

BudnikOleksii (approved):
Good job!
Take into account why we create variables for magic values I answered in the comment.

PR #2287, opened 2023-08-07
add task solution

atrofymovych:


src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
+  const CENTURY_DIVIDER = 100;
+  const filteredPeople = people.filter(person =>
+    (century !== undefined
+    && Math.ceil(person.died / CENTURY_DIVIDER) === century
+    && person.sex === 'm')
+    || (century === undefined && person.sex === 'm')
+  );
+
+  const totalAge = filteredPeople.reduce((sum, person) =>
+    sum + (person.died - person.born), 0);
```
volodymyr-soltys97: Avoid using 'reduce' directly on 'filteredPeople' array without checking if it's empty. This could throw an exception if the array is empty. A possible fix could be to add a condition check before 'reduce'.

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const CENTURY_DIVIDER = 100;
+  const filteredPeople = people.filter(person =>
+    (century !== undefined
```
volodymyr-soltys97, line 20: The filter operation for 'filteredPeople' might result in an empty array if no conditions are met. It's a good idea to add a check if the 'filteredPeople' array is not empty before performing calculations on it.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
  * @return {number}
  */
```
anastasiiavorobiova, line 16: Consider creating here the set of constants and helper functions to use instead of magic strings and numbers and minimizing code repeating

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const CENTURY_DIVIDER = 100;
+  const filteredPeople = people.filter(person =>
+    (century !== undefined
+    && Math.ceil(person.died / CENTURY_DIVIDER) === century
+    && person.sex === 'm')
```
anastasiiavorobiova, line 22:
```suggestion
    && isMan(person))
```
Consider creating a helper function for conducting this check

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
+    || (century === undefined && person.sex === 'm')
```
anastasiiavorobiova, line 23:
```suggestion
    || (century === undefined && isMan(person))
```

src/averageAges.js
anastasiiavorobiova: For example: 
```const MAN = 'm'; ```

File: src/averageAges.js [hunk 6]
```diff
@@ -36,8 +45,15 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
-function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+
+function calculateWomenAverageAge(people, withChildren = false) {
+  const filteredWomen = people.filter(person => person.sex === 'f'
```
anastasiiavorobiova, line 50:
```suggestion
  const filteredWomen = people.filter(person => !isMan(person)
```

src/averageAges.js
```diff
@@ -54,11 +70,21 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
-function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+function calculateAverageAgeDiff(people, onlyWithSon = false) {
+  const children = people.filter(child => people.find(
+    mother => mother.name === child.mother)
+    && (onlyWithSon ? child.sex === 'm' : true));
```
anastasiiavorobiova, line 76:
```suggestion
    && (onlyWithSon ? isMan(child) : true));
```

anastasiiavorobiova (changes requested):
Great work! Let's improve it a bit

PR #2288, opened 2023-08-07
Add task solution

Wesses:


src/averageAges.js
```diff
@@ -14,12 +20,18 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const mensOfCenturyArray = people
+    .filter(person => person.sex === MALE_SEX)
+    .filter(person => century
+      ? century === Math.ceil(person.died / 100) : true);
```
DenisChakhalian: Don't use literals in expressions (HARDCODE).

you can put 100 into a variable

src/averageAges.js
```diff
@@ -14,12 +20,18 @@
...
+
+  return mensOfCenturyArray
+    .reduce((year, person) => year + person.died - person.born, 0)
+    / mensOfCenturyArray.length;
```
DenisChakhalian: It's not clear what exactly the function returns.
you can split it into a few variables to make it more clear.

DenisChakhalian: and in other functions too

DenisChakhalian (changes requested):
good job 🚀
but let's fix some things

src/averageAges.js
```diff
@@ -1,5 +1,11 @@
 'use strict';
 
+/* eslint-disable no-console */
+// write code here
+// learn how to use array methods like .filter .map .some .every .find .reduce
+// avoid using loop and forEach
+// replace `if ()` statement with &&, || or ?:
+// without nesting
```
33Orange: Remove unnecessary comments

src/averageAges.js
```diff
@@ -14,12 +20,21 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const mensOfCenturyArray = people
+    .filter(person => person.sex === MALE_SEX)
+    .filter(person => century
+      ? century === Math.ceil(person.died / CUNTURY_COEFFICIENT) : true);
```
33Orange: You can use 1 filter instead, just move checks from the second filter to the first.

```suggestion
    .filter(person => (
    person.sex === MALE_SEX && (century 
      ? Math.ceil(person.died / CUNTURY_COEFFICIENT) === century 
      : true)
    ))
```

src/averageAges.js
```diff
@@ -14,12 +20,21 @@
...
+
+  const MenAverageAge = mensOfCenturyArray
+    .reduce((year, person) => year + person.died - person.born, 0)
+  / mensOfCenturyArray.length;
```
33Orange: Create a separate helper function for calculating average age and use it here

src/averageAges.js
```diff
@@ -37,7 +52,18 @@ function calculateMenAverageAge(people, century) {
...
+  const womensOfCenturyArray = people
+    .filter(person => person.sex === FEMALE_SEX)
+    .filter(person => withChildren
+      ? people
+        .some(child => child.mother === person.name)
+      : true);
+
+  const WomenAverageAge = womensOfCenturyArray
+    .reduce((year, person) => year + person.died - person.born, 0)
+  / womensOfCenturyArray.length;
```
33Orange: Same here

src/averageAges.js
```diff
@@ -37,7 +52,18 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womensOfCenturyArray = people
+    .filter(person => person.sex === FEMALE_SEX)
+    .filter(person => withChildren
+      ? people
+        .some(child => child.mother === person.name)
+      : true);
```
33Orange: Same about 1 filter instead of 2.
This will remove the extra iteration over the array

src/averageAges.js
```diff
@@ -55,12 +81,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  / peopleWithMothers.length;
+
+  return AverageAgeDiff;
 }
 
+// 1. find a mother of each person (or only for men)
+// 2. keep people who have mothers in the array
+// 3. calculate the difference child.born - mother.born
+// 4. return the average value
+
```
33Orange: unnecessary comment

src/averageAges.js
```diff
@@ -55,12 +81,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    .filter(person => onlyWithSon
+      ? person.sex === MALE_SEX
+      : true)
+    .filter(person => people
+      .some(child => child.name === person.mother));
+
+  const AverageAgeDiff = peopleWithMothers
+    .reduce((year, child) => year + child.born - people
+      .find(mother => child.mother === mother.name).born, 0)
+  / peopleWithMothers.length;
```
33Orange: Create a separate variable for the ages array and use your new `calculateAverageAge` function

33Orange (changes requested):
Almost done

src/averageAges.js
Wesses: could you explain in more detail, I don’t understand how I can get rid of the `find()` in this case

dffUqp (approved):
LGTM🔥

PR #2289, opened 2023-08-07
solution

DenisChakhalian:


Koliras (commented):
imho, there are too many "if", but good job

src/averageAges.js
```diff
@@ -15,11 +15,32 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const ONE_CENTURY = 100;
+  const FEMALE = 'f';
```
AndrMar1939: you can move a variable outside the function and create it for all the document

src/averageAges.js
```diff
@@ -15,11 +15,32 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
```
AndrMar1939, line 21: replace `if ()` statement with &&, || or ?:

looks like a future reject. Your code contain a lot of "if"

AndrMar1939 (approved):

src/averageAges.js
```diff
@@ -14,12 +14,24 @@
...
-  // without nesting
+  const menInCentury = people
+    .filter(person => person.sex === FEMALE ? false
+      : !century ? true
+        : Math.ceil(person.died / ONE_CENTURY) === century);
```
dffUqp: Now it looks like: 
```js
if (person.sex === FEMALE) {
  return false;
} else {
  if (!century) {
    return true;
  } else {
    return Math.ceil(person.died / ONE_CENTURY) === century;
  }
}
```

I think we can make your condition a little bit shorter.


little tip:
`person.sex === "m" & (century === true ? do something : do something)`

src/averageAges.js
```diff
@@ -37,7 +49,22 @@ function calculateMenAverageAge(people, century) {
...
+  const mothers = people
+    .filter(person => person.mother !== null)
+    .map(person => person.mother);
+
+  const women = people
+    .filter(person => person.sex === FEMALE)
+    .filter(person => withChildren ? mothers.includes(person.name) : true);
+
+  const sumAgesWomen = women
+    .reduce((sumAges, person) => sumAges + (person.died - person.born), 0);
```
dffUqp: we already use it twice, let's create a function for these calculations

src/averageAges.js
```diff
@@ -37,7 +49,22 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothers = people
```
dffUqp: looks like name for this variable a little bit wrong

src/averageAges.js
```diff
@@ -14,12 +14,24 @@
...
-  // without nesting
+  const menInCentury = people
+    .filter(person => person.sex === FEMALE ? false
+      : !century ? true
+        : Math.ceil(person.died / ONE_CENTURY) === century);
+
+  const sumAgesMenInCentury = menInCentury
+    .reduce((sumAges, person) => sumAges + (person.died - person.born), 0);
+
+  const amountMenInCentury = menInCentury.length;
```
dffUqp: I think we can remove this variable

src/averageAges.js
```diff
@@ -14,12 +14,24 @@
  *
  * @return {number}
  */
+const MALE = 'm';
+const FEMALE = 'f';
+const ONE_CENTURY = 100;
```
dffUqp, line 19: 👍

dffUqp (changes requested):
Almost done:smiley:

src/averageAges.js
```diff
@@ -37,7 +51,19 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothersNames = people
+    .filter(person => person.mother !== null)
```
```suggestion
    .filter(person => person.mother)
```
If you want to filter persons where the mother exists, you can just return like this instead of comparing with null. It will handle all falsy values.

src/averageAges.js
```diff
@@ -37,7 +51,19 @@ function calculateMenAverageAge(people, century) {
...
+    .map(person => person.mother);
+
+  const women = people
+    .filter(person => person.sex === FEMALE)
+    .filter(person => withChildren ? mothersNames.includes(person.name) : true);
```
BudnikOleksii: The second time you filter only in case the withChildren parameter is passed, in another case you just iterate without any sense. So it's better to filter only if you actually need it.

src/averageAges.js
```diff
@@ -55,10 +81,31 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  let peopleWithMother = people
+    .filter(person => person.mother !== null)
+    .filter(person => onlyWithSon ? person.sex === MALE : true);
```
BudnikOleksii: The same things as I mentioned above

src/averageAges.js
```diff
@@ -55,10 +81,31 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    .filter(person => mothersNames.includes(person.name));
+
+  peopleWithMother = peopleWithMother
+    .filter(person => mothersWithChildren
+      .find(mom => mom.name === person.mother));
+
+  const sumAgesDiff = peopleWithMother
+    .reduce((sumAge, person) => {
+      const mother = mothersWithChildren
+        .find(mom => mom.name === person.mother);
```
BudnikOleksii: You do the same operations, looks like it can be simplified.

BudnikOleksii (changes requested):
Good job👍
Left some comments, check them, and improve your solution.

roman-mirzoian (approved):
Well done 👍

PR #2290, opened 2023-08-07
add task solution

ydem1:


src/averageAges.js
```diff
@@ -14,12 +14,14 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+function calculateMenAverageAge(people, century = false) {
+  const males = people.filter((male) => {
+    return century === false
+      ? male.sex === 'm'
+      : (male.sex === 'm' && Math.ceil(male.died / 100) === century);
+  });
```
xiimiik: depending on the case, we return the appropriate callback

```suggestion
  const males = people.filter(
    century
      ? male => male.sex === 'm'
      : male => male.sex === 'm'
        && Math.ceil(male.died / 100) === century;
  );
```

src/averageAges.js
```diff
@@ -36,8 +38,15 @@ function calculateMenAverageAge(people, century) {
...
-function calculateWomenAverageAge(people, withChildren) {
+function calculateWomenAverageAge(people, withChildren = false) {
   // write code here
+  const females = people.filter((female) => {
+    return withChildren === false
+      ? female.sex === 'f'
+      : (people.find((man) => man.mother === female.name));
+  });
+
+  return calculateAverage(females);
```
xiimiik: same as above

src/averageAges.js
```diff
@@ -54,11 +63,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    });
+
+    return accumulator + currentMother;
+  }, 0) / childs.length).toFixed(2);
+}
+
+function calculateAverage(people) {
+  return +(people.reduce((accumulator, currentValue) => {
+    return (accumulator + (currentValue.died - currentValue.born));
+  }, 0) / people.length).toFixed(2);
```
xiimiik:
```suggestion
function calculateAverage(people) {
  return people.reduce((accumulator, currentValue) => {
    return (accumulator + (currentValue.died - currentValue.born));
  }, 0) / people.length;
```

src/averageAges.js
```diff
@@ -54,11 +63,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // 4. return the average value
+function calculateAverageAgeDiff(people, onlyWithSon = false) {
+  const mothers = people.filter((female) => female.sex === 'f');
+  const childs = mothers.flatMap((mother) => {
+    return people.filter((person) => {
+      return onlyWithSon === false
+        ? person.mother === mother.name
+        : (person.mother === mother.name && person.sex === 'm');
+    });
+  });
```
xiimiik: we can do this with just 1 filter by rotating different flasks as above

src/averageAges.js
```diff
@@ -54,11 +63,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+    childs.forEach((child) => {
+      if (mother.name === child.mother) {
+        currentMother += child.born - mother.born;
+      }
+    });
+
+    return accumulator + currentMother;
+  }, 0) / childs.length).toFixed(2);
+}
```
xiimiik: if we do everything from above in 1 filter, then here we will `reduce` children

xiimiik (changes requested):

src/averageAges.js
```diff
@@ -14,12 +14,14 @@
...
-  // without nesting
+function calculateMenAverageAge(people, century = false) {
+  const males = people.filter(male =>
+    century
+      ? (male.sex === 'm' && Math.ceil(male.died / 100) === century)
+      : male.sex === 'm'
+  );
+
+  return calculateAverage(males);
 }
```
Esceype: In this function, you need to filter the array in two cases to get only men from it. In the next two functions, you also need to filter the array to get only a certain gender. Wouldn't it be better to create a function that would filter the array in advance and then iterate over it?

src/averageAges.js
```diff
@@ -36,8 +38,15 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
-function calculateWomenAverageAge(people, withChildren) {
+function calculateWomenAverageAge(people, withChildren = false) {
   // write code here
+  const females = people.filter(female =>
```
Esceype: Why is it `female` here, if it's just a person, you don't know what gender it is yet

Esceype (changes requested):
Good work, try to make corrections regarding the comments

src/averageAges.js
ydem1: In this task, it is stated that we have the average age of women with and without children, so the name is chosen accordingly

* Implement calculateWomenAverageAge function
 *
 * Function returns average age of women in array. If `withChildren` is
 * specified then function calculates average age only for women with children
 *
 * Hint: To check if a woman has children you should find someone who mention
 * her as mother.

src/averageAges.js
```diff
@@ -54,11 +65,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  , 0) / children.length;
+}
+
+function filterPeople(people, flag, flagIsTrue, flagIsfalse) {
+  return people.filter(person =>
+    flag
+      ? flagIsTrue(person)
+      : flagIsfalse(person)
+  );
+}
```
andriy-shymkiv, line 89: looks good 👌🏻

andriy-shymkiv (approved):
good job

PR #2294, opened 2023-08-08
sulution

valet2102:


src/averageAges.js
```diff
@@ -20,6 +20,22 @@ function calculateMenAverageAge(people, century) {
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  // const MEN = people.filter(person => century
+  //   ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+  //   : person.sex === 'm'
+  // );
```
desevoker:
```suggestion
```
You should delete all commented code and unnecessary comments.

src/averageAges.js
```diff
@@ -20,6 +20,22 @@ function calculateMenAverageAge(people, century) {
...
+
+  const MEN = century
```
desevoker:
```suggestion
  const men = century
```
Wrong naming. UPPERCASE_WITH_UNDERSCORES should only be used for literal constants that will never change, here it's just a variable that will never be reassigned. Such variables should be named in lowerCamelCase style.

src/averageAges.js
```diff
@@ -20,6 +20,22 @@ function calculateMenAverageAge(people, century) {
...
+    ? people.filter(person => person.sex === 'm'
+      && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === 'm');
+
+  const MEN_AGES = MEN
```
desevoker:
```suggestion
  const menAges = men
```
And in all other places throughout the code.

src/averageAges.js
```diff
@@ -38,6 +54,18 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const WEMEN = withChildren
```
desevoker:
```suggestion
  const women = withChildren
```
Spelling mistake.

src/averageAges.js
```diff
@@ -20,6 +20,22 @@ function calculateMenAverageAge(people, century) {
...
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  // const MEN = people.filter(person => century
+  //   ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+  //   : person.sex === 'm'
+  // );
+
+  const MEN = century
+    ? people.filter(person => person.sex === 'm'
+      && Math.ceil(person.died / 100) === century)
```
desevoker:
```suggestion
    ? people.filter(person => person.sex === SEX_MALE
      && Math.ceil(person.died / YEARS_IN_CENTURY) === century)
```
Consider replacing all magic values with constants. Declare them at the top of the file.

src/averageAges.js
```diff
@@ -38,6 +54,18 @@ function calculateMenAverageAge(people, century) {
  */
 function calculateWomenAverageAge(people, withChildren) {
   // write code here
+  const WEMEN = withChildren
+    ? people
+      .filter((person, i, persons) => person.sex === 'f'
+        && persons
```
desevoker:
```suggestion
      .filter((person) => person.sex === 'f'
        && people
```
You don't need `persons`, you already have `people` parameter which is visible anywhere inside the function, you can use it as many times as you need. And definitely don't need `i`.
And the same down below.

src/averageAges.js
```diff
@@ -38,6 +54,18 @@ function calculateMenAverageAge(people, century) {
...
+          .some(p => p.mother === person.name))
+    : people.filter(person => person.sex === 'f');
+
+  const WEMEN_AGES = WEMEN
+    .reduce((total, ages) => total + ages.died - ages.born, 0);
```
desevoker:
```suggestion
    .reduce((total, person) => total + person.died - person.born, 0);
```

src/averageAges.js
```diff
@@ -38,6 +54,18 @@ function calculateMenAverageAge(people, century) {
...
+  const AVARAGE_AGE = WEMEN_AGES / WEMEN.length.toFixed(2);
+
+  return AVARAGE_AGE;
```
desevoker:
```suggestion
  return calculateAverageAge(women);
```
You're doing the same calculations twice at this point, consider extracting that logic to a separate function.

desevoker (changes requested):
Goob job! 🚀

But some changes need to be made, check the comments.

PR #2296, opened 2023-08-08
Develop

maksym-mishchanchuk:


voievodik (changes requested):
Good job! 

But there is some requirements that was described in the function

maksym-mishchanchuk:
> Good job!
> 
> But there is some requirements that was described in the function


strange task check.

too abstract and not clear

voievodik:
> > Good job!
> > But there is some requirements that was described in the function
> 
> strange task check.
> 
> too abstract and not clear

I noted the points where there are problems
image

src/averageAges.js
```diff
@@ -15,11 +15,23 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const manOnly = people.filter(person => person.sex === 'm');
```
roman-mirzoian: It would be cool to put all the "magic values" in constants.
```js
const MALE = 'm';
const FEMALE = 'f';
```
This will improve code readability and help avoid using "magic values".

roman-mirzoian: If we need only one field from the entire object, we can use destructuring.

src/averageAges.js
```diff
@@ -15,11 +15,23 @@
...
+  let sumManAge;
+
+  if (!century) {
+    sumManAge = manOnly.reduce((sumAge, person) =>
+      sumAge + (person.died - person.born), 0);
```
roman-mirzoian, line 27: You count the total number of ages several times for the entire code, this functionality can be transferred to a separate function and used as a helper.

src/averageAges.js
```diff
@@ -55,10 +85,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const whoHasMothers = people.filter((person) => {
+    return onlyWithSon
+      ? (person.sex === 'm' && person.mother !== null)
+      : person.mother !== null;
+  });
+
+  const differenceAgeChildWithMother = [];
+
+  whoHasMothers.find((child) => {
```
roman-mirzoian: Are you sure you need the find method here? This method searches for and returns an element.

src/averageAges.js
```diff
@@ -37,7 +49,25 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
```
roman-mirzoian, line 60: For this function, the wishes are the same as for the previous one.

src/averageAges.js
```diff
@@ -55,10 +85,27 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
```
roman-mirzoian, line 95: For this function, the wishes are the same as for the previous one.

roman-mirzoian (changes requested):

BudnikOleksii (changes requested):
Looks like you forgot to push the corrected solution. Check the comments from previous reviewers.

roman-mirzoian (changes requested):
Currently, it is impossible to check the work, because of the `node_modules` pushed in the repository.

PR #2298, opened 2023-08-08
solution

AndrMar1939:


src/averageAges.js
```diff
@@ -15,11 +15,25 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  return people
+    .filter(({ sex, died }) => {
+      const deathCentury = Math.ceil(died / 100);
+      const boolean = century
```
AndriiKuz: boolean is not a good name for this variable.

AndriiKuz (approved):
LGTM

src/averageAges.js
AndrMar1939: yes, I agree with you

src/averageAges.js
```diff
@@ -15,11 +15,25 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  return people
+    .filter(({ sex, died }) => {
+      const deathCentury = Math.ceil(died / 100);
+      const isCenturyRequired = century
+        ? sex === 'm' && deathCentury === century
+        : sex === 'm';
```
roman-mirzoian: It would be cool to put all the "magic values" in constants. This applies to all code.

src/averageAges.js
```diff
@@ -15,11 +15,25 @@
...
+
+      return isCenturyRequired;
+    })
+    .reduce((acc, person, index, array) => {
+      const { born, died } = person;
+      const age = died - born;
+
+      if (index === array.length - 1) {
```
roman-mirzoian: 1. `array `- it is better not to give such general names.
2. Since our array does not change, we can access it directly without accepting an extra value from the arguments: `people.length`

src/averageAges.js
```diff
@@ -15,11 +15,25 @@
...
+  return people
+    .filter(({ sex, died }) => {
+      const deathCentury = Math.ceil(died / 100);
+      const isCenturyRequired = century
+        ? sex === 'm' && deathCentury === century
+        : sex === 'm';
+
+      return isCenturyRequired;
+    })
+    .reduce((acc, person, index, array) => {
```
roman-mirzoian: You make the same calculations in the `calculateWomenAverageAge` function - it could be transferred to a separate variable and reused as a helper.

src/averageAges.js
```diff
@@ -37,7 +51,23 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
```
roman-mirzoian, line 61: For this function, the wishes are the same as for the previous one.

src/averageAges.js
```diff
@@ -55,12 +85,35 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const peopleWithMother = people.filter(({ mother }) =>
+    mother !== null && people.find(person => person.name === mother));
```
roman-mirzoian: The `name` field can also be destructuring.

roman-mirzoian (changes requested):

src/averageAges.js
AndrMar1939: done

src/averageAges.js
AndrMar1939: done

src/averageAges.js
AndrMar1939: done

src/averageAges.js
AndrMar1939: done

src/averageAges.js
AndrMar1939: done

src/averageAges.js
```diff
@@ -14,12 +14,32 @@
...
+  return people
+    .filter(({ sex, died }) => {
+      const deathCentury = Math.ceil(died / CENTURY_VALUE);
+      const isCenturyRequired = century
+        ? sex === SEX_MALE && deathCentury === century
+        : sex === SEX_MALE;
+
+      return isCenturyRequired;
+    })
+    .reduce(getAverageAge, 0);
```
roman-mirzoian: It is better to completely transfer the calculation logic to another function and transfer only the array that needs to be calculated.
```js
const filteredMen = people.filter(...);
return getAverageAge(filteredMen);
...
const getAverageAge = filteredPeople => {
  return filteredPeople.reduce(...);
};
```
The same for `calculateWomenAverageAge` function.

roman-mirzoian (changes requested):
Good job!
There is quite a bit left.

src/averageAges.js
```diff
@@ -55,12 +84,35 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      const { born, mother: motherName } = child;
+      const mother = people.find(person => person.name === motherName);
+
+      const ageDiff = mother
+        ? born - mother.born
+        : 0;
+
+      if (index === filteredPeopleWithMotherBySex.length - 1) {
+        return (acc + ageDiff) / filteredPeopleWithMotherBySex.length;
+      }
```
BudnikOleksii, line 105:
```javascript
filteredPeopleWithMotherBySex.reduce((acc, child, index, array) => {
  if (index === array.length - 1) {
    // logic
  }
}
```
In reduce and other array iteration methods you have access to current array, from the callback parameter, so you don't need to point to filteredPeopleWithMotherBySex.

Also, I think it's not a good practice to do this calculation inside reduce. It's better to separate logic.

src/averageAges.js
```diff
@@ -14,12 +14,34 @@
...
+const CENTURY_VALUE = 100;
+
+const getAverageAge = (peopleArray) =>
+  peopleArray.reduce((acc, person, personIndex) => {
+    const { born, died } = person;
+    const age = died - born;
+
+    if (personIndex === peopleArray.length - 1) {
+      return (acc + age) / peopleArray.length;
+    }
```
BudnikOleksii, line 28: The same as I mentioned before

BudnikOleksii (approved):
Good job!
Left a few suggestions check them.

PR #2299, opened 2023-08-08
Add solution

FurArt:


src/averageAges.js
```diff
@@ -14,14 +14,40 @@
...
+
+  const men = people.filter((person) => {
+    return checkForMen(person) && checkCentryDied(person);
+  });
+
+  const amountMen = men.length;
+
+  const summAge = summAgeFunction(men);
+
+  return +(summAge / amountMen).toFixed(2);
```
volodymyr-soltys97: The use of Unary Plus (+) to convert a string to a number may lead to unexpected results if the string is not well formed. Consider using Number() instead.

src/averageAges.js
```diff
@@ -54,11 +102,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    function(accumulatorAge, kid) {
+      const mothers = people.find(mother => mother.name === kid.mother);
+
+      return accumulatorAge + (+kid.born) - (+mothers.born);
+    },
+    0);
+
+  const amountKid = filteredKid.length;
+
+  return +(summAgeDiff / amountKid).toFixed(2);
```
volodymyr-soltys97: You can create a helper function for this and use it everywhere

src/averageAges.js
```diff
@@ -14,14 +14,40 @@
...
+function calculateMenAverageAge(people, century = 0) {
   // write code here
   // learn how to use array methods like .filter .map .some .every .find .reduce
   // avoid using loop and forEach
   // replace `if ()` statement with &&, || or ?:
   // without nesting
+  function checkCentryDied(person) {
+    return century === 0
+      ? true
+      : Math.ceil(person.died / 100) === century;
```
volodymyr-soltys97: Think about how it can be rewritten so as not to prescribe clearly truе

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

anastasiiavorobiova (approved):
Good work!

PR #2300, opened 2023-08-08
add task solution

olya-girnyak:


src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person =>
+    person.sex === 'm'
+    && (!century || Math.ceil(person.died / 100) === century));
+
+  const menAverageAge = men.reduce((sum, data) =>
+    sum + data.died - data.born, 0) / men.length;
```
maxim2310: would be nice to make separate func as you use same here and in calculateWomenAverageAge

src/averageAges.js
```diff
@@ -55,10 +66,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const children = people.filter(person => people.find(child =>
+    child.name === person.mother)
+    && (!onlyWithSon
+    || (person.sex === 'm' && people.find(child =>
+      child.name === person.mother)))
+  );
```
maxim2310:
```suggestion
  const children = people.filter(person => people.find(child =>
    child.name === person.mother)
    && (!onlyWithSon
    || (person.sex === 'm'
  );
```

maxim2310: it is not very logical when the mother of a person is the name of the child

src/averageAges.js
```diff
@@ -55,10 +66,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const mothers = people.filter(person =>
+    person.sex === 'f'
+    && (!onlyWithSon || people.find(mommy => mommy.mother === person.name))
+  );
+
+  const averageAgeDiff = children.reduce((sum, data) =>
```
maxim2310:
```suggestion
  const averageAgeDiff = children.reduce((sum, child) =>
```

maxim2310 (changes requested):
Good work just need to make a couple of fixes

PR #2301, opened 2023-08-08
solved task

bormaxvel:
This task took me longer than usual

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -15,11 +15,19 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let filteredMen = people.filter(person => person.sex === 'm');
+
+  if (century) {
+    filteredMen = filteredMen.filter(
+      person => Math.ceil(person.died / 100) === century
+    );
+  }
+
+  const sum = filteredMen.reduce(
+    (total, person) => total + (person.died - person.born), 0
+  );
+
+  return sum / filteredMen.length;
 }
 
 /**
@@ -37,7 +45,19 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let filteredWomen = people.filter(person => person.sex === 'f');
+
+  if (withChildren) {
+    filteredWomen = filteredWomen.filter(person =>
+      people.some(child => person.name === child.mother)
+    );
+  }
+
+  const sum = filteredWomen.reduce(
+    (total, person) => total + (person.died - person.born), 0
+  );
+
+  return sum / filteredWomen.length;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/averageAges.js
```diff
@@ -15,11 +15,19 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let filteredMen = people.filter(person => person.sex === 'm');
```
SanyaBratashchuk:
```suggestion
  let men = people.filter(person => person.sex === 'm');
```

src/averageAges.js
```diff
@@ -15,11 +15,19 @@
...
+
+  if (century) {
+    filteredMen = filteredMen.filter(
+      person => Math.ceil(person.died / 100) === century
+    );
+  }
```
SanyaBratashchuk: image

SanyaBratashchuk: You can handle it in one filter's callback. For example: `people.filter(person => century ? person.sex === 'm' && getCentury(person.died) === century : person.sex === 'm')`

SanyaBratashchuk: Check other functions too

src/averageAges.js
```diff
@@ -55,10 +75,34 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const filteredWomen = people.filter(person =>
+    person.sex === 'f' && people.some(child => person.name === child.mother)
+  );
+
+  let countOfChildren = 0;
+
+  const sum = filteredWomen.reduce(
+    (total, mother) => {
+      // looking for all the children of the mother
+      let allChildren = people.filter(child => child.mother === mother.name);
```
SanyaBratashchuk:
```suggestion
      let children = people.filter(child => child.mother === mother.name);
```

src/averageAges.js
```diff
@@ -55,10 +75,34 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const filteredWomen = people.filter(person =>
+    person.sex === 'f' && people.some(child => person.name === child.mother)
+  );
+
+  let countOfChildren = 0;
```
SanyaBratashchuk:
```suggestion
```
You can use just array length

SanyaBratashchuk (changes requested):

src/averageAges.js
bormaxvel: I can't, because one mother can have several children, and I didn't create a general list of children

src/averageAges.js
```diff
@@ -55,10 +68,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      );
+
+      // calculate the sum of all age differences for each child
+      return total + children.reduce((totalOfMother, childOfMother) => {
+        countOfChildren++;
+
+        return totalOfMother
+          + childOfMother.born
+          - (people.find(person => person.name === childOfMother.mother).born);
+      }, 0
```
SanyaBratashchuk:
```suggestion
      return total + children.reduce((totalOfMother, childOfMother) => {
        return ( totalOfMother
          + childOfMother.born
          - (people.find(person => person.name === childOfMother.mother).born)) / children.length;
      }, 0
```
It will be the same I guess

src/averageAges.js
```diff
@@ -15,11 +15,15 @@
...
-  // without nesting
+  const men = people.filter(person => century
+    ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+    : person.sex === 'm');
+
+  const sum = men.reduce(
+    (total, person) => total + (person.died - person.born), 0
+  );
+
+  return sum / men.length;
```
SanyaBratashchuk: move this part to the separate function and reuse it in your main functions

SanyaBratashchuk (changes requested):

src/averageAges.js
bormaxvel: In this function, we don't have a single array to find out the number of operations. In order to know by what number to divide the sum, I simply made a counter instead of creating another array of children

src/averageAges.js
```diff
@@ -55,10 +70,33 @@ function calculateWomenAverageAge(people, withChildren) {
...
+        : child.mother === mother.name
+      );
+
+      // calculate sum of age difference between mother and each of her child
+      return total + children.reduce((totalOfMother, childOfMother) => {
+        countOfOperations++;
+
+        return totalOfMother
+          + childOfMother.born
+          - (people.find(person => person.name === childOfMother.mother).born);
```
TarasHoliuk: it seems like you are finding the mother of the child that you were found for this mother a few rows above. Long story short, the `mother` variable (argument of the first reducer) stored the value you trying to get here, so:
```suggestion
          - mother.born;
```

TarasHoliuk (changes requested):

src/averageAges.js
bormaxvel: Indeed, this is an unnecessary search operation

SanyaBratashchuk (approved):

PR #2308, opened 2023-08-09
solution

Yevheniia-Sid:


Orpheus29 (approved):
###Hello, good job!

Just a few tiny remarks:
- I guess that 'male' and 'female' constants can be capitalized, what do you think?..
- 'x' could be replaced with 'child' and then with 'mom' or smth like that; 'person' on lines 81-82 - with 'child' again. Just for better understanding, you know...
- '.filter' on line 85 be started on new line...

Yevheniia-Sid:
Thanks a lot!🤝 already made the fix😉

src/averageAges.js
```diff
@@ -15,11 +15,18 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const MALE = 'm';
```
roman-mirzoian: Such constants should be placed before the functions themselves, so as not to recreate them for each function separately.

src/averageAges.js
```diff
@@ -55,10 +73,23 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const MALE = 'm';
```
roman-mirzoian: Such constants should be placed before the functions themselves, so as not to recreate them for each function separately.

src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const MALE = 'm';
+  const men = people.filter(({ sex }) => sex === MALE);
+
+  const menDiedInCentury = century
+    ? men.filter(({ died }) => Math.ceil(died / 100) === century)
+    : men;
```
roman-mirzoian: If `century` is false, then you will simply overwrite the `men` array to `menDiedInCentury`, but this will not be completely correct in terms of the name.
```suggestion
  let men = people.filter(({ sex }) => sex === MALE);
  if (century) {
    men = men.filter(({ died }) => Math.ceil(died / 100) === century);
  }
```

src/averageAges.js
```diff
@@ -15,11 +15,18 @@
...
+
+  const avgAgeMen = menDiedInCentury
```
roman-mirzoian: Do not abbreviate names to unreadable ones.

src/averageAges.js
```diff
@@ -37,7 +44,18 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
```
roman-mirzoian, line 52: For this function, the wishes are the same as for the previous one.

src/averageAges.js
```diff
@@ -37,7 +44,18 @@ function calculateMenAverageAge(people, century) {
...
+  const FEMALE = 'f';
+  const women = people.filter(({ sex }) => sex === FEMALE);
+
+  const womenWithChild = withChildren
+    ? women.filter(woman => people.some(child => child.mother === woman.name))
+    : women;
+
+  const avgAgeWomen = womenWithChild
+    .map(({ born, died }) => died - born)
+    .reduce((sum, age) => sum + age, 0) / womenWithChild.length;
```
roman-mirzoian: You do similar actions in the previous function, so it can be taken into a separate function and used as a helper.

src/averageAges.js
```diff
@@ -55,10 +73,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const children = onlyWithSon
+    ? people.filter(({ sex }) => sex === MALE) : people;
+
+  const peopleWithMother = children
+    .map(child => {
+      child.momObject = people.find(mom => mom.name === child.mother);
+
+      return child;
+    })
+    .filter(child => child.momObject);
```
roman-mirzoian: You don't need to filter the function separately, you can search and return the person object directly through the filter.
Then it will not be necessary to use the extra addition of the momObject object.

src/averageAges.js
```diff
@@ -55,10 +73,23 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const MALE = 'm';
+  const children = onlyWithSon
```
```suggestion
  const childrens = onlyWithSon
```

src/averageAges.js
```diff
@@ -55,10 +73,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const peopleWithMother = children
+    .map(child => {
+      child.momObject = people.find(mom => mom.name === child.mother);
+
+      return child;
+    })
+    .filter(child => child.momObject);
+
+  const avgAgeDiff = peopleWithMother
```
roman-mirzoian: Do not abbreviate names to unreadable ones.

src/averageAges.js
```diff
@@ -55,10 +73,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    .reduce((prev, child) => prev + child.born - child.momObject.born, 0)
```
```suggestion
    .reduce((total, child) => prev + child.born - child.momObject.born, 0)
```

roman-mirzoian (changes requested):

src/averageAges.js
```diff
@@ -14,12 +14,25 @@
  *
  * @return {number}
  */
+const MALE = 'm';
+const FEMALE = 'f';
+
+function averageAgePeople(people) {
```
tiserett, line 20: function should call to action, for example 'calculateAverageAge'

src/averageAges.js
```diff
@@ -14,12 +14,25 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let men = people.filter(({ sex }) => sex === MALE);
+
+  if (century) {
+    men = men.filter(({ died }) => Math.ceil(died / 100) === century);
+  }
```
tiserett, line 31: let's use ternary

src/averageAges.js
```diff
@@ -37,7 +50,16 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let women = people.filter(({ sex }) => sex === FEMALE);
+
+  if (withChildren) {
+    women = women
+      .filter(woman => people.some(child => child.mother === woman.name));
+  }
```
tiserett, line 58: same with ternary here

tiserett (approved):

PR #2310, opened 2023-08-10
ad task solution

Vayts:


KatOlista (approved):
good job!

src/averageAges.js
```diff
@@ -1,64 +1,84 @@
...
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredArr = century
+    ? people.filter((person) => {
+      const personDeathCentury = Math.ceil(person.died / 100);
+
+      return century === personDeathCentury && person.sex === MALE;
+    })
+    : people.filter((person) => person.sex === MALE);
```
```suggestion
    const filteredPeopleByMan = people.filter((person) => {
      if (century) {
       const personDeathCentury = Math.ceil(person.died / 100);
       return century === personDeathCentury && person.sex === MALE;
      }
      
      return person.sex === MALE;
    });
```

src/averageAges.js
```diff
@@ -1,64 +1,84 @@
...
  *
  * @param {object[]} people
  * @param {boolean} withChildren - optional
  *
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const mothers = people.map(person => person.mother);
+  const filteredArr = withChildren
```
roman-mirzoian: Here you need to do the same refactoring as in the previous function.

src/averageAges.js
```diff
@@ -1,64 +1,84 @@
...
+    ? people.filter((person) => {
+      return mothers.includes(person.name) && person.sex === FEMALE;
+    })
+    : people.filter((person) => person.sex === FEMALE);
+
+  const ageSum = filteredArr.reduce((acc, person) => {
+    return acc + (person.died - person.born);
+  }, 0);
```
roman-mirzoian: Similar calculations are made in the previous function, this can be transferred to a separate function and used as a helper.

src/averageAges.js
```diff
@@ -1,64 +1,84 @@
...
+  const peopleWithMothers = people
+    .map((person) => {
+      const mother = people.find((human) => human.name === person.mother);
+
+      return {
+        ...person,
+        mother,
+      };
+    })
+    .filter((person) => person.mother);
```
roman-mirzoian: You don't need to filter the function separately, you can search and return the person object directly through the filter.

src/averageAges.js
```diff
@@ -1,64 +1,84 @@
...
+
+  const filteredArr = onlyWithSon
```
roman-mirzoian: It is necessary to choose a more meaningful naming.

roman-mirzoian (changes requested):

src/averageAges.js
```diff
@@ -1,64 +1,93 @@
...
+
+      return century === personDeathCentury && person.sex === MALE;
+    }
+
+    return person.sex === MALE;
+  });
+
+  const ageSum = calculateAgeSum(filteredPeopleByMan);
+
+  return ageSum / filteredPeopleByMan.length;
```
roman-mirzoian, line 32: Dividing by the length could be transferred to the `calculateAgeSum` function and called `calculateAverageAge`.

roman-mirzoian (approved):
Well done 👍

PR #2311, opened 2023-08-10
Solution

yuliiaant:


cty2802 (approved):
Good job!

src/averageAges.js
```diff
@@ -14,12 +14,21 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menFilteredAges = people
+    .filter(person => century
+      ? person.sex === 'm'
```

roman-mirzoian: For objects, it would be convenient to use destructuring.

src/averageAges.js
```diff
@@ -14,12 +14,21 @@
...
+        && Math.ceil(person.died / 100) === century
```
roman-mirzoian: The result of this calculation can be placed in a separate variable for better readability.

src/averageAges.js
```diff
@@ -37,7 +46,14 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
```
roman-mirzoian, line 32: The requirements for this function are the same as for the previous one.

src/averageAges.js
```diff
@@ -55,10 +71,16 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const filteredChildren = people.filter(person => onlyWithSon
+    ? people.some(mother => mother.name === person.mother
+      && person.sex === 'm')
+    : people.find(mother => mother.name === person.mother)
+  );
```
roman-mirzoian: Now everything is written correctly, but such code is difficult to read and maintain. Break it into smaller parts.

src/averageAges.js
```diff
@@ -55,10 +71,16 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ageDiffernce = filteredChildren.map(child =>
+    child.born - people.find(mother => mother.name === child.mother).born);
```
roman-mirzoian: Now everything is written correctly, but such code is difficult to read and maintain. Break it into smaller parts.

roman-mirzoian (changes requested):
Good job!
But you should not try to write the entire code "in one line", it makes it unreadable and if you need to change something in the future, it will be very difficult.
Try to break large parts of the code into smaller ones, choosing appropriate variable names that will explain what is happening in the code.
This applies to all code (all functions).

src/averageAges.js
```diff
@@ -1,64 +1,51 @@
...
- * @param {number} century - optional
- *
- * @return {number}
- */
+function calculateAverageAge(people) {
+  return people.reduce((acc, { died, born }) =>
+    acc + (died - born), 0) / people.length;
+}
+
+function isParent(people, person) {
```
roman-mirzoian: Function names should start with a verb.
You also return a whole object from this function, so it would be better to call this function `findMother`.

src/averageAges.js
```diff
@@ -1,64 +1,51 @@
...
+  return people.find(({ name }) => name === person.mother);
+}
+
+const calculateCentury = ({ died }) => Math.ceil(died / 100);
+
+const isMale = ({ sex }) => sex === 'm';
+const isFemale = ({ sex }) => sex === 'f';
```

src/averageAges.js
```diff
@@ -1,64 +1,51 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menFilteredAges = people
```
```suggestion
  const menFilteredByAges = people
```

src/averageAges.js
```diff
@@ -1,64 +1,51 @@
...
- * her as mother.
- *
- * @param {object[]} people
- * @param {boolean} withChildren - optional
- *
- * @return {number}
- */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenFilteredAges = people
```
```suggestion
  const womenFilteredByAges = people
```

src/averageAges.js
```diff
@@ -1,64 +1,51 @@
...
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const filteredChildren = people
+    .filter(person => onlyWithSon
+      ? isParent(people, person)
+        && isMale(person)
+      : isParent(people, person))
+    .map((child) => child.born - isParent(people, child).born);
+
+  return filteredChildren.reduce((a, b) => a + b) / filteredChildren.length;
```
roman-mirzoian: Now everything is written correctly, but such code is difficult to read and maintain. Break it into smaller parts.
For example, you filter children by mother in any case, so you could do something like this:
```js
let filteredChildren = people.filter(person => findMother(people, person));
if (onlyWithSon) {
  filteredChildren = filteredChildren.filter(child => isMale(child));
}
etc.
```

roman-mirzoian (changes requested):
Don't try to write everything in "one line" or in one long chain. This degrades the readability and maintainability of the code.

src/averageAges.js
```diff
@@ -1,64 +1,57 @@
...
+    .filter(person => findMother(people, person));
+
+  if (onlyWithSon) {
+    filteredChildren = filteredChildren.filter(child => isMale(child));
+  }
+
+  filteredChildren = filteredChildren
+    .map((child) => child.born - findMother(people, child).born);
+
+  return filteredChildren.reduce((a, b) => a + b) / filteredChildren.length;
```
BudnikOleksii, line 54: a and b names aren't informative in this case.

BudnikOleksii (approved):
Good job!

PR #2313, opened 2023-08-10
add solution

KatOlista:


src/averageAges.js
```diff
@@ -1,64 +1,57 @@
...
- *
- * @param {object[]} people
- * @param {number} century - optional
- *
- * @return {number}
- */
+const FEMALE = 'f';
+const MALE = 'm';
+
+function sumAges(whos) {
```
polinavafik: 'whos' kinda sounds weird, you can use 'people' (because we know than we sum ages of people so yknow)  so it'll be more understandable

src/averageAges.js
```diff
@@ -1,64 +1,57 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const allMen = people.filter(person => person['sex'] === MALE);
+
+  const centuryCadaversMen = century
+    ? allMen.filter(cadaver => Math.ceil(cadaver['died'] / 100) === century)
+    : false;
```
polinavafik: instead of returning false you can return 'allMen' so later in return you don't need to check anything,  you can just return your sumAges function like 
`return sumAges(centuryCadaversMen) / centuryCadaversMen.length `

but in that case, would be better to change the 'centuryCadaversMen' to something else))

src/averageAges.js
```diff
@@ -1,64 +1,57 @@
...
- * @param {number} century - optional
- *
- * @return {number}
- */
+const FEMALE = 'f';
+const MALE = 'm';
+
+function sumAges(whos) {
+  return whos.reduce((accum, person) =>
+    (accum + (person['died'] - person['born'])), 0);
```
polinavafik: you can change that function to calculating avgAge by adding `/ whos.length` after reduce so you don't need to divide by length later in your code

src/averageAges.js
```diff
@@ -1,64 +1,57 @@
...
- */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const allWomen = people.filter(person => person['sex'] === FEMALE);
+
+  const moms = withChildren
+    ? allWomen.filter(woman => {
+      return people.some(person => woman['name'] === person['mother']);
+    })
+    : false;
```
polinavafik: same situation like in 'calculateMenAverageAge'
insted of false you can return allWomen there

(but then you need to change 'moms' to something like filteredWomen))

polinavafik (commented):
Great job🎀 but i have few suggestions

polinavafik (changes requested):
i think you can improve your code a little bit

you can change 'sumAges' to calculating 'avgAge' by adding / whos.length after reduce so you don't need to divide by length later in your code

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
+      === mom['name'])['born'], 0)) / cadaverChildren.length).toFixed(2);
+};
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const allMen = people.filter(person => person['sex'] === MALE);
```
```suggestion
  const allMen = people.filter(({ sex })=> sex === MALE);
```

Moroz-Dmytro: It's better to do like this(`person.sex` not `person['sex']`), when it's possible
But even better would be to use destructuring
Please, fix this in all places

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
+
+  const cadaversMen = century
```
Moroz-Dmytro: Try to make up better name

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
- *
- * @param {object[]} people
- * @param {number} century - optional
- *
- * @return {number}
- */
+const FEMALE = 'f';
+const MALE = 'm';
+
+function sumAges(cadaver) {
```
Moroz-Dmytro: `getAvarageAge` is a better name for the function
Also, try to make up a better name, not `cadaver`

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
- * @param {object[]} people
- * @param {boolean} withChildren - optional
- *
- * @return {number}
- */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const allWomen = people.filter(person => person['sex'] === FEMALE);
+
+  const cadaversWomen = withChildren
```
Moroz-Dmytro: Make up better name

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
+    ? allWomen.filter(woman => {
+      return people.some(person => woman['name'] === person['mother']);
```
Moroz-Dmytro: If function return result without any other calculation you can simplify it, and avoid using `return`

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
+
+  // moms = withChildren
+  //   ? allWomen.filter(woman => {
+  //     return people.some(person => woman['name'] === person['mother']);
+  //   })
+  //   : false;
+
+  // return moms
+  //   ? sumAges(moms) / moms.length
+  //   : sumAges(allWomen) / allWomen.length;
```
Moroz-Dmytro: Delete comments

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
- */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const children = people.filter(child => {
+    return people.some((mother) => {
+      return child['mother'] === mother['name'];
+    });
```
Moroz-Dmytro: The same about `return`

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
+  });
+
+  const sons = children.filter(child => child['sex'] === MALE);
```
Moroz-Dmytro: You don't need to do these calculations, if `onlyWithSon` is `false

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
+
+  return onlyWithSon
+    ? +getAgeDiff(sons, people)
+    : +getAgeDiff(children, people);
```
Moroz-Dmytro: Try to change logic
To rewrite it like this:
`  return getAgeDiff(children, people)`

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
+function sumAges(cadaver) {
+  return cadaver.reduce((accum, person) =>
+    (accum + (person['died'] - person['born'])), 0) / cadaver.length;
+};
+
+function getAgeDiff(cadaverChildren, cadaver) {
+  return ((cadaverChildren.reduce((accum, child) =>
+    accum + child['born'] - cadaver.find(mom => child['mother']
+      === mom['name'])['born'], 0)) / cadaverChildren.length).toFixed(2);
+};
```
Moroz-Dmytro: It's difficult to read
Try to simplify it
Create a variable for `cadaverChildren.reduce(...)` (but make up a better name for `cadaverChildren`)
Create a variable for `cadaver.find(mom => child['mother']` inside reduce
In general split it into simpler parts, for easier reading

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const allMen = people.filter(person => person['sex'] === MALE);
+
+  const cadaversMen = century
+    ? allMen.filter(cadaver => Math.ceil(cadaver['died'] / 100) === century)
```
Moroz-Dmytro: 100 is also a "magic value"

Moroz-Dmytro (changes requested):
Good progress 👍 
But let's improve your code

src/averageAges.js
```diff
@@ -1,64 +1,50 @@
...
- * Divide year of person's death by 100: Math.ceil(person.died / 100)
- *
- * @param {object[]} people
- * @param {number} century - optional
- *
- * @return {number}
- */
+const FEMALE = 'f';
+const MALE = 'm';
+const HUNDRED = 100;
```
roman-mirzoian: In the context of this task, it would be better to call it `CENTURY_VALUE`.

src/averageAges.js
```diff
@@ -1,64 +1,50 @@
...
- * @param {boolean} onlyWithSon - optional
- *
- * @return {number}
- */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const children = people.filter(({ mother }) => people.some(({ name }) =>
```
roman-mirzoian, line 44:
```suggestion
  const childrens = people.filter(({ mother }) => people.some(({ name }) =>
```

src/averageAges.js
```diff
@@ -1,64 +1,50 @@
...
+function getAverageAge(people) {
+  return people.reduce((accum, { died, born }) =>
+    (accum + (died - born)), 0) / people.length;
+};
+
+function getAgeDiff(people1, people2) {
+  return ((people1.reduce((accum, { born, mother }) =>
+    accum + born - people2.find(({ name }) => mother
+      === name).born, 0)) / people1.length).toFixed(2);
+};
```
roman-mirzoian: Now everything is written correctly, but such code is difficult to read and maintain in the future. Let's make it a little simpler:
```suggestion
// we should name this function with "...Average..." because we return not different but literally average difference
function getAverageAgeDiff(childrens, people) { // we can name first arg "childrens" because "sons" also childrens
  const ageDiff = childrens.reduce((acc, { born, mother: motherName }) => {
    const motherBirthYear = people.find(({ name }) => motherName === name).born; // better to find this value separately is to improve code readability.
    return acc + born - motherBirthYear;
  }, 0);
  return +(ageDiff / childrens.length).toFixed(2); // if you move the `+` here, it will not be necessary to write it during the call and it will look better
};
```
I wrote the comments exclusively for you, you don't need to leave them in the code.

src/averageAges.js
```diff
@@ -1,64 +1,50 @@
...
- */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const children = people.filter(({ mother }) => people.some(({ name }) =>
+    mother === name));
+
+  const sons = children.filter(({ sex }) => sex === MALE);
```
roman-mirzoian: When the value of `onlyWithSon` is false, you make an extra filter and worsen the performance, it is better to first check whether this value is true and then do filtering.

roman-mirzoian (changes requested):

src/averageAges.js
```diff
@@ -1,64 +1,52 @@
...
- *
- * @return {number}
- */
+const FEMALE = 'f';
+const MALE = 'm';
+const CENTURY_VALUE = 100;
+
+function getAverageAge(people) {
+  return people.reduce((accum, { died, born }) =>
+    (accum + (died - born)), 0) / people.length;
```
Moroz-Dmytro, line 9:
```suggestion
  return people.reduce((accum, { died, born }) => (
    accum + (died - born)
  ), 0) / people.length;
```

src/averageAges.js
```diff
@@ -1,64 +1,52 @@
...
- *
- * @return {number}
- */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const allWomen = people.filter(({ sex }) => sex === FEMALE);
+
+  const women = withChildren
+    ? allWomen.filter(({ name }) => people.some(({ mother }) =>
+      name === mother))
```
Moroz-Dmytro, line 37:
```suggestion
    ? allWomen.filter(({ name }) => (
      people.some(({ mother }) => name === mother
    ))
```

Moroz-Dmytro (approved):
GJ 👍

PR #2314, opened 2023-08-10
task solution

viktoria-roik:


src/averageAges.js
```diff
@@ -14,12 +14,27 @@
...
  * @return {number}
  */
+
+const SEX_MALE = 'm';
+const SEX_FEMALE = 'f';
+const CENTURY_CHANGE = 100;
+const INITIAL_VALUE = 0;
+
+function getAverageAge(people) {
+  const totalAge = people.reduce((sum, person) =>
```
AntonHerasymov: you don't need variable totalAge you already can return  people.reduce( ... ) /  people.length;

src/averageAges.js
```diff
@@ -55,10 +76,26 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const ageDifferences = filteredPeople
+    .map(child => {
+      const mother = people.find(person => person.name === child.mother);
+
+      return (mother) ? child.born - mother.born : 0;
+    })
+    .filter(difference => difference > 0);
+
+  const totalAgeDifference = ageDifferences
```
AntonHerasymov: Same here

AntonHerasymov (changes requested):
Good job! but i have suggestion

src/averageAges.js
```diff
@@ -14,12 +14,26 @@
...
  */
+
+const SEX_MALE = 'm';
+const SEX_FEMALE = 'f';
+const CENTURY_CHANGE = 100;
+const INITIAL_VALUE = 0;
+
+function getAverageAge(people) {
+  return people
+    .reduce((sum, person) => sum + (person.died - person.born), INITIAL_VALUE)
```
roman-mirzoian: Here you can use object destructuring.

src/averageAges.js
```diff
@@ -55,10 +75,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const ageDifferences = filteredPeople
+    .map(child => {
+      const mother = people.find(person => person.name === child.mother);
+
+      return (mother) ? child.born - mother.born : 0;
+    })
+    .filter(difference => difference > 0);
+
+  return ageDifferences
+    .reduce((sum, age) => sum + age, INITIAL_VALUE) / ageDifferences.length;
```
roman-mirzoian: Now everything is written correctly, but you make a pass through the array three times, and here you can get by with one (this affects the performance a lot).
To do this, transfer all the logic to the middle of the reduce (search for the mother and calculation of the difference).

roman-mirzoian (changes requested):

src/averageAges.js
```diff
@@ -55,10 +75,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const filteredPeople = people.filter(person => (
+    onlyWithSon
+      ? person.sex === SEX_MALE
+      : true) && (people.some(child => child.name === person.mother)));
```
tiserett, line 81: weird formatting

tiserett (approved):

PR #2315, opened 2023-08-10
Solution

cty2802:


00Mass00 (approved):
Good job!

src/averageAges.js
```diff
@@ -14,12 +14,15 @@
...
  */
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const mens = people.filter((men) => men.sex === 'm'
+    && (century === undefined || Math.ceil(men.died / 100) === century));
```
BudnikOleksii: Don't compare with undefined/null or empty string. Use the logical not to check the falsy values, `!century`.

src/averageAges.js
```diff
@@ -36,8 +39,16 @@ function calculateMenAverageAge(people, century) {
...
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter((woman) => woman.sex === 'f'
+    && (withChildren === undefined
+    || people.find((motherName) => motherName.mother === woman.name)));
+
+  const averageWomensAge = Math.round(women.reduce((sum, currentPerson) => sum
+    + currentPerson.died - currentPerson.born, 0) / women.length * 100) / 100;
```
BudnikOleksii: We have the same callback for men to calculate age, let's create a separate function.

src/averageAges.js
```diff
@@ -54,11 +65,20 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const children = people.map((person) => (
+    {
+      ...person,
+      motherObject: people
+        .find((motherName) => motherName.name === person.mother),
+    }))
+    .filter(person => person.motherObject && (onlyWithSon === undefined
+  || person.sex === 'm'));
```
BudnikOleksii: It's difficult to read, motherObject looks strange, it can be just mother. Also, I think you actually don't need this transformation, you can just filter people that have mothers.

BudnikOleksii (changes requested):
Good job👍
Left some comments, please, check them.
Also, let's get rid of all "magic values" such as 'm', 'f', etc. Move them to constants.

src/averageAges.js
cty2802: about "It's difficult to read, motherObject looks strange, it can be just mother"  - I totally agree.
about "Also, I think you actually don't need this transformation, you can just filter people that have mothers." - in line 80 I get mother's "born", so it would be the second time I look for mother. I think searching for the mother just once is better, so I'm saving the first search results.

src/averageAges.js
```diff
@@ -14,12 +14,12 @@
...
  * @return {number}
  */
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const mens = people.filter((men) => men.sex === 'm'
```
roman-mirzoian: men - чоловіки, mens - чоловічий

src/averageAges.js
```diff
@@ -36,8 +36,19 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter((woman) => woman.sex === 'f'
+    && (withChildren === undefined
```
roman-mirzoian: Don't compare with` undefined`/`null` or empty string. Use the logical not to check the falsy values.

src/averageAges.js
```diff
@@ -36,8 +36,19 @@ function calculateMenAverageAge(people, century) {
...
+    || people.find((motherName) => motherName.mother === woman.name)));
+
+  return getAverageAge(women);
+}
+
+function getAverageAge(genderList) {
+  return Math.round(genderList.reduce((sum, currentPerson) => sum
+    + currentPerson.died - currentPerson.born, 0)
+    / genderList.length * 100) / 100;
```
roman-mirzoian: Everything is written correctly now, but such a feature is difficult to read and maintain in the future.
Let's break it down into smaller parts.
```suggestion
function getAverageAge(people) {
  const totalAge = people.people((total, { died, born }) => total + died - born, 0);
  return +(totalAge / people.length).toFixed(2);
```

src/averageAges.js
```diff
@@ -54,11 +65,20 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const children = people.map((person) => (
+    {
+      ...person,
+      motherInfo: people
+        .find((motherName) => motherName.name === person.mother),
+    }))
+    .filter(person => person.motherInfo && (onlyWithSon === undefined
+  || person.sex === 'm'));
```
roman-mirzoian: It's written correctly now, but it's hard to read and maintain in the future.
We can break the code into several parts:
1. Filter the `people` array and leave only those objects in which there are `mothers`.
2. If `onlyWithSon` is `true` - filter the `children` array and leave only `male` objects there.
3. When calculating the `average age`, `find` for the `mother` for each `child` in the middle of the `reduce` method and then perform the calculation.

roman-mirzoian (changes requested):
Don't forget to convert static variables to constants to avoid using "magic values".
Also, do not try to write everything in "one line" or in one chain, this worsens the readability and maintainability of the code.

BudnikOleksii (changes requested):
Looks like you forgot to push fixes in accordance with suggestions before requesting a review. Please, check the previous interviewer's comments.

src/averageAges.js
```diff
@@ -36,8 +39,26 @@ function calculateMenAverageAge(people, century) {
...
+    && (!withChildren
+    || people.find((motherName) => motherName.mother === woman.name)));
+
+  return getAverageAge(women);
+}
+
+function getAverageAge(genderList) {
+  // return Math.round(genderList.reduce((sum, currentPerson) => sum
+  //   + currentPerson.died - currentPerson.born, 0)
+  //   / genderList.length * 100) / 100;
```
BudnikOleksii, line 54: Remove comments

src/averageAges.js
```diff
@@ -54,11 +75,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  let children = people
+    .filter(person => people
+      .find((motherName) => motherName.name === person.mother));
+
+  if (onlyWithSon) {
+    children = children.filter(person => person.sex === MAN_GENDER);
+  }
```
BudnikOleksii, line 86: You can filter with this condition in the first filter.

src/averageAges.js
```diff
@@ -54,11 +75,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  return Math.round(children.reduce((sum, person) => {
+    const mother = people
+      .find((motherName) => motherName.name === person.mother);
```
BudnikOleksii, line 90:
```suggestion
      .find((mother) => mother.name === person.mother);
```
motherName.name looks strange, just mother

BudnikOleksii (approved):
Good job!

PR #2320, opened 2023-08-10
solution

OlhaMomot:


src/averageAges.js
```diff
@@ -37,7 +40,17 @@ function calculateMenAverageAge(people, century) {
...
+  const ages = women.reduce((sum, { died, born }) => sum + (died - born), 0);
+  const womenWithChildren = women.filter(woman => {
+    return people.some(person => person.mother === woman.name);
+  });
+  const agesWithChildren = womenWithChildren
+    .reduce((sum, { died, born }) => sum + (died - born), 0);
+
+  return withChildren
+    ? agesWithChildren / womenWithChildren.length
+    : ages / women.length;
```
ihorid: You could make a function (for lines 51 and 25)

ihorid (approved):
Good job!

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === 'm');
```

src/averageAges.js
```diff
@@ -15,11 +15,14 @@
...
+  const ages = men.reduce((sum, { died, born }) => sum + (died - born), 0);
+  const menByCentury = men
+    .filter(({ died }) => Math.ceil(died / 100) === century);
+  const agesByCentury = menByCentury
+    .reduce((sum, { died, born }) => sum + (died - born), 0);
+
+  return century ? (agesByCentury / menByCentury.length) : (ages / men.length);
```
roman-mirzoian: Now everything is written correctly, but when our `century` value is false, you are doing unnecessary calculations. Try to check this condition first, and then filter if necessary. Fix this for all functions.

Also, the calculation itself looks similar to what is done next for women, so the repeated part can be taken out into a separate function and reused.

roman-mirzoian (changes requested):
Try to check the necessary conditions first, and then make calculations based on these conditions to improve the performance of your code.

src/averageAges.js
```diff
@@ -14,12 +14,23 @@
  *
  * @return {number}
  */
+const MALE_SIGN = 'm';
+const FEMALE_SIGN = 'f';
+
+function countAverageAge(arr) {
+  return arr.reduce((sum, { died, born }) =>
+    sum + (died - born), 0) / arr.length;
```
```suggestion
  return arr.reduce((sum, { died, born }) => (
    sum + (died - born)
  ), 0) / arr.length;
```

src/averageAges.js
```diff
@@ -14,12 +14,23 @@
...
+}
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => {
+    return century
```
Moroz-Dmytro: We don't need `return` if function returns result without any calculations
`() => { return ... } === () => ...`
Same in all code

src/averageAges.js
```diff
@@ -14,12 +14,23 @@
...
+      ? person.sex === MALE_SIGN && Math.ceil(person.died / 100) === century
+      : person.sex === MALE_SIGN;
+  });
+  const averageAge = countAverageAge(men);
```
Moroz-Dmytro: We can just return result without creation useless variable

src/averageAges.js
```diff
@@ -55,10 +74,19 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const children = people.filter(person => {
+    return onlyWithSon
+      ? people.some(({ name }) => name === person.mother
+        && person.sex === MALE_SIGN)
+      : people.some(({ name }) => name === person.mother);
+  });
+
+  const ageDifference = children.reduce((sum, child) => {
+    return sum + (child.born - people.find(({ name }) =>
+      name === child.mother).born);
```
Moroz-Dmytro: To make it easier to read we can create constant value for `people.find(({ name }) =>
      name === child.mother).born`

Moroz-Dmytro (changes requested):
Good work 👍 
Small comments and it's perfect 🚀

Moroz-Dmytro (approved):
LGTM 🔥

PR #2321, opened 2023-08-10
Solution

neovendano:


src/averageAges.js
```diff
@@ -55,10 +82,30 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const MALE_SIGN = 'm';
+
```
OlhaMomot: You use this const few times in your code. Maybe, it make sense to put it at the very beggining of the script and just reuse?

OlhaMomot (approved):
Your code is clear and realy easy to understand👍 Good job!

src/averageAges.js
```diff
@@ -15,11 +15,22 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const MALE_SIGN = 'm';
+  const CENTURY_SIZE = 100;
```
Moroz-Dmytro: All constants should be places outside the function, at the top of the file

src/averageAges.js
```diff
@@ -15,11 +15,22 @@
...
+
+  if (century) {
+    males = males.filter(({ died }) => {
+      return Math.ceil(died / CENTURY_SIZE) === century;
+    });
+  }
+
+  const ages = males.map(({ born, died }) => died - born);
+  const maleTotalAge = ages.reduce((accumulator, age) => accumulator + age, 0);
+  const maleAverageAge = maleTotalAge / ages.length;
```
Moroz-Dmytro: Remember about "DRY". You have the same logic in the second function

src/averageAges.js
```diff
@@ -37,7 +48,23 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const FEMALE_SIGN = 'f';
+
+  let females = people.filter(({ sex }) => sex === FEMALE_SIGN);
+
+  if (withChildren) {
+    females = females.filter(({ name }) => {
+      return people.some(({ mother }) => mother === name);
```
Moroz-Dmytro, line 65: If function returns result without any calculations, you can simplify it and avoid using `return`

src/averageAges.js
```diff
@@ -55,10 +82,30 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const MALE_SIGN = 'm';
+
+  let children = people.filter(({ mother }) => {
+    return people.some(({ name }) => name === mother);
```
Moroz-Dmytro, line 90: The same about `return`

src/averageAges.js
```diff
@@ -55,10 +82,30 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+    return born - motherBorn;
+  });
+
+  const totalAgeDiff = ageDiff.reduce((accumulator, age) => {
+    return accumulator + age;
+  }, 0);
+  const averageAgeDiff = totalAgeDiff / ageDiff.length;
+
+  return averageAgeDiff;
```
Moroz-Dmytro: We can avoid creating useless variable, can just return result

Moroz-Dmytro (changes requested):
Good progress 🚀 
But some comments below

src/averageAges.js
neovendano: Sure. The only reason I did it is to make line shorter, because

females = females.filter(({ name }) => people.some(({ mother }) => mother === name));

is longer than 80 chars.

src/averageAges.js
```diff
@@ -37,7 +54,17 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  let females = people.filter(({ sex }) => sex === FEMALE_SIGN);
+
+  if (withChildren) {
+    females = females.filter(({ name }) => {
+      return people.some(({ mother }) => mother === name);
+    });
+  }
+
+  const ages = females.map(({ born, died }) => died - born);
```
roman-mirzoian: You do a similar calculation twice, you should move it to the middle of the `averageOfArray` function and to the middle of `reduce`, to avoid traversing the array twice when it can be done once.

src/averageAges.js
```diff
@@ -14,12 +18,25 @@
  *
  * @return {number}
  */
+
+function averageOfArray(array) {
```
roman-mirzoian: Try to add more meaningful names and function names should start with a verb.
```suggestion
function getAverageAge (ages) { 
```
If you transfer the repeated part of the logic and transfer the entire array here, then the argument can be called "people".

src/averageAges.js
```diff
@@ -14,12 +18,25 @@
...
+  const arraySum = array.reduce((accumulator, el) => accumulator + el, 0);
```
```suggestion
  const totalAge = ages.reduce((acc, currentAge) => acc + currentAge, 0);
```

src/averageAges.js
```diff
@@ -55,10 +82,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  let children = people.filter(({ mother }) => {
+    return people.some(({ name }) => name === mother);
+  });
+
+  if (onlyWithSon) {
+    children = children.filter(({ sex }) => sex === MALE_SIGN);
+  }
+
+  const ageDiff = children.map(({ born, mother }) => {
+    const motherBorn = people.find(({ name }) => {
```
```suggestion
    const motherBirthYear = people.find(({ name }) => {
```

roman-mirzoian (changes requested):

src/averageAges.js
neovendano: Sure, I can move that calculation to some function, but I won't agree that I should move it exactly to `averageOfArray`, because i call `averageOfArray` inside of `calculateAverageAgeDiff` function, while this calculation is not used in `calculateAverageAgeDiff`.

PR #2325, opened 2023-08-11
solution

00Mass00:


src/averageAges.js
```diff
@@ -1,64 +1,74 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const isMale = person => person.sex === 'm';
+
+  const man = people.filter((person) =>
```
BudnikOleksii: Move all helpers outside as it's not a good practice to create functions inside a function.
isMale looks like a result instead of a function. I think a better name is checkIsMale or something like this.
Also, man is only one, men - plural

src/averageAges.js
```diff
@@ -1,64 +1,74 @@
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter((person) =>
+    person.sex === 'f' && (!withChildren
+    || people.some((child) => child.mother === person.name))
+  );
+
+  const totalWomenAge = women.reduce((sum, person) =>
+    sum + (person.died - person.born), 0
+  );
```
BudnikOleksii: Looks like you use the same callback here, let's create a helper

src/averageAges.js
```diff
@@ -1,64 +1,74 @@
...
- *
  * @param {object[]} people
  * @param {boolean} withChildren - optional
  *
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = people.filter((person) =>
+    person.sex === 'f' && (!withChildren
```
BudnikOleksii: Let's get rid of magic values such as 'f', 'm'

src/averageAges.js
```diff
@@ -1,64 +1,74 @@
...
  * @param {boolean} onlyWithSon - optional
  *
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const isMale = person => person.sex === 'm';
```
BudnikOleksii: You have this helper already, so the one more reason to keep them outside

src/averageAges.js
```diff
@@ -1,64 +1,74 @@
...
+
+  const ageDifferences = relevantPeople.map(person => {
+    const mother = people.find(candidateMother =>
+      candidateMother.name === person.mother
+    );
+
+    return mother
+      ? person.born - mother.born
+      : null;
+  }).filter(ageDiff => ageDiff !== null);
```
BudnikOleksii: [CODE STYLE] - All chained methods called on the same level should have same indentation. This makes code much more readable.
You don't need to compare with null, you can just return ageDiff in the filter and it will check if the value exists.

BudnikOleksii (changes requested):
Good job!

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
- * To calculate century:
- * Divide year of person's death by 100: Math.ceil(person.died / 100)
- *
- * @param {object[]} people
- * @param {number} century - optional
- *
- * @return {number}
- */
+function checkIsMale(person) {
+  return person.sex === 'm';
```

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter((person) =>
+    checkIsMale(person) && (!century
+    || Math.ceil(person.died / 100) === century)
```
roman-mirzoian: 100 is also "magic value".

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
+  const women = people.filter((person) =>
+    checkIsFemale(person) && (!withChildren
+    || people.some((child) => child.mother === person.name))
+  );
+
+  const totalWomenAge = women.reduce((sum, person) =>
+    sum + calculateAge(person), 0
+  );
+
+  return women.length ? totalWomenAge / women.length : 0;
```
roman-mirzoian: You make similar calculations for `men`, so it is worth taking it to a separate function and using it as a helper.

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
+
+  const ageDifferences = relevantPeople.map(person => {
+    const mother = people.find(candidateMother =>
+      candidateMother.name === person.mother
+    );
+
+    return mother
+      ? person.born - mother.born
+      : mother;
+  }).filter(ageDiff => ageDiff);
```
roman-mirzoian: Transfer this logic to the last reduce so as not to go through the array one more time and improve the performance of calculations.

src/averageAges.js
```diff
@@ -1,64 +1,63 @@
...
- * @param {boolean} onlyWithSon - optional
- *
- * @return {number}
- */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const relevantPeople = onlyWithSon
```
roman-mirzoian: Relevant people by what?
It is worth choosing a more meaningful name.

roman-mirzoian (changes requested):

src/averageAges.js
```diff
@@ -1,64 +1,97 @@
...
+    : people;
+
+  const { totalAgeDifference, ageDifferencesCount } = peopleWithSon.reduce(
+    (accumulator, person) => {
+      const mother = people.find(candidateMother =>
+        candidateMother.name === person.mother
+      );
+
+      if (mother) {
+        const ageDiff = person.born - mother.born;
```
roman-mirzoian, line 78:
```suggestion
    (acc, { born, mother: motherName }) => {
      const motherBirthYear = people.find(({ name }) => name === motherName).born;

      const ageDiff = born - motherBirthYear;
```

roman-mirzoian (approved):
Well done 👍

PR #2326, opened 2023-08-11
initial commit

maksym-mishchanchuk:


src/averageAges.js
```diff
@@ -14,12 +14,26 @@
...
+
+  if (!century) {
+    sumManAge = manOnly.reduce((sumAge, person) =>
+      sumAge + (person.died - person.born), 0);
+
+    return +(sumManAge / manOnly.length).toFixed(2);
+  }
+
+  const menInCentury = manOnly.filter(man =>
+    Math.ceil(man.died / 100) === century);
```
FrontKid: Take out the number 100 in a separate variable
u name it like AGE__DIVIDER

FrontKid (approved):
Good Job BRO!

src/averageAges.js
BudnikOleksii: Agree, let's move it to a variable.

BudnikOleksii: You can filter by century conditionally in the first filter and then you can get rid of `if ()` statement

src/averageAges.js
```diff
@@ -36,8 +50,31 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+
+function agesSumPerson(avaregeAgesPeople) {
```
BudnikOleksii: It's better to name something like getTotalPeopleAge in this case or something similar

src/averageAges.js
```diff
@@ -36,8 +50,31 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const womenOnly = people.filter(({ sex }) => sex === FEMALE);
+  let sumWomenAge;
+
+  if (!withChildren) {
+    sumWomenAge = womenOnly.reduce((sumAge, person) =>
+      sumAge + (person.died - person.born), 0);
+
+    return +(sumWomenAge / womenOnly.length).toFixed(2);
+  }
```
BudnikOleksii: The same as above, let's filter conditionally and get rid of `if` statement.

src/averageAges.js
```diff
@@ -55,10 +92,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const mothers = people.filter(person => person.sex === FEMALE);
+
+  const whoHasMothers = people.filter((person) => {
+    return onlyWithSon
+      ? (person.sex === MALE && person.mother !== null)
+      : person.mother !== null;
```
```suggestion
      ? (person.sex === MALE && person.mother)
      : person.mother;
```
You don't need to compare it with null, just check if exists.
Also, it's better to name peopleWithMothers or something like this instead of whoHasMothers.

src/averageAges.js
```diff
@@ -55,10 +92,27 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const differenceAgeChildWithMother = [];
+
+  whoHasMothers.filter((child) => {
+    mothers.forEach((mom) => {
+      if (mom.name === child.mother) {
+        differenceAgeChildWithMother.push(child.born - mom.born);
+      }
+    });
+  }
+  );
```
BudnikOleksii: filter like this makes no sense, you can do all calculations directly in reduce

BudnikOleksii (changes requested):
Nice work!

src/averageAges.js
```diff
@@ -14,12 +14,21 @@
...
+const AGE_DIVIDER = 100;
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const manOnly = people.filter((person) => {
+    return !century
```
Moroz-Dmytro: You don't need `return` if function just returns result

src/averageAges.js
```diff
@@ -14,12 +14,21 @@
...
+      ? person.sex === MALE
+      : century === Math.ceil(person.died / AGE_DIVIDER) && person.sex === MALE;
+  });
+
+  const sumManAge = manOnly.reduce((sumAge, person) =>
+    sumAge + (person.died - person.born), 0);
+
+  return +(sumManAge / manOnly.length).toFixed(2);
```
Moroz-Dmytro: Remember about DRY. You have the same logic in the second function. Move this part to separate function

src/averageAges.js
```diff
@@ -55,10 +73,17 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const peopleWithMothers = people.filter((person) => {
+    return (onlyWithSon
+      ? (person.sex === MALE)
+      : true
+    ) && people.some((mom) => mom.name === person.mother);
```
```suggestion
    return (!onlyWithSon || person.sex === MALE
      && people.some((mom) => mom.name === person.mother)
    );
```

src/averageAges.js
```diff
@@ -55,10 +73,17 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  });
+
+  return peopleWithMothers
+    .reduce((acc, person) => acc + person.born - people
+      .find((mom) => mom.name === person.mother).born, 0)
+    / peopleWithMothers.length;
```
Moroz-Dmytro: Don't write this into one line, it's difficult to read. Better will be to create variable for `.find((mom) => mom.name === person.mother).born`

Moroz-Dmytro (changes requested):
GJ 👍 
But check comments below

src/averageAges.js
```diff
@@ -14,12 +14,26 @@
...
+  const manOnly = people.filter((person) => (
+    !century
+      ? (person.sex === MALE)
+      : (century
+        === Math.ceil(person.died / AGE_DIVIDER) && person.sex === MALE)
+  ));
+
+  const sumManAge = sumAges(manOnly);
+
+  return +(sumManAge / manOnly.length).toFixed(2);
```
roman-mirzoian: Move the logic of the calculation completely into a separate function, because it is repeated also in the next function.

src/averageAges.js
```diff
@@ -14,12 +14,26 @@
...
+    sumAge + (person.died - person.born), 0);
+}
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const manOnly = people.filter((person) => (
```
```suggestion
  const menOnly = people.filter((person) => (
```

src/averageAges.js
```diff
@@ -55,10 +77,18 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      && people.some((mom) => mom.name === person.mother)
+    );
+  });
+
+  return peopleWithMothers.reduce((acc, person) => {
+    const mother = people.find((mom) => mom.name === person.mother);
+    const ageDifference = person.born - mother.born;
+
+    return acc + ageDifference;
+  }, 0) / peopleWithMothers.length;
```
```suggestion
  const ageDiff = peopleWithMothers.reduce((acc, { born, mother: motherName }) => {
    const motherBirthYear = people.find(({ name }) => name === motherName);
    return acc + born - motherBirthYear;
  }, 0);
  return ageDiff / peopleWithMothers.length;
```

roman-mirzoian (changes requested):

BudnikOleksii (approved):
Nicely done!

PR #2328, opened 2023-08-11
create age calculator func

rostyslav48:


src/averageAges.js
```diff
@@ -14,12 +14,22 @@
...
  * @return {number}
  */
-function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+function calculateMenAverageAge(people, century = false) {
+  const MALE_SEX = 'm';
```
Moroz-Dmytro: Move all constants outside the functions(to the top of file)

src/averageAges.js
```diff
@@ -14,12 +14,22 @@
...
+  const menFilteredArray = people.filter(({ sex, died }) => {
```
Moroz-Dmytro: Avoid using words 'array' 'object' and so on in variable names

src/averageAges.js
```diff
@@ -14,12 +14,22 @@
...
+    const personCentury = Math.ceil(died / 100);
+    const isSuitableCenruty = century ? century === personCentury : true;
```
```suggestion
    const isSuitableCenruty = !century || Math.ceil(died / 100);
```

src/averageAges.js
```diff
@@ -14,12 +14,22 @@
...
+    return sex === MALE_SEX && isSuitableCenruty;
+  });
+
+  const menSumAge = menFilteredArray.reduce((averageAge, { born, died }) => {
+    const age = died - born;
+
+    return averageAge + age;
+  }, 0);
+
+  return menSumAge / menFilteredArray.length;
```
Moroz-Dmytro: Remember about "DRY". You have the same logic in the second function

src/averageAges.js
Moroz-Dmytro: And 100 is also a "magic value"

src/averageAges.js
```diff
@@ -54,11 +80,24 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
-function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+function calculateAverageAgeDiff(people, onlyWithSon = false) {
+  const MALE_SEX = 'm';
+  const childrenWithMothers = people.filter(({ mother, sex }) => {
+    const isMotherInList = people.find(({ name }) => name === mother);
```
Moroz-Dmytro: `find` doesn't return `boolean` value, so try to make up better name(that not starts with `is`)

Moroz-Dmytro (changes requested):
Good progress 👍 
But small comments

src/averageAges.js
```diff
@@ -14,12 +14,28 @@
...
+  const men = people.filter(({ sex, died }) => {
+    const personCentury = Math.ceil(died / ONE_CENTURY);
+    const isSuitableCenruty = century ? century === personCentury : true;
+
+    return sex === MALE_SEX && isSuitableCenruty;
+  });
+
+  const menSumAge = calculateAgeSum(men);
+
+  return menSumAge / men.length;
```
Moroz-Dmytro, line 38: This logic - `/men.length` could also be moved to `calculateAgeSum` function(but renaming would be needed)

Moroz-Dmytro (approved):
GJ 👍 
But small comment

PR #2331, opened 2023-08-12
Solution

Oleh0293:


k1NAs (commented):
GJ! :muscle:

Taras2705 (commented):
looks good

NazarVotkanych (approved):
LGTM!

src/averageAges.js
```diff
@@ -14,51 +14,57 @@
...
+      ? person.sex === MALE
+      : true) && (people.some(child => child.name === person.mother)
+    ));
+
+  const arrayOfChildMotherYears = peopleWithMothers.map(person =>
+    ({
+      died: person.born,
+      born: people.find(mother => person.mother === mother.name).born,
+    })
+  );
```
BudnikOleksii, line 65: Naming is a bit strange. Also `died: person.born`, they die when born and born when their mother is born?💀

I understand that you changed an array to reuse the "calculateAverageAge" function, but it looks confusing, it's better don't do that and calculate separately without this function so it will be more clear.

BudnikOleksii (changes requested):
Good job!
Let's change the "calculateAverageAgeDiff" function to avoid misunderstanding.

PR #2332, opened 2023-08-12
solution averageAges

ihorid:


thaiduk (changes requested):
Hi. On my mind it's great job.
But I think, it's better make some const for all tasks, like const CENTURY, and sex: MALE / FEMALE.

src/averageAges.js
```diff
@@ -1,5 +1,15 @@
 'use strict';
 
+const MALE = 'm';
+const FEMALE = 'f';
+const CENTURY = 100;
+
+function getAverengeAge(listOfPeople) {
```
```suggestion
function getAverengeAge(people) {
```

src/averageAges.js
```diff
@@ -1,5 +1,15 @@
...
+  const sumAge = listOfPeople.map((person) => person.died - person.born)
+    .reduce((acc, value) => acc + value, 0);
```
roman-mirzoian: Move the calculation from the `map` inside the `reduce`, so as not to go through the array twice.
Try using destructuring on the object.

src/averageAges.js
```diff
@@ -14,12 +24,15 @@
...
  * @return {number}
  */
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const menList = people.filter((person) => person.sex === MALE);
```
```suggestion
  const filteredMen = people.filter((person) => person.sex === MALE);
```

roman-mirzoian: Try using destructuring on the object.

src/averageAges.js
```diff
@@ -14,12 +24,15 @@
...
+  const menFromCentury = menList
+    .filter((man) => Math.ceil(man.died / CENTURY) === century);
+
+  return (!century)
+    ? getAverengeAge(menList)
+    : getAverengeAge(menFromCentury);
```
roman-mirzoian: If `century` will have a false value, then you will be doing extra filtering. 
Move the check of this value to the `menFromCentury` filtering, so as not to make an unnecessary loop through the array.

src/averageAges.js
```diff
@@ -36,8 +49,16 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
```
roman-mirzoian, line 50: The comments for this function are the same as for the previous one.

src/averageAges.js
```diff
@@ -55,10 +76,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    peopleWhitMother.some(child => mother.name === child.mother));
+
+  const sumOfDifference = peopleWhitMother
+    .map(child => {
+      const motherOfPerson = womenMother
+        .find(woman => woman.name === child.mother);
+
+      return child.born - motherOfPerson.born;
+    })
+    .reduce((acc, value) => acc + value, 0);
```
roman-mirzoian: Move the calculation from the `map` inside the `reduce`, so as not to go through the array twice.

src/averageAges.js
```diff
@@ -55,10 +76,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const peopleWhitMother = people.filter(person => !onlyWithSon
+    ? people.find(women => person.mother === women.name)
+    : (people.find(women => person.mother === women.name
+      && person.sex === MALE)
+    ));
+
+  const womenMother = people.filter(mother =>
```
```suggestion
  const mothers = people.filter(mother =>
```

src/averageAges.js
```diff
@@ -55,10 +76,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const peopleWhitMother = people.filter(person => !onlyWithSon
+    ? people.find(women => person.mother === women.name)
+    : (people.find(women => person.mother === women.name
+      && person.sex === MALE)
+    ));
```
roman-mirzoian: To simplify this code, you can first filter all `people` separately and leave only those who have mothers, and then check if `onlyWithSon` equal to trues, then filter that those who remain are only men.

roman-mirzoian (changes requested):

src/averageAges.js
```diff
@@ -14,12 +23,14 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredMen = people.filter(({ sex }) => sex === MALE)
+    .filter(({ died }) => century
+      ? Math.ceil(died / CENTURY) === century
+      : true);
```
Moroz-Dmytro: Would be better to combine it to one `filter`

src/averageAges.js
```diff
@@ -55,10 +72,22 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const peopleWhitMother = people.filter(person =>
+    people.find(women => person.mother === women.name))
+    .filter(person => onlyWithSon ? person.sex === MALE : true);
```
Moroz-Dmytro: Also would be nice to combine it to 1 `filter`

src/averageAges.js
```diff
@@ -55,10 +72,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const mothers = people.filter(mother =>
+    peopleWhitMother.some((child) => mother.name === child.mother));
```
```suggestion
  const mothers = people.filter(mother => (
    peopleWhitMother.some((child) => mother.name === child.mother)
  ));
```

Moroz-Dmytro (changes requested):
GJ 👍 
Small fixes and it's perfect 🚀

BudnikOleksii (approved):
Nicely done!

PR #2333, opened 2023-08-12
solution

helenmarth:


src/averageAges.js
```diff
@@ -54,11 +69,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const children = people
+    .filter(person => people.find(mother => mother.name === person.mother)
+    && (onlyWithSon ? person.sex === 'm' : person));
+
+  const ageDiff = children
+    .map(child => {
+      const mother = people
+        .find(mom => child.mother === mom.name);
+
+      return child.born - mother.born;
```
etojeDenys: mother can be undefined here
```suggestion
      const mother = people
        .find(mom => child.mother === mom.name);

      return child.born - (mother?.born || 0);
```

src/averageAges.js
```diff
@@ -54,11 +69,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
+
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const children = people
+    .filter(person => people.find(mother => mother.name === person.mother)
+    && (onlyWithSon ? person.sex === 'm' : person));
```
etojeDenys:
```suggestion
    .filter(person => people.some(mother => mother.name === person.mother)
    && (onlyWithSon ? person.sex === 'm' : person));
```
better to use `some` because you just want to know if this item exists in the array

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):
GJ!

PR #2334, opened 2023-08-12
add task solution

irahos:


src/averageAges.js
```diff
@@ -15,7 +15,24 @@
...
+      .filter(person => person['sex'] === 'm')
+      .map(person => (person['died'] - person['born']));
+
+    const totalMaleAge = maleAges.reduce((sum, age) => sum + age, 0);
+
+    return totalMaleAge / maleAges.length;
+  } else {
+    const maleAges = people
+      .filter(person => person['sex'] === 'm'
+        && Math.ceil(person['died'] / 100) === century)
```
vatatan: You can assign 'f', 'm', and 100 to variables

vatatan (approved):
Looks nice to me👌

src/averageAges.js
```diff
@@ -15,7 +15,24 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
+  if (century === undefined) {
```
Moroz-Dmytro: Try to simplify this and avoid using `if else`

src/averageAges.js
Moroz-Dmytro: Good point, please fix this

src/averageAges.js
```diff
@@ -15,7 +15,24 @@
...
+    const maleAges = people
+      .filter(person => person['sex'] === 'm')
+      .map(person => (person['died'] - person['born']));
+
+    const totalMaleAge = maleAges.reduce((sum, age) => sum + age, 0);
+
+    return totalMaleAge / maleAges.length;
```
Moroz-Dmytro: Remember about "DRY". You can create function for this

src/averageAges.js
```diff
@@ -37,7 +54,28 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  function isMother(name) {
+    return people.some(person => person['mother'] === name);
+  };
+
+  if (withChildren !== undefined) {
```
Moroz-Dmytro: Also try to simplify this, avoid using `if else`, because now you have a lot of code that is the same in `if` and `else` block

src/averageAges.js
```diff
@@ -54,11 +92,38 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  let childrenWithMothers = people
+    .filter(child => child['mother'] !== null);
+
+  if (onlyWithSon) {
+    childrenWithMothers = childrenWithMothers
+      .filter(child => child['sex'] === 'm');
```
Moroz-Dmytro: We can use destructuring where it's possible

src/averageAges.js
```diff
@@ -54,11 +92,38 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  }
+
+  const ageDifferences = childrenWithMothers.map(child => {
+    const mother = people.filter(person => person['sex'] === 'f')
```
Moroz-Dmytro: We don't need this filtering here, it will work without it

src/averageAges.js
```diff
@@ -54,11 +92,38 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      .find(person => person['name'] === child['mother']);
+
+    if (mother) {
+      return child['born'] - mother['born'];
+    } else {
+      return null;
+    }
```
Moroz-Dmytro: Better to use [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator)

Moroz-Dmytro: But do we really need this if?
`childrenWithMothers` is already filtered and here we have only children that have mother

src/averageAges.js
```diff
@@ -54,11 +92,38 @@ function calculateWomenAverageAge(people, withChildren) {
...
  * @return {number}
  */
+
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  let childrenWithMothers = people
+    .filter(child => child['mother'] !== null);
```
Moroz-Dmytro: But better will be to check if we have mother with such name in our `people` array

src/averageAges.js
```diff
@@ -54,11 +92,38 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      return null;
+    }
+  });
+
+  const ageDifferencesSorted = ageDifferences
+    .filter(ageDiff => ageDiff !== null);
+
+  if (ageDifferencesSorted.length === 0) {
+    return null;
+  }
```
Moroz-Dmytro: I guess we don't really need this

Moroz-Dmytro (changes requested):
Good progress 👍 
But some comments

src/averageAges.js
```diff
@@ -14,12 +14,26 @@
...
-  // without nesting
+  const maleAges
+    = century === undefined
+      ? people
+        .filter(({ sex }) => sex === MALE_GENDER)
+        .map(({ died, born }) => died - born)
+      : people
+        .filter(({ sex, died }) => sex === MALE_GENDER
+        && Math.ceil(died / CENTURY_DURATION) === century)
+        .map(({ died, born }) => died - born);
```
BudnikOleksii, line 34: You don't need to compare with undefined, just `!century` is enough.
Also, you repeat `map` in both cases, DRY. You can do the `!century` check inside a callback so it reduces your code.

src/averageAges.js
```diff
@@ -36,8 +50,23 @@ function calculateMenAverageAge(people, century) {
...
+
+  const womenAges
+    = withChildren === undefined
+      ? people
+        .filter(({ sex }) => sex === FEMALE_GENDER)
+        .map(({ died, born }) => died - born)
+      : people
+        .filter(({ sex, name }) => sex === FEMALE_GENDER
+          && isMother(name))
+        .map(({ died, born }) => died - born);
```
BudnikOleksii, line 67: The same as above, and the callback in `map` is the same, we can reuse it.

src/averageAges.js
```diff
@@ -54,11 +83,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
  * @return {number}
  */
+
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  let childrenWithPresentMothers = people
+    .filter(child => people.some(({ name }) => name === child['mother']));
```
BudnikOleksii, line 89:
```suggestion
    .filter(child => people.some(({ name }) => name === child.mother));
```
You can use syntaxis with dot as it's a simple property, the same in other places below

BudnikOleksii (approved):
Good work 👍 
I left a few comments, check them

PR #2342, opened 2023-08-14
Solution

serhiilapin:


src/averageAges.js
```diff
@@ -36,8 +47,24 @@
  *
  * @return {number}
  */
-function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+function calculateWomenAverageAge(people, withChildren = 0) {
+  let arrayOfAgeWoman;
+  let avgYear = 0;
+  const parameter = withChildren;
```
VikVolSen, line 55: Unnecessary variable 'parameter'. You could use 'withChildren' directly in your conditional statement.

src/averageAges.js
```diff
@@ -14,12 +14,23 @@
...
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+function calculateMenAverageAge(people, century = 0) {
+  let arrayOfAgeMan;
+  let avgYear = 0;
+
+  century !== 0
```
VikVolSen: Avoid unnecessary empty lines.

src/averageAges.js
```diff
@@ -36,8 +47,24 @@
...
  * @return {number}
  */
-function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+function calculateWomenAverageAge(people, withChildren = 0) {
+  let arrayOfAgeWoman;
+  let avgYear = 0;
+  const parameter = withChildren;
+
+  parameter !== 0
```
VikVolSen: Avoid unnecessary empty lines.

src/averageAges.js
```diff
@@ -55,10 +82,23 @@
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const parameter = onlyWithSon;
```
VikVolSen: Unnecessary variable 'parameter'. You could use 'onlyWithSon' directly in your conditional statement.

VikVolSen (changes requested):
Please check comments

src/averageAges.js
```diff
@@ -36,8 +46,22 @@
  *
  * @return {number}
  */
-function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+function calculateWomenAverageAge(people, withChildren = 0) {
```
VadimMentorQA: Why do we need to assign any default value as the parameter for the function `calculateWomenAverageAge`?

src/averageAges.js
```diff
@@ -14,12 +14,22 @@
  *
  * @return {number}
  */
-function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+function calculateMenAverageAge(people, century = 0) {
```
VadimMentorQA: The function `calculateMenAverageAge` should use the default parameter as `false` for `century` and not `0`, as `0` is a falsy value.

src/averageAges.js
```diff
@@ -55,10 +79,21 @@
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const filteredPeople = people.filter((person) =>
+    onlyWithSon
```
VadimMentorQA, line 96: It's more efficient to find a mother once and reuse the result in both conditions instead of finding the same mother twice.

src/averageAges.js
```diff
@@ -55,10 +79,21 @@
...
+      ? people.some(
+        (mother) => mother.name === person.mother && person.sex === 'm'
+      )
+      : people.find((mother) => mother.name === person.mother)
+  );
+  const ageDifferences = filteredPeople.map(
+    (person) =>
+      person.born - people.find((mother) => mother.name === person.mother).born
```
VadimMentorQA: You are finding the same mother again, it's better to find it once and reuse the result.

VadimMentorQA (changes requested):
1. Exclude your package-lock.json from your PR.
2. Make sure all tests pass successfully before requesting for review.

src/averageAges.js
```diff
@@ -55,10 +79,22 @@
...
+    onlyWithSon
+      ? people.some(
+        (mother) => mother.name === person.mother && person.sex === 'm'
+      )
+      : people.find((mother) => mother.name === person.mother)
+  );
+  
+  const ageDifferences = filteredPeople.map(
+    (person) =>
+      person.born - people.find((mother) => mother.name === person.mother).born
```
VikVolSen: The 'find' method is used inside a 'map' method which will increase time complexity. Consider finding all mothers before mapping over 'filteredPeople'.

VikVolSen (changes requested):
Please fix failed tests in CI, fix errors in lines 21, 53

VadimMentorQA (approved):
Good job

PR #2343, opened 2023-08-14
add task solution

edgergard:


src/averageAges.js
```diff
@@ -15,11 +15,16 @@
...
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const males = people.filter((person) =>
+    century
+      ? person.sex === 'm' && Math.ceil(person.died / 100) === century
+      : person.sex === 'm');
```
BudnikOleksii: Let's create variables for "m", "f". It will be easier to manage if in the future sex field will have other values for male and female

src/averageAges.js
```diff
@@ -37,7 +42,17 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const females = people.filter((person) =>
+    withChildren
+      ? person.sex === 'f'
+        && people.find((child) => person.name === child.mother)
+      : person.sex === 'f');
+
+  const womenAverageAge = females.reduce((acc, person) => (
+    acc + person.died - person.born
+  ), 0) / females.length;
```
BudnikOleksii: You use the same callback to calculate the average, let's move it to a separate helper function.

BudnikOleksii (changes requested):
Good work!
Left a few comments, check them

src/averageAges.js
```diff
@@ -1,5 +1,13 @@
 'use strict';
 
+const SEX_MALE = 'm';
+const SEX_FEMALE = 'f';
+
+function calculateAverageAge(people) {
+  return people.reduce((acc, person) => (
+    acc + person.died - person.born
+  ), 0) / people.length;
```
tiserett, line 9: it's gonna be more readable if you create variable for people.reduce result and then return this variable / people.length;

tiserett (approved):

PR #2344, opened 2023-08-14
add task solution

NazarVotkanych:


src/averageAges.js
```diff
@@ -14,7 +14,26 @@
...
+
+    return acc + age;
+  }, 0) / array.length;
+}
+
 function calculateMenAverageAge(people, century) {
+  const filterMen = century
+    ? people.filter(person => person.sex === 'm'
+    && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === 'm');
```
BudnikOleksii: You could just pass different callbacks to the filter method depending on the condition.
You can name just men in this case.
Also, let's create variables for sex values such as "m", "f"

src/averageAges.js
```diff
@@ -14,7 +14,26 @@
...
+
+  const age = filterMen.reduce((acc, person) =>
```
```suggestion
  const totalAge = filterMen.reduce((acc, person) =>
```

src/averageAges.js
```diff
@@ -37,7 +56,13 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const womenWithChild = people.filter(person =>
+    person.sex === 'f' && people.some(perso => perso.mother === person.name));
+  const women = people.filter(person => person.sex === 'f');
```
BudnikOleksii: One of the arrays is redundant as we don't use it

src/averageAges.js
```diff
@@ -14,7 +14,26 @@
...
+
 function calculateMenAverageAge(people, century) {
+  const filterMen = century
+    ? people.filter(person => person.sex === 'm'
+    && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === 'm');
+
+  const age = filterMen.reduce((acc, person) =>
+    acc + (person.died - person.born), 0);
+  const averageAge = age / filterMen.length;
```
BudnikOleksii: You have a helper function calculateAge, you can use it

BudnikOleksii (changes requested):
Good work!

src/averageAges.js
```diff
@@ -14,7 +14,28 @@
  *
  * @return {number}
  */
+
+const sexMale = 'm';
+const sexFemale = 'f';
```
roman-mirzoian: Static constants are usually named using uppercase.

src/averageAges.js
```diff
@@ -14,7 +14,28 @@
...
+
+    return acc + age;
+  }, 0) / array.length;
+}
+
 function calculateMenAverageAge(people, century) {
+  const filterMen = century
+    ? people.filter(person => person.sex === sexMale
+    && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === sexMale);
```
roman-mirzoian: You can move the condition inside the filter and then the code will be shorter.

src/averageAges.js
```diff
@@ -14,7 +14,28 @@
...
+function calculateAge(array) {
+  return array.reduce((acc, person) => {
+    const age = person.died - person.born;
+
+    return acc + age;
+  }, 0) / array.length;
+}
+
 function calculateMenAverageAge(people, century) {
+  const filterMen = century
```
```suggestion
  const filteredMen = century
```

src/averageAges.js
```diff
@@ -14,7 +14,28 @@
...
+    ? people.filter(person => person.sex === sexMale
+    && Math.ceil(person.died / 100) === century)
+    : people.filter(person => person.sex === sexMale);
+
+  const totalAge = calculateAge(filterMen);
+
+  return totalAge;
```
```suggestion
  return calculateAge(filteredMen);
```

src/averageAges.js
```diff
@@ -37,7 +58,15 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const womenWithChild = people.filter(person =>
+    person.sex === sexFemale && people.some(perso =>
+      perso.mother === person.name)
+  );
+  const women = people.filter(person => person.sex === sexFemale);
+
+  return withChildren
+    ? calculateAge(womenWithChild)
+    : calculateAge(women);
```
roman-mirzoian: Move the `withChildren` check inside the `filter`, so you don't have to go through the array unnecessarily.

src/averageAges.js
```diff
@@ -14,7 +14,28 @@
  *
  * @return {number}
  */
+
+const sexMale = 'm';
+const sexFemale = 'f';
+
+function calculateAge(array) {
+  return array.reduce((acc, person) => {
+    const age = person.died - person.born;
```
roman-mirzoian: Use destructuring on the object.

src/averageAges.js
```diff
@@ -14,7 +14,28 @@
  *
  * @return {number}
  */
+
+const sexMale = 'm';
+const sexFemale = 'f';
+
+function calculateAge(array) {
```
roman-mirzoian: You don't just count the age, but return the `average` value, so you need to add the name more precisely. 
Also, don't use an `array` to specify an argument, choose a more meaningful name.

src/averageAges.js
```diff
@@ -55,10 +84,24 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const childrenWithMothers = people.filter(child => {
+    return onlyWithSon
```
roman-mirzoian: Move the check to the middle of the `filter` to shorten the code.

src/averageAges.js
```diff
@@ -55,10 +84,24 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    const mother = people.find(person => person.name === child.mother);
+
+    return child.born - mother.born;
+  });
+
+  const sumAgeDiffBetween = ageDiffBetween.reduce((sum, ageDiff) =>
+    sum + ageDiff, 0);
+  const averageAgeDifference = sumAgeDiffBetween / ageDiffBetween.length;
+
+  return averageAgeDifference;
```
```suggestion
  return sumAgeDiffBetween / ageDiffBetween.length;
```

src/averageAges.js
```diff
@@ -55,10 +84,24 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  });
+
+  const ageDiffBetween = childrenWithMothers.map(child => {
+    const mother = people.find(person => person.name === child.mother);
+
+    return child.born - mother.born;
+  });
+
+  const sumAgeDiffBetween = ageDiffBetween.reduce((sum, ageDiff) =>
+    sum + ageDiff, 0);
```
roman-mirzoian: Move the logic from the `map` to the middle of the `reduce` so as not to go through the arrays one more time. 
Also, don't forget to use destructuring for the object.

roman-mirzoian (changes requested):

BudnikOleksii (approved):
Nicely done!

PR #2346, opened 2023-08-14
solution

mr-olezha:


src/averageAges.js
```diff
@@ -1,64 +1,93 @@
...
- * Divide year of person's death by 100: Math.ceil(person.died / 100)
- *
  * @param {object[]} people
  * @param {number} century - optional
  *
  * @return {number}
  */
+
+const man = 'm';
+const woman = 'f';
```
roman-mirzoian, line 11: Static constants are usually named using uppercase.

src/averageAges.js
```diff
@@ -1,64 +1,93 @@
...
-  // 4. return the average value
+  const peopleWithMothers = people
+    .map((person) => {
+      const mother = people.find((human) => human.name === person.mother);
+
+      return {
+        ...person,
+        mother,
+      };
+    });
```
roman-mirzoian, line 73: The search for the mother can be done in finite `reduce`, then there will be no need to go through the array and add a new field one more time.
The same logic for `filteredPeopleByMother`.

src/averageAges.js
```diff
@@ -1,64 +1,93 @@
...
  */
+
+const man = 'm';
+const woman = 'f';
+
+function calculateAgeSum(people) {
+  return people.reduce((acc, person) => {
+    return acc + (person.died - person.born);
+  }, 0);
+}
```
roman-mirzoian, line 17:
```suggestion
function calculateAgeSum(people) {
  return people.reduce((acc, { died, born }) => acc + (died - born), 0);
}
```

src/averageAges.js
```diff
@@ -1,64 +1,93 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const filteredPeopleByMan = people.filter((person) => {
+    if (century) {
+      const personDeathCentury = Math.ceil(person.died / 100);
```
roman-mirzoian, line 22: 100 is also "magic value".

roman-mirzoian (changes requested):

PR #2347, opened 2023-08-14
solution

Valentyn-Bilyk:


src/averageAges.js
```diff
@@ -14,12 +14,31 @@
...
+const MALE_SEX = 'm';
+const FEMALE_SEX = 'f';
+const CENTURY_VALUE = 100;
+
+function calculateAverage(arrayPeople) {
+  return arrayPeople.reduce((sumAges, person) => {
+    const age = person.died - person.born;
+
+    return sumAges + age;
+  }, 0) / arrayPeople.length;
```
tiserett, line 26: it would be better if you save result of reduce in a variable, and then use it with return and / arrayPeople.length, gonna be more readable

tiserett (approved):

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #2349, opened 2023-08-15
Solution

oksom:


src/averageAges.js
```diff
@@ -55,10 +63,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  )
+    : people;
+
+  const ageDiffs = filteredPeople.map(person => {
+    const mother = people.find(m => m.name === person.mother);
+
+    return mother ? person.born - mother.born : 0;
+  });
+
+  const validAgeDiffs = ageDiffs.filter(diff => diff !== 0);
```
anastasiia-tilikina:
```suggestion
  const validAgeDiffs = ageDiffs.filter(diff => diff);
```

anastasiia-tilikina:
```suggestion
  const validAgeDiffs = ageDiffs.filter(diff => diff);
```

src/averageAges.js
```diff
@@ -55,10 +63,22 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const totalAgeDiff = validAgeDiffs.reduce((sum, ageDiff) => sum + ageDiff, 0);
```
anastasiia-tilikina, line 79: looks like we have the same logic for finding the average age. Let's move it to helper and reuse

anastasiia-tilikina (changes requested):
GJ!

let's jus make some improvements.

lerastarynets (approved):
You've got this, keep pushing forward!

PR #2350, opened 2023-08-15
Solution

okanivan:


src/averageAges.js
```diff
@@ -36,8 +45,20 @@ function calculateMenAverageAge(people, century) {
  *
  * @return {number}
  */
+
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = withChildren
+    ? people.filter(person =>
+      person.sex === 'f' && people.some(child => child.mother === person.name))
```
roman-mirzoian: Use the created constants to avoid using "magic values".

src/averageAges.js
```diff
@@ -36,8 +45,20 @@ function calculateMenAverageAge(people, century) {
...
+    : people.filter(person => person.sex === FEMALE);
+
+  const womenTotalAge = women.reduce((acc, woman) => {
+    return acc + (woman.died - woman.born);
+  }, 0);
+
+  const averafeWoomanAge = womenTotalAge / women.length;
```
roman-mirzoian: You do a similar calculation for `men`, so you should put it in a separate function and use it as a helper.

src/averageAges.js
```diff
@@ -55,10 +76,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const children = people.filter((person, _, peopleArr) => {
+    const hasMother = peopleArr.find(mother => mother.name === person.mother);
+
+    return hasMother && (onlyWithSon ? person.sex === 'm' : true);
```
roman-mirzoian: Use the created constants to avoid using "magic values".

src/averageAges.js
```diff
@@ -55,10 +76,23 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  const childrenTotalAge = children.reduce((acc, child,) => {
+    const mother = people.find(person => person.name === child.mother);
+
+    if (mother) {
+      return acc + child.born - mother.born;
+    }
+
+    return acc;
+  }, 0);
```
```suggestion
  const childrenTotalAge = children.reduce((acc, { born, mother: motherName }) => {
    const motherBirthYear = people.find(({ name }) => name === motherName).born;

    if (motherBirthYear) {
      return acc + born - motherBirthYear;
    }

    return acc;
  }, 0);
```

src/averageAges.js
```diff
@@ -55,10 +76,23 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const children = people.filter((person, _, peopleArr) => {
+    const hasMother = peopleArr.find(mother => mother.name === person.mother);
```
```suggestion
  const children = people.filter((person) => {
    const hasMother = children.find(mother => mother.name === person.mother);
```

roman-mirzoian (changes requested):

src/averageAges.js
```diff
@@ -14,14 +14,24 @@
...
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+const MALE = 'm';
+const FEMALE = 'f';
+const ONE_CENTURY = 100;
+
+function getAverageAges(persons) {
+  return persons.reduce((sumAge, person) =>
```
tiserett, line 22: replace persons with people

src/averageAges.js
```diff
@@ -14,14 +14,24 @@
...
+    sumAge + (person.died - person.born), 0) / persons.length;
 }
 
+function calculateMenAverageAge(people, century) {
+  const men = people.filter(person => person.sex === MALE && (century
+    ? Math.ceil(person.died / ONE_CENTURY) === century
+    : true));
```
tiserett, line 29: not the best formatting actually, i'd rather immediately check if there's century or no, and then write filter function

src/averageAges.js
```diff
@@ -36,8 +46,18 @@ function calculateMenAverageAge(people, century) {
...
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const women = withChildren
+    ? people
+      .filter(person => person.sex === FEMALE && people
+        .some(child => child.mother === person.name))
+    : people
+      .filter(person => person.sex === FEMALE);
+
+  const averafeWoomanAge = getAverageAges(women);
```
tiserett, line 58: averafe -> average

src/averageAges.js
```diff
@@ -55,10 +75,25 @@ function calculateWomenAverageAge(people, withChildren) {
...
+  const filteredPeople = people.filter(person => {
+    const hasMother = people.find(mother => mother.name === person.mother);
+
+    return hasMother && (onlyWithSon ? person.sex === MALE : true);
+  });
+
+  const childrenTotalAge = filteredPeople
+    .reduce((acc, { born, mother: motherName }) => {
+      const motherBirthYear = people
+        .find(({ name }) => name === motherName).born;
```
tiserett, line 87: you put methods on the new line if the're few of them, in your case you can just put condition on the new line and wrap it with braces, should be more readable

tiserett (approved):

PR #2351, opened 2023-08-15
add task solution

ghost:


src/averageAges.js
```diff
@@ -14,8 +14,18 @@
  *
  * @return {number}
  */
+const maleSex = 'm';
+const femaleSex = 'f';
```
roman-mirzoian, line 18: Static constants are usually named using uppercase.

src/averageAges.js
```diff
@@ -14,8 +14,18 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
+  const men = people.filter(person =>
+    isMale(person)
+    && (century
+      ? (Math.ceil(person.died / 100) === century)
```
roman-mirzoian, line 24: 100 is also "magic value".

src/averageAges.js
```diff
@@ -59,8 +77,41 @@ function calculateAverageAgeDiff(people, onlyWithSon) {
   // 2. keep people who have mothers in the array
   // 3. calculate the difference child.born - mother.born
   // 4. return the average value
+  const childrens = people.filter(
```
roman-mirzoian, line 80:
```suggestion
  const children = people.filter(
```

src/averageAges.js
```diff
@@ -59,8 +77,41 @@ function calculateAverageAgeDiff(people, onlyWithSon) {
...
+
+function getAge(person) {
+  return person.died - person.born;
 }
 
+function getAverageAge(people) {
+  return +(people.reduce((sum, person) =>
+    sum + getAge(person), 0) / people.length
+  ).toFixed(2);
+}
```
roman-mirzoian, line 110: Now everything is written correctly, but such code is difficult to read and maintain. Let's take it apart a little bit.
```suggestion
function getAverageAge(people) {
  const totalAge = people.reduce((sum, person) => sum + getAge(person), 0);
  return +(totalAge / people.length).toFixed(2);
}
```

src/averageAges.js
```diff
@@ -59,8 +77,41 @@ function calculateAverageAgeDiff(people, onlyWithSon) {
...
+  );
+
+  return childrens.reduce((sum, child) => {
+    const mother = people.find(
+      person =>
+        child.mother === person.name
+    );
+
+    return sum + (child.born - mother.born);
+  }, 0) / childrens.length;
```
roman-mirzoian, line 95: Don't forget to use destructuring for objects.
```suggestion
  const ageDiff = children.reduce((sum, { born, mother: motherName }) => {
    const motherBirthYear = people.find(({ name })=> name === motherName).born;
    return sum + (born - motherBirthYear);
  }, 0);
  return ageDiff / children.length;
```

roman-mirzoian (changes requested):

PR #2354, opened 2023-08-15
Solution

AnastasiiaHombalevska:


src/averageAges.js
```diff
@@ -1,64 +1,71 @@
...
  * @param {number} century - optional
  *
  * @return {number}
  */
+const MALE_SEX = 'm';
+const FEMALE_SEX = 'f';
+
+function findMother(people, person) {
+  return people.find(p => p.name === person.mother);
+}
```
roman-mirzoian, line 14: Try not to shorten the names to unreadable ones and don't forget to use destructuring for objects.
```suggestion
function findMother(people, person) {
  return people.find(({ name }) => name === person.mother);
}
```

src/averageAges.js
```diff
@@ -1,64 +1,71 @@
...
+
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const men = people.filter(person => person.sex === MALE_SEX
+    && (century
+      ? Math.ceil(person.died / 100) === century
```
roman-mirzoian, line 19: 100 is also "magic value". You can name constant `CENTURY_VALUE` for this number.

src/averageAges.js
```diff
@@ -1,64 +1,71 @@
...
+      : true));
+
+  const mensAge = men.map(man => man.died - man.born);
+
+  const averageMansAge = mensAge.reduce((acc, age) => {
+    return acc + age;
+  }, 0) / mensAge.length;
```
roman-mirzoian, line 26: Move the calculation logic from the `map` to the middle of `reduce`, so as not to go through the array one more time and improve the performance of the code.

src/averageAges.js
```diff
@@ -1,64 +1,71 @@
...
- * her as mother.
- *
  * @param {object[]} people
  * @param {boolean} withChildren - optional
  *
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  const femaleAge = people.filter(person => person.sex === FEMALE_SEX
```
roman-mirzoian, line 38: Whole objects will be stored in this variable, not only the age of women, so you need to choose a more suitable name.

src/averageAges.js
```diff
@@ -1,64 +1,71 @@
...
+    && (withChildren
+      ? people.some((child) => child.mother === person.name)
+      : person.sex === FEMALE_SEX));
+
+  const femaleAverageAge = femaleAge.reduce((ageAccum, person) =>
+    ageAccum + (person.died - person.born), 0) / femaleAge.length;
```
roman-mirzoian, line 44: A similar calculation is done for `men`, so it is worth taking it out into a separate function and using it as a helper.

src/averageAges.js
```diff
@@ -1,64 +1,71 @@
...
+
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  let filteredChildren = people.filter(person => findMother(people, person));
+
+  if (onlyWithSon) {
+    filteredChildren = filteredChildren.filter(child => child.sex === MALE_SEX);
```
roman-mirzoian, line 61:
```suggestion
    filteredChildren = filteredChildren.filter(({ sex }) => sex === MALE_SEX);
```

src/averageAges.js
```diff
@@ -1,64 +1,71 @@
...
+  }
+
+  filteredChildren = filteredChildren
+    .map((child) => child.born - findMother(people, child).born);
+
+  return filteredChildren.reduce((childEverageAge, motherEverageAge) =>
+    childEverageAge + motherEverageAge) / filteredChildren.length;
```
roman-mirzoian, line 68: Move the calculation logic from the `map` to the middle of `reduce`, so as not to go through the array one more time and improve the performance of the code.

roman-mirzoian (changes requested):

src/averageAges.js
BudnikOleksii: Yeah, you can use destructuring, or just a better name for a variable as `p` isn't informative.

BudnikOleksii (changes requested):
Good job!
Looks like Roman misclicked and commented instead of reject. Please, check the comments and improve your solution in accordance to suggestions.

BudnikOleksii (changes requested):
Good job!
Looks like Roman misclicked and commented instead of reject. Please, check the comments and improve your solution in accordance with suggestions.

PR #2356, opened 2023-08-15
Solution

Nikita211112:


src/averageAges.js
```diff
@@ -15,11 +15,13 @@
...
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  const peopleFilter = people.filter(person => century ? person.sex === 'm'
+  && Math.ceil(person.died / 100) === century : person.sex === 'm');
+
+  const peopleReduce = peopleFilter.reduce((sum, person) => sum
+    + (person.died - person.born), 0);
+
+  return peopleReduce / peopleFilter.length;
 }
```
IvanFesenko: You have similar code in `calculateWomenAverageAge` function - do not repeat your self, create helper function that will calculate avarage age
```
  const peopleReduce = peopleFilter.reduce((sum, person) => sum
    + (person.died - person.born), 0);

  return peopleReduce / peopleFilter.length;
  ```

File: src/averageAges.js [hunk 2]
```diff
@@ -37,7 +39,14 @@ function calculateMenAverageAge(people, century) {
...
-  // write code here
+  const womenFilter = people.filter(person => withChildren
+    ? people.find(mother => person.name === mother.mother)
+    : person.sex === 'f');
+
+  const womenrReduce = womenFilter.reduce((sum, pers) => sum
+  + (pers.died - pers.born), 0);
+
+  return womenrReduce / womenFilter.length;
 }
```
IvanFesenko: Use helper mentioned above

IvanFesenko (changes requested):
Good job, lets make your code cleaner

IvanFesenko (changes requested):
Comment above

Nikita211112:
> Good job, lets make your code cleaner

can you check my work?

PR #2360, opened 2023-08-16
average ages v1

oltarasov:


src/averageAges.js
```diff
@@ -55,10 +78,35 @@ function calculateWomenAverageAge(people, withChildren) {
...
+      const mother = people.find(person => person.name === child.mother);
+
+      if (mother) {
+        return child.born - mother.born;
+      }
+    });
+    const sumAge = mothersAndChildDiff.reduce((acc, current) => acc + current);
+
+    return sumAge / mothersAndChildDiff.length;
+  }
```
volodymyr-soltys97: There is a lot of duplication of the same code here, try to simplify it and not explicitly check for undefined

src/averageAges.js
```diff
@@ -15,11 +15,20 @@
...
+    const sumAge = peopleAge.reduce((acc, current) => acc + current);
+
+    return sumAge / peopleAge.length;
+  } else {
+    const currentPeople = people.filter((man) => man.sex === 'm');
+    const peopleAge = currentPeople.map((man) => man.died - man.born);
+    const sumAge = peopleAge.reduce((acc, current) => acc + current);
+
+    return sumAge / peopleAge.length;
+  }
```
volodymyr-soltys97: The same as below

src/averageAges.js
```diff
@@ -37,7 +46,21 @@ function calculateMenAverageAge(people, century) {
...
+    const womensAge = mothers.map((women) => women.died - women.born);
+    const sumAge = womensAge.reduce((acc, current) => acc + current);
+
+    return sumAge / womensAge.length;
+  } else {
+    const mothers = people.filter(person => person.sex === 'f');
+    const womensAge = mothers.map((women) => women.died - women.born);
+    const sumAge = womensAge.reduce((acc, current) => acc + current);
+
+    return sumAge / womensAge.length;
```
volodymyr-soltys97: You can create a helper function for this logic and use it everywhere

src/averageAges.js
```diff
@@ -37,7 +46,21 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  if (withChildren !== undefined) {
+    const mothers = people.filter(person =>
+      people.some(p => p.mother === person.name) && person.sex === 'f'
```
volodymyr-soltys97: Create a variable for magic word `f` and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Well done 🔥

PR #2364, opened 2023-08-18
Add task solution

spojrzenie:


src/averageAges.js
```diff
@@ -15,13 +15,21 @@
  * @return {number}
```
OleziO, line 15: Remove all comments.

src/averageAges.js
```diff
@@ -37,9 +45,23 @@ function calculateMenAverageAge(people, century) {
...
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let womens = people.filter(person => person.sex === 'f');
+
+  if (withChildren === true) {
```
OleziO: Replace with `if (withChildren)`. It will be the same

src/averageAges.js
```diff
@@ -37,9 +45,23 @@ function calculateMenAverageAge(people, century) {
...
+
+  const totalAge = womens.reduce((sum, person) => sum
+    + (person.died - person.born), 0);
+
+  const averageAge = parseFloat((totalAge / womens.length).toFixed(2));
+
+  return averageAge;
 }
 
+// console.log(calculateWomenAverageAge(people, true));
```

src/averageAges.js
```diff
@@ -55,10 +77,36 @@ function calculateWomenAverageAge(people, withChildren) {
...
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const mothers = people.filter(mother => people.some(person =>
+    person.mother === mother.name));
+
+  let kids = people.filter(mummy => mummy.mother != null);
+
+  if (onlyWithSon === true) {
```
OleziO: Same as previous.

OleziO (changes requested):

src/averageAges.js
```diff
@@ -37,7 +43,19 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
-  // write code here
+  let womens = people.filter(person => person.sex === 'f');
```
darokrk: The variable 'womens' should be named 'women' for grammatical correctness and use `const` instead of let

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
  * @return {number}
  */
 function calculateMenAverageAge(people, century) {
-  // write code here
-  // learn how to use array methods like .filter .map .some .every .find .reduce
-  // avoid using loop and forEach
-  // replace `if ()` statement with &&, || or ?:
-  // without nesting
+  let mens = people.filter(person => person.sex === 'm');
```
darokrk: The variable 'mens' should be named 'men' for grammatical correctness and should be `const` instead of let

src/averageAges.js
```diff
@@ -55,10 +73,36 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+  if (onlyWithSon) {
+    kids = people.filter(son => son.mother != null && son.sex === 'm');
+  }
+
+  let ageSum = 0;
+  let numberOfPairs = 0;
+
+  kids.forEach(kid => {
+    const mother = mothers.find(mummy => mummy.name === kid.mother);
```
darokrk, line 86: The usage of forEach() is discouraged as the task explicitly mentioned to avoid using loop and forEach. You can replace the forEach with a reduce function.

src/averageAges.js
```diff
@@ -15,11 +15,17 @@
...
-  // without nesting
+  let mens = people.filter(person => person.sex === 'm');
+
+  if (century !== undefined) {
+    mens = mens.filter(person => Math.ceil(person.died / 100) === century);
+  }
+
+  const totalAge = mens.reduce((sum, person) => sum
+    + (person.died - person.born), 0);
+  const averageAge = parseFloat((totalAge / mens.length).toFixed(2));
```
darokrk: toFixed(2) method rounds the number to keep only two decimals. But parseFloat() could introduce floating point precision issue. It's better to use Number() instead of parseFloat().

src/averageAges.js
```diff
@@ -37,7 +43,19 @@ function calculateMenAverageAge(people, century) {
...
+
+  if (withChildren) {
+    womens = womens.filter(woman => people.some(person =>
+      person.mother === woman.name));
+  }
+
+  const totalAge = womens.reduce((sum, person) => sum
+    + (person.died - person.born), 0);
+
+  const averageAge = parseFloat((totalAge / womens.length).toFixed(2));
```
darokrk: same as above

src/averageAges.js
```diff
@@ -55,10 +73,36 @@ function calculateWomenAverageAge(people, withChildren) {
...
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const mothers = people.filter(mother => people.some(person =>
+    person.mother === mother.name));
+
+  let kids = people.filter(mummy => mummy.mother != null);
```
darokrk, line 79: `const kids` instead of let

src/averageAges.js
```diff
@@ -55,10 +73,36 @@ function calculateWomenAverageAge(people, withChildren) {
...
+
+      ageSum += ageDifference;
+      numberOfPairs++;
+    }
+  });
+
+  if (numberOfPairs > 0) {
+    const averageAgeDiff = ageSum / numberOfPairs;
+
+    return parseFloat(averageAgeDiff.toFixed(2));
```
darokrk: same as comments above

darokrk (changes requested):
Looks good 👍🏽  left mentions for code improvements 🛠️

darokrk (approved):
good job 👍🏽

PR #2369, opened 2023-08-28
solution

TommasoGiacopini:


src/averageAges.js
```diff
@@ -55,10 +82,35 @@ function calculateWomenAverageAge(people, withChildren) {
  * @return {number}
  */
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const validPeople = people.filter(person => {
+    if (onlyWithSon) {
+      return person.sex === 'm' && person.mother !== undefined;
```
lerastarynets: You should not return 'undefined'. Instead, you should return a boolean value. You can use !! operator to convert undefined to boolean false.

src/averageAges.js
```diff
@@ -37,7 +44,27 @@ function calculateMenAverageAge(people, century) {
...
+
+    averageFemaleSum = femaleWithChild.reduce((all, person) =>
+      all + (person.died - person.born), 0);
+
+    const averageAgeFemaleChild = averageFemaleSum / femaleWithChild.length;
+
+    return averageAgeFemaleChild;
+  }
+
+  averageFemaleSum = filteredWomen.reduce((all, person) =>
```
lerastarynets: You can optimize your code by calculating the 'averageFemaleSum' and 'averageAgeFemale' in one pass using the .reduce() method. This will improve the performance of your code.

src/averageAges.js
```diff
@@ -55,10 +82,35 @@ function calculateWomenAverageAge(people, withChildren) {
...
+    } else {
+      return accumulator;
+    }
+  }, {
+    totalAgeDiff: 0, count: 0,
+  });
+
+  const averageAgeDiff = count > 0 ? totalAgeDiff / count : 0;
+
+  return parseFloat(averageAgeDiff.toFixed(2));
```
lerastarynets: You are converting the 'averageAgeDiff' to a string using the .toFixed() method and then converting it back to a number using parseFloat(). This is unnecessary. You can use the .toFixed() method if you want to display the number, but you need to keep it as a number for internal calculations.

src/averageAges.js
```diff
@@ -55,10 +82,35 @@ function calculateWomenAverageAge(people, withChildren) {
...
 function calculateAverageAgeDiff(people, onlyWithSon) {
-  // 1. find a mother of each person (or only for men)
-  // 2. keep people who have mothers in the array
-  // 3. calculate the difference child.born - mother.born
-  // 4. return the average value
+  const validPeople = people.filter(person => {
+    if (onlyWithSon) {
+      return person.sex === 'm' && person.mother !== undefined;
+    } else {
+      return person.mother !== undefined;
```
lerastarynets:
```suggestion
      return !!person.mother;
```

src/averageAges.js
```diff
@@ -37,7 +44,27 @@ function calculateMenAverageAge(people, century) {
  * @return {number}
  */
 function calculateWomenAverageAge(people, withChildren) {
```
lerastarynets, line 58: you can make reusable function `calculateAverageAge` and use it inside `calculateWomenAverageAge` and `calculateMenAverageAge`

lerastarynets (changes requested):
Nice start! Let's fix the solution according to the checklist

TommasoGiacopini:
I corrected the code as suggested.

anastasiia-tilikina (changes requested):
We still have to optimize `calculateWomenAverageAge` function.
steps to optimize:
1. filter `women` array to get only women with children if needed
`filteredWomen = people.filter(person => person.sex === 'f').filter(person => !withChildren ||
      people.some(child => child.mother === person.name));`

2. get women ages
`const ages =  filteredWomen.map(woman => woman.died - women.born)`

3. Now we can use reduce method only once without code duplicates.

Also, create helper `calculateAverageAge` and use it everywhere

TommasoGiacopini:
Optimized calculateWomenAverageAge function.

anastasiia-tilikina (changes requested):
Great! We're almost there!

the only one thing left:
create a helper function `getAverageValue` to calculate average age and use it everywhere

witflash (approved):
Well done 👍

PR #2376, opened 2023-12-09
Solution

Spstv-sys:


src/averageAges.js
```diff
@@ -37,7 +48,21 @@ function calculateMenAverageAge(people, century) {
...
+  let count = 0;
+
+  people.forEach(person => {
+    if (person.sex === 'f' && (!withChildren || hasChildren(person, people))) {
+      totalAge += person.died - person.born;
+      count++;
+    }
+  });
+
+  return count === 0 ? 0 : totalAge / count;
```
vlasiuksasha, line 69: Instead of checking if count equals 0 at the end, check at the start if the array is empty and if it is, return 0.

src/averageAges.js
```diff
@@ -15,7 +15,18 @@
...
+
+  people.forEach(person => {
+    if (person.sex === 'm' && (!century
+      || Math.ceil(person.died / 100) === century)) {
+      totalAge += person.died - person.born;
+      count++;
+    }
+  });
+
+  return count === 0 ? 0 : totalAge / count;
```
vlasiuksasha: Instead of checking if count equals 0 at the end, check at the start if the array is empty and if it is, return 0.

src/averageAges.js
```diff
@@ -55,7 +80,44 @@ function calculateWomenAverageAge(people, withChildren) {
...
+          totalAgeDiff += person.born - mother.born;
+          count++;
+        }
+      }
+    });
+  }
+
+  const tot = totalAgeDiff / count;
+
+  return count === 0 ? 0 : tot;
```
vlasiuksasha, line 124: Instead of checking if count equals 0 at the end, check at the start if the array is empty and if it is, return 0.

vlasiuksasha (changes requested):
check the comments

PR #2377, opened 2023-12-09
Age soultion version 1

steveforde:
Average age Soultion

### DIFF
--- a/src/averageAges.js
+++ b/src/averageAges.js
@@ -1,13 +1,13 @@
 'use strict';
 
+function calculateAverage(people) {
+  const totalAge = people.reduce((sum, person) => sum
+   + (person.died - person.born), 0);
+
+  return totalAge / people.length;
+}
+
 function calculateMenAverageAge(allPeople, century) {
-  function calculateAverage(people) {
-    const totalAge = people.reduce((sum, person) =>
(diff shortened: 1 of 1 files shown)


anastasiia-tilikina (changes requested):
well done!

seems like your logic of calculating average value as well as finding person
lets move this to helpers to avoid code duplication

src/averageAges.js
```diff
@@ -1,31 +1,44 @@
...
-function calculateMenAverageAge(people, century) {
-  const men = people.filter(person =>
-    person.sex === 'm' && (!century || Math.ceil(person.died
-      / 100) === century));
-  const totalAge = men.reduce((sum, man) =>
-    sum + (man.died - man.born), 0);
-
-  return totalAge / men.length;
+function calculateMenAverageAge(allPeople, century) {
+  function calculateAverage(people) {
```
lerastarynets: this func is same for `calculateMenAverageAge` and `calculateWomenAverageAge` pls move it to the global scope and reuse in both of the mentioned functions

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit more

witflash (approved):
Great work!
Your code looks pretty fine!
Well done 👍🏻

